PR #24, opened 2018-08-10
Removed three instances of vcdvdchref

auppunda:
I updated composevapp, composerawvapp, and InstantiateVAppTemplate to use the href of the vdc calling the method instead of the vcdvdchref. I changed composevapp to be called by a vdc instead of a vapp since vdc's handle the creation of vapps. It didn't make sense for composerawvapp to be a vdc function and composevapp not to be. Also updated the composevapp test to use a real vcd. Updated every test that uses composevapp to be called by a vdc not a vapp. Changed getParentVdc test to be more robust.


lvirbalas:
> In general, 1-letter variables are acceptable only if used for a loop counter, in a short block of code. In all other cases, the names should tell what the variable is.

Fully agree.

> Another convention that we should establish is a more telling abbreviation for vcd and vdc. These two entities are quite different but easy to misspell, especially if used together. Can we agree on using vcdir and vdcent for these two entities all over the project? (or similar differentiating abbreviations)

vCD and vDC are such core abbreviations in vCloud Director, that people just get used to the one letter difference between the two. Same in code, for a person who is working with vCD for years (and that is our customers), "vcd" and "vdc" are **less** confusing than the suggested "vcdir" and "vdcent". BTW, a common longer abbreviation in the wide open internet for "vcd" is "vcloud".

As there's potentially more harm than benefit of changing these abbreviations, let's keep using "vcd" and "vdc".

PR #28, opened 2018-08-13
Removed the main instances of vcdvdchref

auppunda:
I updated composevapp, composerawvapp, and InstantiateVAppTemplate to use the href of the vdc calling the method instead of the vcdvdchref. I changed composevapp to be called by a vdc instead of a vapp since vdc's handle the creation of vapps. It didn't make sense for composerawvapp to be a vdc function and composevapp not to be. Also updated the composevapp test to use a real vcd. Updated every test that uses composevapp to be called by a vdc not a vapp. Changed getParentVdc test to be more robust.


auppunda:
The commit moves it into the file to make it a vdc function. I looked at how ComposeVApp was called and it was called using an empty VApp, so there is no way of actually getting the vdc href from that. For me to actually remove vcdvdchref from the method I had to move it to the right class. It also doesn't make sense for composevapp to be a vapp function while composeRawVApp is a vdc function. Also my commit message says this as well. I can update my commit title though

auppunda:
It removes all instances of vcdvdchref besides in the function retrievevdc. Retrievevdc will be removed when I clean api_vcd.go, but I left it in there right now to make sure I don’t break the SDK

PR #30, opened 2018-08-14
Travis interaction - iteration #1

dataclouder:
This is the first step towards Travis interaction, to allow an automated build.

* Current status:
  * Instructions in .travis.yml are ~ 3 years old, and do not reflect the status of the project.
  * Automated tests with `go test ./...` fail because of the setup and lack of tests in all directories. At the very minimum, they would require a different default location for the configuration file.
  * The project installation with `go get -t` fails for lack of go files in the root directory.
  * The web hooks defined in `.travis.yml` can't be validated.
* Changes made:
  * included a dummy `main.go` in the root directory to allow `go get -t` to work. This change can be reverted when we have a better Makefile.
  * removed from `.travis.yml` all instructions that don't presently work.
* To do:
  * This task will be reviewed as soon as we have a working automated test (even a partial one.)

I will need at least 2 LGTM to push these changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,7 @@ before_install:
 - go get github.com/mattn/goveralls
 
 script:
-- PATH="$HOME/gopath/bin:$PATH"
-- hack/coverage
+- go build .
 
 after_success:
 - echo "Build Successful!" 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 sudo: false
-install: true
+install: false
 
 before_install:
 - go get golang.org/x/tools/cmd/cover
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,18 +4,13 @@ install:
 - go get -t 
 - go get golang.org/x/tools/cmd/cover
 - go get github.com/mattn/goveralls
+
 script:
 - PATH="$HOME/gopath/bin:$PATH"
 - hack/coverage
-notifications:
-  slack:
-    secure: IdWZxF2eNtPPjJ9NFD7ebTxjyFpjNeCUEa87RXLhvM7Ne1g/VUltsRvCX9a7Mnw4K468/iUmO+jS9y5CFBuOlO2lOzFzaTJZlKNDxw5V2RaSbDYsrgDmhg0TsrajFwDbgNHCP8LPbBejGjsPBi80oRvc/00czLF3vKO28rAfbpE=
-  webhooks:
-    urls:
-      - secure: "IVp6dIdLTZS28zQVhrTn88U2W75cM1KNrbJw/U+wx7odg5JWDaqnwnI7c1zXPhyXh2rISG+HXuAxVlGVvcpwY73FlYPMQ97B5YbBhOX+Dv8w20sI18FHsCJUQ7SQYETK5vNaYcup2n8v5G7J2XbCQ7bZl125xwVEGCN8lPnLcIY="
-    on_start: true
+
 after_success:
-- figlet "Build Successful!" | /usr/games/cowsay -n -e "^^"
+- echo "Build Successful!" 
 after_failure:
-- figlet "Build Failed!" | /usr/games/cowsay -n -s
+- echo "Build Failed!" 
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,21 +1,16 @@
...
-- hack/coverage
-notifications:
-  slack:
-    secure: IdWZxF2eNtPPjJ9NFD7ebTxjyFpjNeCUEa87RXLhvM7Ne1g/VUltsRvCX9a7Mnw4K468/iUmO+jS9y5CFBuOlO2lOzFzaTJZlKNDxw5V2RaSbDYsrgDmhg0TsrajFwDbgNHCP8LPbBejGjsPBi80oRvc/00czLF3vKO28rAfbpE=
-  webhooks:
-    urls:
-      - secure: "IVp6dIdLTZS28zQVhrTn88U2W75cM1KNrbJw/U+wx7odg5JWDaqnwnI7c1zXPhyXh2rISG+HXuAxVlGVvcpwY73FlYPMQ97B5YbBhOX+Dv8w20sI18FHsCJUQ7SQYETK5vNaYcup2n8v5G7J2XbCQ7bZl125xwVEGCN8lPnLcIY="
-    on_start: true
+- go build .
+
```
auppunda: We would also like to check for go fmt.  https://stackoverflow.com/questions/42510140/check-golang-format-for-continous-integration .

main.go
```diff
@@ -0,0 +1,7 @@
+package main
+
+import "os"
+
+func main() {
+  os.Exit(0)
+}
```
auppunda: we should look into changing the makefile instead of adding a main.go

.travis.yml
```diff
@@ -1,21 +1,16 @@
...
 - go get golang.org/x/tools/cmd/cover
 - go get github.com/mattn/goveralls
+
 script:
-- PATH="$HOME/gopath/bin:$PATH"
-- hack/coverage
-notifications:
-  slack:
-    secure: IdWZxF2eNtPPjJ9NFD7ebTxjyFpjNeCUEa87RXLhvM7Ne1g/VUltsRvCX9a7Mnw4K468/iUmO+jS9y5CFBuOlO2lOzFzaTJZlKNDxw5V2RaSbDYsrgDmhg0TsrajFwDbgNHCP8LPbBejGjsPBi80oRvc/00czLF3vKO28rAfbpE=
-  webhooks:
```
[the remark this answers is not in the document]
hodgesrm: @auppunda, let's do the notifications later.  For now it's enough that you'll see a failure in the PR thanks to Travis running.

.travis.yml
hodgesrm: +1. That's an important check.

main.go
```diff
@@ -0,0 +1,7 @@
+package main
```
hodgesrm: We should not add a main.go here.  Instead, let's have travis prove it can build the sample.  That will show that we can link to the library.

dataclouder:
I have removed main.go and updated `Makefile` and `.travis.yml` to run the tests that are available today.

.travis.yml
```diff
@@ -1,21 +1,18 @@
...
-    urls:
-      - secure: "IVp6dIdLTZS28zQVhrTn88U2W75cM1KNrbJw/U+wx7odg5JWDaqnwnI7c1zXPhyXh2rISG+HXuAxVlGVvcpwY73FlYPMQ97B5YbBhOX+Dv8w20sI18FHsCJUQ7SQYETK5vNaYcup2n8v5G7J2XbCQ7bZl125xwVEGCN8lPnLcIY="
-    on_start: true
+- make
```
hodgesrm: How does this detect code that's not correctly formatted?  The Makefile does not look like examples I have seen, which fail if there's gofmt output.  (Apologies if I'm missing obvious behavior.)

auppunda: @hodgesrm gofmtcheck.sh checks if each file is go formatted properly. Check https://github.com/vmware/go-vcloud-director/blob/master/scripts/gofmtcheck.sh. Make runs fmtcheck which runs this script

PR #35, opened 2018-08-19
Tests refactoring part 1

dataclouder:
This is the first part of the tests refactoring (Issue#10) to get rid of the mock server in tests.
To simplify the review, I committed each file separately.

The purpose of this task (and a similar one by @auppunda with the remaining tests) is to make the test run with a real vCD, replicating the behavior of the previous test. 
There are more tests that can be created and modified, but that will come later, once the foundation for the test suite is completed.

A general note on the code: the test suite used `check.v1`, which recommends some odd behavior, like not using a name space for the testing library and squash the function names into the test name space. We will address This problem once the present task is complete (Issue#34)

If you are not acquainted with [check.v1](https://labix.org/gocheck) and some of the code looks odd, please ping me for help.

### DIFF
--- a/govcd/task_test.go
+++ b/govcd/task_test.go
@@ -4,25 +4,25 @@
 
 package govcd
 
+/*
+TO BE REMOVED (or reintroduced with different scope) : task completion is tested as part of vdc_test.go
 import (
+	"fmt"
 	. "gopkg.in/check.v1"
 )
 
-func (vcd *TestVCD) Test_WaitTaskCompletion(c *C) {
+func (vcd *TestVCD) Test_WaitTaskCompletion(check *C) {
 
-	testServer.Response(200, nil, taskExample)
+	fmt.Printf("Running: %s\n", check.TestName())
+	check.Skip("Disabled: need a reliable way of triggering a task")
+	fmt.Printf("%#v\n", vcd.vapp.VApp)
 	task, err := vcd.vapp.Deploy()
-	_ = testServer.WaitRequest()
-	testServer.Flush()
-	c.Assert(err, IsNil)
 
-	testServer.Response(200, nil, taskExample)
 	err = task.WaitTaskCompletion()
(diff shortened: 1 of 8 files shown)


govcd/task_test.go
```diff
@@ -5,22 +5,19 @@
...
+	"fmt"
 	. "gopkg.in/check.v1"
 )
 
-func (vcd *TestVCD) Test_WaitTaskCompletion(c *C) {
+func (vcd *TestVCD) Test_WaitTaskCompletion(check *C) {
 
-	testServer.Response(200, nil, taskExample)
+	fmt.Printf("Running: %s\n", check.TestName())
+	check.Skip("Disabled: need a reliable way of triggering a task")
```
auppunda: We can trigger a task with calling a method that creates a task for you.

dataclouder:
Updated PR with vdc test and minor changes to catalog item test

govcd/vdc_test.go
```diff
@@ -5,80 +5,97 @@
 package govcd
 
 import (
-	"github.com/vmware/go-vcloud-director/testutil"
-	types "github.com/vmware/go-vcloud-director/types/v56"
+	"github.com/vmware/go-vcloud-director/types/v56"
 	. "gopkg.in/check.v1"
-	"strconv"
+	// "strconv"
```
auppunda: delete strconv

govcd/vdc.go
```diff
@@ -268,6 +268,7 @@ func (v *Vdc) ComposeVApp(orgvdcnetworks []*types.OrgVDCNetwork, vapptemplate VA
 				Info: "Configuration parameters for logical networks",
 			},
 		},
+		AllEULAsAccepted: true,
```
auppunda: should pass this as a parameter

auppunda:
LGTM.

PR #37, opened 2018-08-21
Tests Refactoring Part 2

auppunda:
This is the second part of the test refactoring that was started by @dataclouder . This part deals with catalog_test.go, edgegateway_test.go, vapp_test.go, and query_test.go. 

The purpose of this pr (and a similar one by @dataclouder  with the remaining tests) is to make the tests run with a real vCD, replicating the behavior of the previous test.

I also didn't specify that I removed the vcd.config.VCD.VApp field because it wasn't needed, as I could use my vcd.vapp vapp that I create in the setup suite

The idea is to compose one vapp for all vapp testing, which the tests set up in the setup suite. It removes the vapp in the teardown suite method. This way we don't need a vapp element in our config file.

For edgegateway_tests, the user needs to provide an external ip and internal ip along with an edge gateway for the tests to be run

Look at https://github.com/vmware/go-vcloud-director/pull/35 for part 1

### DIFF
--- a/govcd/api_vcd_test.go
+++ b/govcd/api_vcd_test.go
@@ -44,11 +44,12 @@ type TestConfig struct {
 // an org, vdc, vapp, and client to run
 // tests on
 type TestVCD struct {
-	client *VCDClient
-	org    Org
-	vdc    Vdc
-	vapp   VApp
-	config TestConfig
+	client        *VCDClient
+	org           Org
+	vdc           Vdc
+	vapp          VApp
+	config        TestConfig
+	skipVappTests bool
 }
 
 var testServer = testutil.NewHTTPServer()
@@ -138,29 +139,25 @@ func (vcd *TestVCD) SetUpSuite(check *C) {
 		config.VCD.Catalog.Name != "" && config.VCD.Catalog.Catalogitem != "" {
 		vcd.vapp, err = vcd.createTestVapp("go-vapp-tests")
 		if err != nil {
-			panic(err)
+			fmt.Printf("%v", err)
+			vcd.skipVappTests = true
 		}
 	} else {
-		err = vcd.vdc.ComposeRawVApp("go-vapp-tests")
-		if err != nil {
-			panic(err)
-		}
-		vcd.vapp, err = vcd.vdc.FindVAppByName("go-vapp-tests")
-		if err != nil {
-			panic(err)
-		}
+		vcd.skipVappTests = true
+		fmt.Printf("Skipping all vapp tests because one of the following wasn't given: Network, StorageProfile, Catalog, Catalogitem")
 	}
 
 }
 
 func (vcd *TestVCD) TearDownSuite(check *C) {
+	if vcd.skipVappTests {
+		check.Skip("Vapp tests skipped, no vapp to be deleted")
+	}
 	err := vcd.vapp.Refresh()
 	if err != nil {
 		panic(err)
 	}
-	task, _ := vcd.vapp.PowerOn()
-	_ = task.WaitTaskCompletion()
-	task, _ = vcd.vapp.Undeploy()
+	task, _ := vcd.vapp.Undeploy()
 	_ = task.WaitTaskCompletion()
 	task, err = vcd.vapp.Delete()
 	if err != nil {
--- a/govcd/api_vcd_test.go
+++ b/govcd/api_vcd_test.go
@@ -134,11 +134,23 @@ func (vcd *TestVCD) SetUpSuite(check *C) {
 		panic(err)
 	}
 	// creates a new VApp for vapp tests
-	vapp, err := vcd.createTestVapp("go-vapp-tests")
-	if err != nil {
-		panic(err)
+	if config.VCD.Network != "" && config.VCD.StorageProfile.SP1 != "" &&
+		config.VCD.Catalog.Name != "" && config.VCD.Catalog.Catalogitem != "" {
+		vcd.vapp, err = vcd.createTestVapp("go-vapp-tests")
+		if err != nil {
+			panic(err)
+		}
+	} else {
+		err = vcd.vdc.ComposeRawVApp("go-vapp-tests")
+		if err != nil {
+			panic(err)
+		}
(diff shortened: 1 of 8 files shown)


govcd/api_vcd_test.go
```diff
@@ -131,7 +134,30 @@ func (vcd *TestVCD) SetUpSuite(test *C) {
 		panic(err)
 	}
 	// creates a new VApp for vapp tests
-	vcd.vapp = *NewVApp(&vcd.client.Client)
+	vapp, err := vcd.createTestVapp("go-vapp-tests")
```
dataclouder: Is this creation necessary? Can it be deferred to when a test requires such a vApp? 
If users are required to pass an existing vApp, then we should create a new one only when running destructive tests.

govcd/api_vcd_test.go
```diff
@@ -131,7 +134,30 @@ func (vcd *TestVCD) SetUpSuite(test *C) {
...
+	}
+	vcd.vapp = vapp
+}
+
+func (vcd *TestVCD) TearDownSuite(check *C) {
+	err := vcd.vapp.Refresh()
+	if err != nil {
+		panic(err)
+	}
+	task, _ := vcd.vapp.PowerOn()
```
dataclouder: Why do we have a power **ON** call in a tear down app? Shouldn't it be power OFF?
Also, we should check whether the vApp exists.

govcd/query_test.go
```diff
@@ -6,24 +6,16 @@ package govcd
...
 import (
 	// "fmt"
-	"github.com/vmware/go-vcloud-director/testutil"
+	//"github.com/vmware/go-vcloud-director/testutil"
 	. "gopkg.in/check.v1"
 )
 
-func (vcd *TestVCD) Test_Query(c *C) {
-
+// Right now it just checks if their is an error while querying
```
dataclouder: "their is an error" should be "there is an error"

govcd/edgegateway_test.go
```diff
@@ -5,192 +5,117 @@
...
-	_ = testServer.WaitRequests(2)
-	testServer.Flush()
-
-	c.Assert(err, IsNil)
-	c.Assert(edge.EdgeGateway.Name, Equals, "M916272752-5793")
-
-	testServer.Response(200, nil, edgegatewayExample)
+func (vcd *TestVCD) Test_Refresh(check *C) {
+	if vcd.config.VCD.EdgeGateway == "" {
+		check.Skip(fmt.Sprintf("skipping %s because no edgegatway given", check.TestName()))
```
dataclouder: `Skip()` provides already the name of the test (when we run the tests with `-check.v`)

govcd/edgegateway_test.go
```diff
@@ -5,192 +5,117 @@
...
+	check.Assert(edge.EdgeGateway.Name, Equals, vcd.config.VCD.EdgeGateway)
+
+	task, err := edge.AddNATMapping("DNAT", vcd.config.VCD.Externalip, vcd.config.VCD.Internalip, "77")
+	check.Assert(err, IsNil)
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
+	task, err = edge.RemoveNATMapping("DNAT", vcd.config.VCD.Externalip, vcd.config.VCD.Internalip, "77")
+	check.Assert(err, IsNil)
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
```
dataclouder: There should be a check of the final state of the mapping. (Future test enhancement to be marked in a TODO comment)

govcd/edgegateway_test.go
```diff
@@ -5,192 +5,117 @@
...
+	check.Assert(edge.EdgeGateway.Name, Equals, vcd.config.VCD.EdgeGateway)
+	task, err := edge.AddNATPortMapping("DNAT", vcd.config.VCD.Externalip, "1177", vcd.config.VCD.Internalip, "77")
+	check.Assert(err, IsNil)
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
+	task, err = edge.RemoveNATPortMapping("DNAT", vcd.config.VCD.Externalip, "1177", vcd.config.VCD.Internalip, "77")
+	check.Assert(err, IsNil)
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
```
dataclouder: Same set of checks as for TestNATMapping

govcd/edgegateway_test.go
```diff
@@ -5,192 +5,117 @@
...
+	check.Assert(edge.EdgeGateway.Name, Equals, vcd.config.VCD.EdgeGateway)
+	task, err := edge.Create1to1Mapping(vcd.config.VCD.Internalip, vcd.config.VCD.Externalip, "description")
+	check.Assert(err, IsNil)
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
+	task, err = edge.Remove1to1Mapping(vcd.config.VCD.Internalip, vcd.config.VCD.Externalip)
+	check.Assert(err, IsNil)
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
```
dataclouder: Same set of checks as for TestNATMapping

govcd/edgegateway_test.go
```diff
@@ -5,192 +5,117 @@
...
-	c.Assert(edge.EdgeGateway.Name, Equals, "M916272752-5793")
-
-	testServer.ResponseMap(2, testutil.ResponseMap{
-		"/api/admin/edgeGateway/00000000-0000-0000-0000-000000000000":                          testutil.Response{200, nil, edgegatewayExample},
-		"/api/admin/edgeGateway/00000000-0000-0000-0000-000000000000/action/configureServices": testutil.Response{200, nil, taskExample},
-	})
-
+func (vcd *TestVCD) Test_AddIpsecVPN(check *C) {
+	if vcd.config.VCD.Externalip == "" {
+		check.Skip(fmt.Sprintf("skipping %s because no valid ip given", check.TestName()))
```
dataclouder: `Skip()` already reports the test name

Issue #36, opened 2018-08-20
Add an "how to write a new test" guide


dataclouder:
Add a new guide explaining how to write a test that integrates into our suite.

The guide could be either stand-alone or a new section inside TESTING.md

Main topics:
* Which imports are needed;
* How to write a test function integrated into the existing suite;
* What is the structure of TestVCD and what contents to expect;
* What is [check.v1](https://labix.org/gocheck) and what it can do for us;
* Guidelines on testing code readability.

This issue depends on Issue#34, and should not start before its completion.

lvirbalas:
This is a good idea. Please amend to the existing TESTING.md - we want a single testing experience and type of tests for now.

dataclouder:
Added using PR #42

PR #42, opened 2018-08-22
Add a test writing guide to TESTING.md

dataclouder:
Signed-off-by: Giuseppe Maxia

## Description

Related issue: #36 

TESTING.md is now enhanced with instructions and examples on how to write a test.

The document includes guidelines and examples on how to create a test.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -8,9 +8,9 @@ export VCLOUD_CONFIG = $HOME/test.yaml
 If no environmental variable is set it will default to $HOME/config.yaml.
 
 
-# Example Config file
+## Example Config file
 
-```
+```yaml
 provider:
   user: root
   password: root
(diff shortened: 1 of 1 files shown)


TESTING.md
```diff
@@ -36,20 +36,156 @@ vcd:
...
+
+
+### Basic test function organization.
+
+Within the testing function, you should perform four actions:
+
+1. Run all operations that are needed for the test, such as finding a particular organization, or vDC, or vApp, deploying a VM, etc.
+2. Run the operation being tested (such as deploy a VM, retrieve a VM or vApp, etc.)
+3. Run the tests, which usually means using `check.Assert` or `check.Check` to compare known data with what was found during the test execution
+4. Optionally clean up all entities and data that may affect other test runs (e.g. delete something just deployed)
```
auppunda: This shouldn't be optional. I think the wording is a little confusing. It should state that if creating entities and data please clean it up as tests should be idempotent. If entities are not cleaned then it might ruin the idempotent behavior or the usefulness of the test. Also tests should not create any lasting entities on an user's vcd.

dataclouder: Thanks @auppunda . 
In fact, the wording is misleading, and it may look like the cleanup can be avoided. What I need to say instead is that the cleanup is mandatory when the test creates something new, while it is not needed when the test is read-only. 
I will make it more clear. Perhaps I will introduce a new heading for this purpose.

TESTING.md
```diff
@@ -2,60 +2,198 @@
...
+    vdc: org-vdc
+    catalog:
+        name: test
+        description: test catalog
+        catalogitem: ubuntu
+        catalogitemdescription: description
+    storageprofile:
+        storageprofile1: Development
+        storageprofile2: "*"
+    vapp
```
lvirbalas: Missing colon and vApp name.

TESTING.md
```diff
@@ -2,60 +2,198 @@
...
 If no environmental variable is set it will default to $HOME/config.yaml.
 
+## Example Config file
 
-# Example Config file
-
-```
+```yaml
 provider:
-  user: root
```
lvirbalas: I have tried with administrator account and the tests failed with a cryptic error message. I was able to run them after changing to _Org_ administrator. I suggest changing the sample from "root" to "orgAdmin".

Granted, we want the tests to create all of the below items automatically. As soon as we start creating Orgs and vDCs, we will require the System administrator.

auppunda: yes for that we will add an org variable into the provider as well, so that when the tests authenticate it does so with the provider org instead of the vcd org used for testing.

dataclouder:
Addressed @lvirbalas requests.
Please note that the configuration example will be improved with Issue #48

Issue #17, opened 2018-08-02
Remove testutil package from project


auppunda:
Eventually when all the tests are no longer reliant on testutil, we want to remove this package from the project.

dataclouder:
This issue is related to Issue#10 and it is addressed by pull requests #35 and #37.

dataclouder:
Removed with PR #43

PR #43, opened 2018-08-22
Remove testutil package from project

dataclouder:
As requested by Issue #17, since the test now work with a real vCD, we can dismiss the testutil package.
I've removed the last references to testutil in the code, and deleted the directory `testutil`.

Also removed the Mozilla license from the LICENSE file.

Signed-off-by: Giuseppe Maxia


auppunda:
lgtm. I am going to make a pull request that removes all the unnecessary xml after this

govcd/api_vcd_test.go
```diff
@@ -52,7 +51,6 @@ type TestVCD struct {
 	skipVappTests bool
 }
 
-var testServer = testutil.NewHTTPServer()
 
 var vcdu_api, _ = url.Parse("http://localhost:4444/api")
 var vcdu_v, _ = url.Parse("http://localhost:4444/api/versions")
```
auppunda: you can remove all localhost variables.

PR #57, opened 2018-08-24
Initial Catalog CRUD operations

auppunda:
Added Create, Get, Update, Delete functions for catalogs. Introduced the admin Catalog type and the CatalogOperations interface. Tests for Catalogs are also included. pr #32  should be merged before this one is though.


types/v56/types.go
```diff
@@ -758,6 +758,28 @@ type Catalog struct {
 	VersionNumber int64            `xml:"VersionNumber"`
 }
 
+// AdminCatalog represents the Admin view of a Catalog object.
```
dataclouder: Compared to `Catalog`, the only differences that I see are:
* `AdminCatalog` has `XMLName`  and `Xmlns` and `Catalog` doesn't (but it can have both because the API definition contains them)
* `Catalog` has `DateCreated` and `AdminCatalog` doesn't (but it should).
In short, the two structures are identical and should share the definition.

OTOH, the [API entry for AdminCatalog](https://code.vmware.com/apis/287/vcloud#/doc/doc/types/AdminCatalogType.html) defines several elements that are not in this structure, such as `CatalogStorageProfiles`, `ExternalCatalogSubscriptionParams`, `PublishExternalCatalogParams`. 
I think this definition needs more analysis.

auppunda: you are right, and I am going to add those elements to make the definition complete

auppunda:
@hodgesrm Fixed the issues

govcd/catalog.go
```diff
@@ -24,18 +24,94 @@ import (
...
+		Description: adminCatalog.AdminCatalog.Description,
+		IsPublished: adminCatalog.AdminCatalog.IsPublished,
+	}
+	adminCatalogHREF, err := url.ParseRequestURI(adminCatalog.AdminCatalog.HREF)
+	if err != nil {
+		return Task{}, fmt.Errorf("error parsing admin catalog's href: %v", err)
+	}
+	output, err := xml.MarshalIndent(vcomp, "  ", "    ")
+	if err != nil {
+		return Task{}, fmt.Errorf("error marshalling xml data for update", err)
```
dataclouder: fmt.Errorf requires a format directive in the string.

govcd/catalog_test.go
```diff
@@ -29,5 +29,44 @@ func (vcd *TestVCD) Test_FindCatalogItem(check *C) {
...
+	check.Assert(err, IsNil)
+}
+
+// Creates a Catalog, updates the description, and checks the changes against the
+// newly updated catalog. Then deletes the catalog
+func (vcd *TestVCD) Test_UpdateCatalog(check *C) {
+	org, err := GetAdminOrgByName(vcd.client, vcd.config.VCD.Org)
+	check.Assert(org, Not(Equals), AdminOrg{})
+	check.Assert(err, IsNil)
+	adminCatalog, err := org.CreateCatalog("UpdateCatalogTest", "UpdateCatalogTest", true)
```
dataclouder: The procedure should check whether the catalog already exists, and then either delete it or skip the creation.
Also, and _this request extends to all CRUD tests,_ the literal strings that are repeated (`"UpdateCatalogTest"`, `"DeleteCatalogTest"`, `"CatalogCreationTest"`) should become a local variable.

govcd/catalog.go
```diff
@@ -24,18 +24,94 @@ import (
...
+		return Task{}, fmt.Errorf("error updating catalog: %s : %s", err, adminCatalogHREF.Path)
+	}
+
+	catalog := NewAdminCatalog(adminCatalog.c)
+	if err = decodeBody(resp, catalog.AdminCatalog); err != nil {
+		return Task{}, fmt.Errorf("error decoding task response: %s", err)
+	}
+	adminCatalog.AdminCatalog = catalog.AdminCatalog
+	return Task{
+		Task: catalog.AdminCatalog.Tasks.Task[0],
```
dataclouder: This operation can easily fail. We are assuming that there is a `catalog.AdminCatalog.Tasks.Task[0]` element without checking whether `catalog.AdminCatalog.Tasks` is defined. In fact, the test fails exactly at this point with an invalid pointer dereference error.

Thus, what does it mean in this case to have an operation that responds with "200 OK" but there is no task? Is the task expected?
Please add the URL of the API being used.

Issue #68, opened 2018-08-27
Add a logging module to the library.


dataclouder:
Related to Issues #38 and #67, of which this one is a needed precondition, the library should adopt a consistent logging system. 

Currently, there are sparse calls to Go standard `log` package, without any configuration. The default output for these calls is stderr, i.e. the terminal screen.

The log module should implement the following:
* A global variable to enable logging, disabled by default.
* A global variable for log output, so that we can define where to send the logs when we don't want them on the screen;
* The log format (the default is `2009/11/10 23:00:00 message`
* (Optionally) The log level of detail (the default is INFO).

lvirbalas:
Let's print dates in ISO standard form: yyyy-mm-dd

Issue #67, opened 2018-08-27
Add logging and more error checking for all API calls


dataclouder:
The API calls in the library are all route through this paradigm:

(1)
`request := vcd.client.NewRequest(parameters)`
This implementation skips explicitly the possible errors returned by `http.NewRequest`. 
An error check (and logging) should be added to the function in api.go.

(2)
The APi is called using `response, err := checkResp(v.c.Http.Do(request))`
This call loses any visibility on the request, which can't be easily logged unless by adding it to every occurrence of `checkResp`
A possible improvement would be replacing `checkResponse(response as result of http.Do)` with `RunRequest(client, request)`. This change would allow us to log the contents of the request for all calls, before processing the response

(3)
The response is processed inside `checkResp` and sometimes it is lost before we can see its raw contents. (See Issue #54) 
We need to log the response within `checkResp` (or its proposed replacement `RunRequest`).

This refactoring would involve minimal global modification, with the logging handled in two or three places only.

Issue #38, opened 2018-08-22
Make all DEBUG output in the library uniform


dataclouder:
There are many calls that output a "DEBUG XML" line followed by the XML of the entity being processed.
In some cases, the output is conditional, depending on the value of an environment variable `GOVCLOUDAIR_DEBUG`.

We want to make this behavior predictable and consistent across all modules, as follows:
* Every debug message will depend on a global `bool` variable named `Vcd_debug`;
* The global variable will be initialized in the `init` function of api.go;
* The `Vcd_debug` variable will be modified by the environment variable `GOVCLOUDAIR_DEBUG` (any non-empty value will enable the feature). Clients can override this setting with their own criteria, since the variable will be public.
* Every current occurrence of `log.Printf(format_string, variable)` will be replaced by `Debug_msg(format_string, variable)`, where `Debug_msg` is a function in api.go defined as follows:

```go
func Debug_msg(format_message string, value interface{}) {
    if Vcd_debug {
        log.Printf(format_message, value)
    }
}
```

auppunda:
Yes this would be a nice feature to have.

PR #76, opened 2018-08-29
Add global logging of HTTP operations

dataclouder:
Added a global log management utility (in the new util package)
that regulates logging in the whole library.

Logging is disabled by default and it can be enabled and directed to
either a file or the screen using public variables or environment
variables.
When enabled, all HTTP requests and responses are logged automatically.

Logging is documented in util/LOGGING.md

Changed in the library: 
* added calls to util processing functions into api.go.
* added util/logging.go

Implements Issue #68.
Updates Issue #67 and partially alleviates problems from Issue #38.

Signed-off-by: Giuseppe Maxia

### DIFF
--- a/util/LOGGING.md
+++ b/util/LOGGING.md
@@ -3,9 +3,11 @@
 
 ## Defaults for logging
 
-Use of the standard Go `log` package is regulated through `logging.go` in the `util` package.
+Use of the standard Go `log` package is cwdeprecated and should be avoided. 
+The recommended way of logging is through the logger `util.GovcdLogger`, which supports [all the functions normally available to `log`](https://golang.org/pkg/log/#Logger).
 
-By default, **logging is disabled**. Any `log.Printf` statement will simply be discarded.
+
+By default, **logging is disabled**. Any `GovcdLogger.Printf` statement will simply be discarded.
 
 To enable logging, you should use
 
@@ -25,16 +27,14 @@ util.ApiLogFileName = "my_file_name.log"
 If you want logging on the screen, use
 
 ```go
-util.LogOnScreen = "stdout"
+util.GovcdLogger.SetOutput(os.Stdout)
 ```
 
 or
 
-```go
-util.LogOnScreen = "stderr"
 ```
-
-Don't forget to call `util.SetLog()` after changing any parameter.
+util.GovcdLogger.SetOutput(os.Stderr)
+```
 
 ## Automatic logging of HTTP requests and responses.
 
@@ -48,14 +48,20 @@ The output of these functions can be quite large. If you want to mute the HTTP p
 ```go
 util.LogHttpRequest = false
 util.LogHttpResponse = false
-util.SetLog()
 ```
 
 During the request and response processing, any password or authentication token found through pattern matching will be automatically hidden. To show passwords in your logs, use
 
 ```go
 util.LogPasswords = true
-util.SetLog()
+```
+
+## Custom logger
+
+If the configuration options are not enough for your needs, you can supply your own logger.
+
+```go
+util.SetCustomLogger(mylogger)
 ```
 
 ## Environment variables
--- a/util/logging.go
+++ b/util/logging.go
@@ -36,15 +36,15 @@ const (
 
 	// Name of the environment variable that enables logging of HTTP responses
 	envLogSkipHttpResp = "GOVCD_LOG_SKIP_HTTP_RESP"
-
-	// Name of the environment variable that limits the payload size
-	envLogMaxPayloadSize = "GOVCD_MAX_PAYLOAD_SIZE"
-
-	// Name of the environment variable that limits the result size
-	envLogMaxMaxResultSize = "GOVCD_MAX_RESULT_SIZE"
 )
 
 var (
+	// All go-vcloud director logging goes through this logger
+	GovcdLogger *log.Logger
+
+	// It's true if we're using an user provided logger
+	customLogging bool = false
+
 	// Name of the log file
 	// activated by GOVCD_LOG_FILE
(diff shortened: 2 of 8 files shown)


govcd/api.go
```diff
@@ -51,6 +53,13 @@ func (c *Client) NewRequest(params map[string]string, method string, u url.URL,
 		req.Header.Add("Accept", "application/*+xml;version="+c.APIVersion)
 	}
 
+	payload := ""
+	if req.ContentLength > 0 {
```
auppunda: Could you put a comment explaining these changes in api.go

util/logging.go
```diff
@@ -0,0 +1,205 @@
...
+	envUseLog = "GOVCD_LOG"
+
+	// Name of the environment variable with the log file name
+	envLogFileName = "GOVCD_LOG_FILE"
+
+	// Name of the environment variable with the screen output
+	envLogOnScreen = "GOVCD_LOG_ON_SCREEN"
+
+	// Name of the environment variable that enables logging of passwords
+	envLogPasswords = "GOVCD_LOG_PASSWORDS"
```
auppunda: Do you set GOVCD_LOG_PASSWORDS to true ?

dataclouder: no. 
This is the name of the environment variable to change the default, which is
```LogPasswords bool = false```

File: util/LOGGING.md [hunk 3]
```diff
@@ -0,0 +1,75 @@
...
+By default, **logging is disabled**. Any `log.Printf` statement will simply be discarded.
+
+To enable logging, you should use 
+
+```go
+util.EnableLogging = true
+util.SetLog()
+```
+
+When enabled, the default output for logging is a file named `govcd-API-YYYY-MM-DD.HH.log`.
```
lvirbalas: I was thinking about the default frequency of the logs generated. We want the defaults to be easy to use and not overload the user. It's now generating a log file each hour, but even if we reduce it to getting a new log each day it's still too noisy for a default.

Please change the default log filename to `go-vcloud-director.log` without date/time reference, but add a write-up that the YYYY-MM-DD, etc. are supported, so the advanced clients know that's available.

BTW, this will make it consistent with the Python vcd CLI default logging.

auppunda:
An idea that I would like the rest of the teams opinion on: Having an env variable to limit payloads. As in no XML payload greater will be dumped

lvirbalas:
@auppunda, if we limit how much of the payload we show, we risk not having enough information when we actually need it, hence it feels cleaner to implement #77 which would allow us to spit out the payload only when an error occurs.

dataclouder:
There is no risk in adding the `GOVCD_MAX_PAYLOAD_SIZE` variable. It's up to the user to use it or not. It doesn't cost much in terms of code (it's already done, in fact.)

hodgesrm:
@auppunda I think we should not add an extra variable to control length of payloads.  I would instead think of ways we can take variables away. (Back to an in important principle in Ousterhout's 'Philosophy of Software Design.')

dataclouder:
@lvirbalas, @auppunda, @hodgesrm 
I threw away the extra variables (by the time Ankith was outvoted I had already added a size limit for response text) and added comments where requested.
I have also added a check to detect whether a payload contains binary data instead of text (using the headers contents.)
The log file is now named `go-vcloud-director.log`.

util/logging.go
```diff
@@ -0,0 +1,242 @@
...
+	LogHttpRequest bool = true
+
+	// Enable logging of Http responses
+	// disabled by GOVCD_LOG_SKIP_HTTP_RESP
+	LogHttpResponse bool = true
+
+	// Sends log to screen. If value is either "stderr" or "err"
+	// logging will go to os.Stderr. For any other value it will
+	// go to os.Stdout
+	LogOnScreen string = ""
```
hodgesrm: It would be more flexible if we allowed clients to pass in their own Logger.  This would allow the SDK to integrate with whatever logging scheme the client uses.   If they don't we use the options to configure a default logger.

auppunda: Are you suggesting for us to create a Logger interface as well, so that clients can add function implementations if they are not satisfied with the default?

auppunda: Ya I think logging is one of those things where everybody has their own preferences, so allowing the client to choose how go-vcloud-director logs is probably a good thing

dataclouder: @hodgesrm if we use a customized logger, we would have control only on what we do within the logging module. My purpose is also to alter the global behavior of `log.anyfunction` so that the output is the same for all the occurrence of logging within the library.
Would it be acceptable to pass instead an `io.writer`?
https://golang.org/pkg/log/#SetOutput

dataclouder: On second thought, I should probably go the extra mile and use a private logger for our module. This will allow our logs to not mix with client logging. Users can change it by giving us their preferred logger.
Will be back soon with a cleaner implementation.

hodgesrm: OK, sounds as if you are on the case.  I also thought about the io.writer approach and even wrote it in my comment and then erased it before pressing Comment. :)  

As far as config vs. environmental variables I notice Kubernetes API clients mostly use a single environmental variable that points to a config file ($KUBECONFIG). It would be worth checking out some other Golang API clients to see what they do.

dataclouder: I have reworked the logging module to use a private logger, and to accept a custom logger from users.
I have also cleaned up the library from all the occurrences of `log.Printf`, changing them to `util.GovcdLogger.Printf`.

I have not, however, introduced a configuration file. I feel that configuration files should belong in the client, not the library. We use a configuration file in our test, which is, for this purpose, a client of the library.

The next iteration on testing (already in progress) will introduce a `log` section in the test configuration file.

An updated commit will come soon.

PR #83, opened 2018-08-31
 Added in refresh functions to org and AdminOrg

auppunda:
I added refresh to the OrgOperations interface. Added in tests for Refresh to in org_test.go


auppunda:
This PR is not neat, closing it

PR #85, opened 2018-08-31
Added in refresh functions to org and AdminOrg

auppunda:
I added refresh to the OrgOperations interface. Added in tests for Refresh to in org_test.go

### DIFF
--- a/govcd/org.go
+++ b/govcd/org.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	types "github.com/vmware/go-vcloud-director/types/v56"
 	"net/url"
-	"reflect"
 	"strconv"
 	"strings"
 )
@@ -19,8 +18,7 @@ import (
 type OrgOperations interface {
 	FindCatalog(catalog string) (Catalog, error)
 	GetVdcByName(vdcname string) (Vdc, error)
-	getHREF() string
-	getClient() *Client
+	Refresh() error
 }
 
 type Org struct {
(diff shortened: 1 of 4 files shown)


auppunda:
@hodgesrm changes made

govcd/org.go
```diff
@@ -48,6 +48,62 @@ func NewAdminOrg(c *Client) *AdminOrg {
 	}
 }
 
+// Generic function that takes in a pointer to a Org or Admin org as an
+// entity and refreshes it by getting the current state from the
+// vcd rest api. Users should use this function after creating,
+// updating, deleting resources under this org. Since the we use
```
hodgesrm: Typo: "the we" should be "we".

govcd/org.go
```diff
@@ -48,6 +48,62 @@ func NewAdminOrg(c *Client) *AdminOrg {
...
+// org links to get elements within the org, users must refresh
+// after performing an operation that can change the links. Otherwise
+// users might get stale metadata, and for users trying to delete an org
+// may run into issues deleting everything.
```
hodgesrm: Hmm.  This does not seem better than the previous implementation which had two copy/paste functions.  It's still effectively two versions because there are a bunch of if statements to break up the control flow--this seems more complex than what you had before. 

Unless there's a way in Go to paper over the differences between the structs and allow a single path for both types, I would revert this back to the previous implementation.

hodgesrm:
@auppunda, before you revert changes have a look at the following article: https://medium.com/@simplyianm/why-gos-structs-are-superior-to-class-based-inheritance-b661ba897c67.  

The way to get what I suggested is have a common type that includes shared struct elements from both Org and AdminOrg.  You can then have a 'method' that operates on this common type.  The trick is that you'll have to use reflection to figure out if you have an Org or AdminOrg so you can tell which to deserialize.  

Let me know if you need more hints.  Golang reflection is described here: https://golang.org/pkg/reflect/.

govcd/org.go
```diff
@@ -49,50 +48,58 @@ func NewAdminOrg(c *Client) *AdminOrg {
...
-func (org *Org) Refresh() error {
-	if *org == (Org{}) || org.Org.HREF == "" {
-		return fmt.Errorf("cannot refresh, Object is empty")
+// Generic function that takes in a pointer to a Org or Admin org as an
+// entity and refreshes it by getting the current state from the
+// vcd rest api. Users should use this function after creating,
+// updating, deleting resources under this org. Since the we use
+// org links to get elements within the org, users must refresh
+// after performing an operation that can change the links. Otherwise
+// users might get stale metadata, and for users trying to delete an org
```
lvirbalas: Thank you for a write-up! Would you add a specific example when this can happen and Refresh should be used?

auppunda:
@lvirbalas Yes we can do an automatic refresh in update to avoid a stale org.

auppunda:
@hodgesrm So I explored composition and I wasn't able to use composition to combine it into one function. What I tried was using the OrgOperations interface to make refresh generic, however I had to add one line get functions to be able to extrapolate the HREF and client from the interface. However, I can add these as parameters to the function so that I can take the getters away. This just makes the function implementation less clean. So right now I am weighing whether I should add these as parameters and make the refresh function less clean or we can leave the getters in the interface api which isn't clean either. Or we can revert back to what I first had.

auppunda:
Also separated admin org and org code so they look like two different things

govcd/org.go
```diff
@@ -32,20 +33,39 @@ func NewOrg(client *Client) *Org {
...
-// AdminOrg includes all members of the Org element, and adds several
-// elements that can be viewed and modified only by system administrators.
-type AdminOrg struct {
-	AdminOrg *types.AdminOrg
-	c        *Client
-}
+// Given a valid catalog name, FindCatalog returns a Catalog object.
+// If no catalog is found, then returns an empty catalog and no error.
+// Otherwise it returns an error.
+func (org *Org) FindCatalog(catalogName string) (Catalog, error) {
```
hodgesrm: 'av' is not descriptive please use a different name.

govcd/org.go
```diff
@@ -75,6 +95,74 @@ func (org *Org) GetVdcByName(vdcname string) (Vdc, error) {
...
+		return fmt.Errorf("error decoding org response: %s", err)
+	}
+	org.Org = unmarshalledOrg
+	// The request was successful
+	return nil
+}
+
+// Given an adminorg with a valid HREF, the function refetches the adminorg
+// and updates the users adminorg data. Otherwise if the function fails,
+// it returns an error.  Users should use refresh whenever they have
```
hodgesrm: Typo: the users -> the user's

govcd/org_test.go
```diff
@@ -9,6 +9,40 @@ import (
...
+	// Test Refresh on admin org
+	err = adminOrg.Refresh()
+	check.Assert(err, IsNil)
+	check.Assert(adminOrg.AdminOrg.FullName, Equals, "govcd")
+	// Delete, with force and recursive true
+	err = adminOrg.Delete(true, true)
+	check.Assert(err, IsNil)
+}
+
 // Creates a org DELETEORG and then deletes it to test functionality of
```
hodgesrm: Typo: 'a org' -> 'an org'

govcd/org.go
auppunda: fixed

Issue #77, opened 2018-08-29
Add log setting to dump failing HTTP request/response values using a ring buffer


hodgesrm:
Diagnosing HTTP call failures often requires detailed logging of the request/response.  The problem is that such logging is either feast or famine:
- HTTP logging is turned off so there's no additional information beyond the error type itself
- HTTP logging is turned on so the log is flooded with information that is useless if there are no errors.  Even when there's an error you need to trek through irrelevant information to get to the error.

As a result users tend to keep logging off and then have to repeat operations with logging on once an error appears.  This is inefficient and does not help with errors that appear intermittently. 

We can address this problem using an in-memory ring buffer to hold the last N request/responses.  If an error occurs we dump the ring buffer as if logging had been turned on.  This gives clients diagnostic information without requiring HTTP logging to be turned on. 

There are some subtleties in the implementation that must be considered.
- If HTTP logging is turned on the ring buffer is disabled.
- The ring buffer keeps entries in memory, so they cannot be large.  Binary data like OVA uploads should not be stored.  
- Clients should be able to control the number of entries to keep and dump in the event of an error. Just keeping the last request is a good default.
- Each Client struct should have its own ring buffer to prevent confusion in concurrent applications.

Golang has a ring implementation that may be helpful for implementation.  See https://golang.org/pkg/container/ring/ for more information.

dataclouder:
@hodgesrm,
I had a look at container/ring, but it does not seem to do what we want. However, I was able to create a fixed sized stack using container/list as storage.

```go
type FixedSizeStack struct{
	list     *list.List
	maxItems int
}

func NewFixedSizeStack(wanted_size int) (fss FixedSizeStack) {
	fss.maxItems = wanted_size
	fss.list = list.New()
	return
}

func (fss *FixedSizeStack) Len() int {
	return fss.list.Len()
}

func (fss *FixedSizeStack) Push(item interface{}) {
	if fss.list.Len() >= fss.maxItems {
		fss.list.Remove(fss.list.Back())
	}
	fss.list.PushFront(item)
}

func (fss *FixedSizeStack) Pop() interface{} {
	latest := fss.list.Front()
	fss.list.Remove(fss.list.Front())
	return latest
}
```
And it does what we want, allowing us to store arbitrary objects.
```go
	type Operation struct {
		Req Request
		Resp Response
	}
	max_elements := 2
	var fss FixedSizeStack =  NewFixedSizeStack(max_elements)

	for {
[...]

PR #94, opened 2018-09-06
Dump failing HTTP operations

dataclouder:
Issue #77 implemented as a Fixed-size stack, based on standard Go double-linked
list. The stack is created with a size limit. When the limit is reached,
the oldest element is purged from the back end.

The stack is enabled when logging is disabled.
When there is a failure, the HTTP processing functions in api.go invoke
the dumping of the stack, which will briefly enable logging and write
down all elements in the stack (by default are 4: two requests and two
responses). The operations are listed in reverse order of execution
(most recent first) with a header that explains that we are emptying the
stack and which element number we are seeing.

Related issue: #77 "Add log setting to dump failing HTTP request/response values using a ring buffer"

Added:
* `fixed-size-stack.go` implements the stack
* `fixed-size-stack_test.go` includes a test for generic data and operations data similar to what we use for HTTP logging.

Changes in existing code:  
* `util.logging` adds integration between logging and FixedSizeStack
* some functions in `api.go` include a call to `util.LogHttpOperations()`

Signed-off-by: Giuseppe Maxia

### DIFF
--- a/util/logging.go
+++ b/util/logging.go
@@ -18,6 +18,17 @@ import (
 	"strings"
 )
 
+// Used by GovcdHttpStack to store operations
+type OperationRec struct {
+	Caller    string
+	Operation string
+	Url       string
+	Data      string
+	OpType    string
+	Req       *http.Request
+	Resp      *http.Response
+}
+
 const (
 	// Name of the environment variable that enables logging
 	envUseLog = "GOVCD_LOG"
@@ -41,6 +52,22 @@ const (
 var (
 	// All go-vcloud director logging goes through this logger
 	GovcdLogger *log.Logger
+
+	// HTTP operation stack
+	// When logging is **disabled**, the stack keeps in memory
+	// several (GovcdHttpStackSize) HTTP operations, to be used
+	// in case of failure.
+	// When there is a failure, main HTTP processing functions in api.go
+	// will call util.LogHttpOperations() to empty the stack into the
+	// default log, which is enabled on-the-fly.
+	GovcdHttpStack FixedSizeStack
+
+	// HTTP operation stack size
+	// 4 operations == two requests and two responses
+	GovcdHttpStackSize int = 4
+
+	// Enable storing HTTP operations
+	EnableHttpStack bool = true
 
 	// It's true if we're using an user provided logger
 	customLogging bool = false
@@ -151,7 +178,7 @@ func isBinary(data string, req *http.Request) bool {
 	return false
 }
 
-// Scand the header for known keys that contain authentication tokens
+// Scans the header for known keys that contain authentication tokens
 // and hide the contents
 func logSanitizedHeader(input_header http.Header) {
 	for key, value := range input_header {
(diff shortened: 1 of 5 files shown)


util/logging.go
```diff
@@ -42,6 +53,16 @@ var (
 	// All go-vcloud director logging goes through this logger
 	GovcdLogger *log.Logger
 
+	// HTTP operation stack
+	GovcdHttpStack FixedSizeStack
```
auppunda: I think you should add a more descriptive comment or explain what this is and why it is useful in logging.md, or both.

util/logging.go
```diff
@@ -42,6 +53,16 @@ var (
...
+
+	// HTTP operation stack size
+	// 4 operations == two requests and two responses
+	GovcdHttpStackSize int = 4
```
auppunda: Is there a way users can set this

util/logging.go
```diff
@@ -42,6 +53,16 @@ var (
...
+
+	// Enable storing HTTP operations
+	EnableHttpStack bool = true
```
auppunda: Is there a way users can set this

util/logging.go
dataclouder: Yes. It's public interface. We may add these parameters to the test configuration, to be used instead of regular logging.

util/logging.go
```diff
@@ -165,6 +186,9 @@ func logSanitizedHeader(input_header http.Header) {
 
 // Logs the essentials of a HTTP request
 func ProcessRequestOutput(caller, operation, url, payload string, req *http.Request) {
+	if EnableHttpStack {
+		GovcdHttpStack.Push(OperationRec{OpType: "request", Caller: caller, Req: req, Data: payload, Url: url, Operation: operation})
```
auppunda: Does this use log santitized header? If it doesn't, doesn't it have the potential to log confidential passwords

util/logging.go
dataclouder: Yes. I am adding some info to LOGGING.md

util/logging.go
dataclouder: The header is sanitized when the HTTP operation is dumped, as it uses the same function as the regular logging.

util/logging.go
dataclouder: Will do, and add some more info in TESTING.md

util/logging.go
auppunda: Doesn't it call this function later in the method though. Wouldn't those calls be redundant then.

dataclouder: When we dump the HTTP records, `EnableHttpStack` gets disabled, to avoid an endless loop.

auppunda: ahh ok that makes more sense

dataclouder:
Added documentation in the code and in LOGGING.md

dataclouder:
> How do you actually dump the failed response to the user?

As the issue says:  "_If an error occurs we dump the ring buffer as if logging had been turned on_. "
So, it goes to the predetermined log file (or screen, or custom logger, as defined by the user).

auppunda:
@dataclouder Ah I see it now. Before I didn't see changes to api.go.

auppunda:
LGTM

dataclouder:
Thanks for your review.
I have realized that the functional test is needed. 
I will re-submit after finishing with the tests-related issues that I am currently addressing.

dataclouder:
Closing this PR for the time being. Needs more work than anticipated. Will re-submit later

Issue #33, opened 2018-08-17
Clean-up code of one-letter identifiers


dataclouder:
The whole code base is littered with identifiers (variables, struct types, and types) made of a single letter.

To make the library code more readable, and easier to maintain and expand, these identifiers must be replaced with more telling names. 

The only place where single letter variables can be accepted is for a *counter* in a *short* loop.

Moreover, variables and field name `vdc` and `vcd` should be avoided as they are easily mixed up, and replaced with `vcDirector` and `vdCenter` or similarly telling names.

dataclouder:
Comment from @lvirbalas:
>vCD and vDC are such core abbreviations in vCloud Director, that people just get used to the one
> letter difference between the two. Same in code, for a person who is working with vCD for years
> (and that is our customers), "vcd" and "vdc" are less confusing than the suggested "vcdir" and
> "vdcent". BTW, a common longer abbreviation in the wide open internet for "vcd" is "vcloud".
> As there's potentially more harm than benefit of changing these abbreviations, let's keep using "vcd" and "vdc".

PR #95, opened 2018-09-07
Replace all one-letter identifiers

dataclouder:
Implements Issue #33

All one-letter struct fields or variable identifiers have been replaced by
more telling names. I have left only a few one-letter counters in short
loops.

There are no behavioral changes or additions in this patch. Only renames.

I have also removed a few leftover XML examples from the tests.

Signed-off-by: Giuseppe Maxia


govcd/api.go
```diff
@@ -29,28 +29,28 @@ type Client struct {
 
 // NewRequest creates a new HTTP request and applies necessary auth headers if
 // set.
-func (c *Client) NewRequest(params map[string]string, method string, u url.URL, body io.Reader) *http.Request {
+func (cli *Client) NewRequest(params map[string]string, method string, reqUrl url.URL, body io.Reader) *http.Request {
 
-	p := url.Values{}
+	values := url.Values{}
```
lvirbalas: Could be reqValues to match code style below.

Issue #93, opened 2018-09-05
Add check for empty Org/Vdc in SetupTestSuite


auppunda:
Since the Get functions changed format (issue #69), we need to add an extra check to check if the org/vdc is empty. If empty we need to panic to prevent a segmentation fault.
```
	// set org
	vcd.org, err = GetOrgByName(vcd.client, config.VCD.Org)
	if err != nil {
		panic(err)
	}
	// set vdc
	vcd.vdc, err = vcd.org.GetVdcByName(config.VCD.Vdc)
	if err != nil {
		panic(err)
	}
```
Change these to check for vcd.org == (Org{}) and vcd.vdc == (Vdc{}). If this is the case that means that no org/vdc was found, so we have to panic.

Issue #92, opened 2018-09-05
Adding System check in test Suite


auppunda:
Basically when performing admin org tests check if the provider org is System. If it isn't skip those tests.

Issue #87, opened 2018-08-31
Clean up all resources created during Tests in case of failure


auppunda:
We should add to our cleanup suite in api_vcd.go a check of all resources we have created during the test run. If anything is left remaining we should delete it. This allows users to rerun the tests if they fail without deleting the resource the tests created. Right now we only do that for the vapp that we create in the beginning of testing.

PR #96, opened 2018-09-10
Add cleanup function to test suite

dataclouder:
* Implements Issue #87 : Clean up all resources created during tests in case of failure. The cleanup is implemented using an array of entities created during the tests. After a successful creation, the entity is added to the list. At the end of the tests, the cleanup function checks whether the entity still exists, and removes it if it still in the system.

* Implements Issue #92 : Adding System check in test Suite

* Implements Issue #93 : Add check for empty Org/Vdc in `SetupTestSuite`

Changes in this pull request:
* Added a call to `AddToCleanupList` after all successful creation of a new entity in the test operations;
* Replaced the existing `TearDownSuite` with the handling of the cleanup list.
* Added a `VerboseCleanup` field to the test configuration options
* Added checks for empty entities in `SetupTestSuite`
* Added test skip statements for when the provided "sysOrg" is not "System"

### DIFF
--- a/govcd/api_vcd_test.go
+++ b/govcd/api_vcd_test.go
@@ -13,6 +13,22 @@ import (
 	"testing"
 )
 
+const (
+	// Names for entities created by the tests
+	TestCreateOrg          = "TestCreateOrg"
+	TestDeleteOrg          = "TestDeleteOrg"
+	TestUpdateOrg          = "TestUpdateOrg"
+	TestCreateCatalog      = "TestCreateCatalog"
+	TestCreateCatalogDesc  = "Catalog created by tests"
+	TestRefreshOrgFullName = "TestRefreshOrgFullName"
+	TestUpdateCatalog      = "TestUpdateCatalog"
+	TestDeleteCatalog      = "TestDeleteCatalog"
+	TestRefreshOrg         = "TestRefreshOrg"
+	TestComposeVapp        = "TestComposeVapp"
+	TestComposeVappDesc    = "vApp created by tests"
+	TestSetUpSuite         = "TestSetUpSuite"
+)
+
 // Struct to get info from a config yaml file that the user
 // specifies
 type TestConfig struct {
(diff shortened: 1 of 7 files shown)


govcd/api_vcd_test.go
```diff
@@ -13,6 +13,22 @@ import (
 	"testing"
 )
 
+const (
+	// Names for entities created by the tests
+	OrgCreateTest            = "CreateOrgTest"
+	OrgDeleteTest            = "DeleteOrgTest"
+	OrgUpdateTest            = "UpdateOrgTest"
+	CatalogCreateTest        = "CreateCatalogTest"
```
lvirbalas: One suggestion for improvement. The purpose of this ask is to enable better readability and easier troubleshooting.

Right now the order of the words is not consistent (e.g. OrgCreateTest vs. CreateOrgTest). Suggestion is to rename the variables and the values to better match the test functions themselves as in the example below.

Function:
`Test_CreateCatalog(check *C)`
Change from:
`CatalogCreateTest        = "CreateCatalogTest"`
Change to:
`TestCreateCatalog        = "TestCreateCatalog"`

Having "Test" at the beginning of entity names should also proof more clear when looking at the vCD instance on which the tests are running right now. You would instantly see that this is a test.

dataclouder:
Fixed constant names and their contents, using the function name where they are used as base.

lvirbalas:
Thanks! LGTM. Please get one more LGTM from @vbauzysvmware and merge.

govcd/api_vcd_test.go
```diff
@@ -150,57 +181,144 @@ func (vcd *TestVCD) SetUpSuite(check *C) {
...
+func (vcd *TestVCD) removeLeftoverEntities(entity CleanupEntity) {
+	var introMsg string = "removeLeftoverEntries: [INFO] Attempting cleanup of %s '%s' instantiated by %s\n"
+	var notFoundMsg string = "removeLeftoverEntries: [INFO] No action for %s '%s'\n"
+	var removedMsg string = "removeLeftoverEntries: [INFO] Removed %s '%s' created by %s\n"
+	var notDeletedMsg string = "removeLeftoverEntries: [ERROR] Error deleting %s '%s': %s\n"
+	// NOTE: this is a cleanup function that should continue even if errors are found.
+	// For this reason, the [ERROR] messages won't be followed by a program termination
+	vcd.infoCleanup(introMsg, entity.Name, entity.EntityType, entity.CreatedBy)
+	switch entity.EntityType {
+	case "vapp":
```
vbauzysvmware: I am only concern here is that method soon will be huge, cause all the cases will have they own code. So general method readability what it does will degrade.

vbauzysvmware: Great

PR #97, opened 2018-09-14
Rename util.GovcdLogger to util.Logger

dataclouder:
Shortened logger name as requested

Signed-off-by: Giuseppe Maxia

There was no issue for this PR.


dataclouder:
We need to call it Logger, because it is a logger, not a generic log.
We want to keep the semantics clear to the fact that users can provide their own logger.  We agreed to this in the latest call.

Issue #50, opened 2018-08-23
Upload a Catalog Item


lvirbalas:
We need ability to upload OVAs to the catalog. The implementation is not trivial, because they way vCD API works. You first have to extract the OVA, because vCD first needs the OVF description which is inside. Then vCD returns a set of links (which represent the `References->File` entries in the OVF XML), to which you upload each file separately.

In addition to the basic OVA upload, the following is also needed:

1. Support chunked OVA files.
2. Wait for the vCD to finish the import after uploading (this is above what `pyvcloud` does Today).
3. Utilise existing Task mechanism.
4. Report percentage progress.

vbauzysvmware:
Things todo:
    Error handling return within xml of response
    Utilise existing Task mechanism.
    Report percentage progress
    Remove lock to xml first element
    Better error handling
    Missing tests
    Verify that "sanitizedName" func is enough to cover security issue -> See vmware/pyvcloud#268.

vbauzysvmware:
Additional things after trying upload with progress:
Check against nil catalog
Investigate how it works with admin org catalog (works with standard user catalog)
Clean temp folder
Investigate to use system temp folder(It's possible that the directory containing the OVA is not writable, or users may not want to write there. We need to check whether the directory exists, and remove old files. Finally, we need to remove the files after an upload (successful or not)  we need to compare the file size 
Check files: from the tarball header with the size we have extracted

vbauzysvmware:
Check if multiupload to same instance is working.

vbauzysvmware:
> Two more checks:
> 
>     * [IMPORTANT] Look whether the catalog item already exists, so that we will avoid cryptic error messages when we try to upload;
> 
>     * Recover from a previous failure, when the catalog item is created, but we couldn't upload anything because of some error. [Optional, as if everything is checked correctly, we should not find ourselves in this predicament.]
I added check against existing existing item name as user won't understand 500 error with "contact administrator" message. IMHO, there is not point having recovering procedure of previous failure as we won't be able to see if it was govcd or any system or user manual tasks which hanging or failed. So deleting item with error or which is waiting is up to user to manage. API will return error that item already exist.

vbauzysvmware:
Regarding multi upload: works fine as progress variable is local function variable and not shared between uploads. Mutex or other solution isn't needed as progress not shareable.

vbauzysvmware:
[...]

PR #98, opened 2018-09-25
Catalog upload feature2

vbauzysvmware:
Second part of missing functionality Related issue: (#50)

Enhanced upload functionality for catalog item. Returns special upload task which allows monitor upload if required and later can wait until vCD import task finishes. Upload happens in background. Other improvements:
Added Missing tests
Code improvements
Removed lock to xml first element
Better error handling
Verified and improved that "sanitizedName" func is enough to cover security issue -> See vmware/pyvcloud#268.
Use new logging system
Error handling return within xml of response
Checked and parse errors when vCD fails after ovf file upload.
Multiupload verified.
Report percentage progress.
Check against nil catalog
Files unarchived to temp  folder and cleaned
Checked files: from the tarball header with the size we have extracted
Make upload chunking working as function parameter
Specific check to avoid cryptic error then item already exist. 

Note: Exist one question how to handle between AdminCatalog and Catalog object if needed. Current implementation allows working from Catalog object perspective.


govcd/catalog.go
```diff
@@ -351,52 +406,77 @@ func uploadOvfDescription(client *Client, ovfFile string, ovfUploadUrl *url.URL)
...
 			uploadURL, err := url.ParseRequestURI(item.HREF)
 			if err != nil {
 				return nil, err
 			}
 
+			util.Logger.Printf("[TRACE] findCatalogItemUploadLink - upload url found: %s \n", uploadURL)
 			return uploadURL, nil
 		}
 	}
 	return nil, errors.New("catalog upload url isn't found")
```
hodgesrm: Message should be:  "Catalog upload URL not found"

govcd/catalog.go
```diff
@@ -351,52 +406,77 @@ func uploadOvfDescription(client *Client, ovfFile string, ovfUploadUrl *url.URL)
...
+func getExistingCatalogItems(catalog *Catalog) (catalogItemNames []string) {
+	for _, catalogItems := range catalog.Catalog.CatalogItems {
+		for _, catalogItem := range catalogItems.CatalogItem {
+			catalogItemNames = append(catalogItemNames, catalogItem.Name)
+		}
+	}
+	return
+}
+
+func uploadFile(client *Client, uploadLink, filePath string, uploadedBytes, fileSizeToUpload int64, uploadPieceSize int64, callBack func(bytesUpload, totalSize int64)) (int64, error) {
```
hodgesrm: This needs more commentary to understand the protocol.  I can't tell what this code is really promising to do.

govcd/catalog_test.go
```diff
@@ -84,3 +96,179 @@ func (vcd *TestVCD) Test_DeleteCatalog(check *C) {
 	check.Assert(catalog, Equals, Catalog{})
 
 }
+
```
hodgesrm: Please add a comment to each test case to explain the feature it is validating.

govcd/catalog_test.go
```diff
@@ -84,3 +96,179 @@ func (vcd *TestVCD) Test_DeleteCatalog(check *C) {
...
+func (vcd *TestVCD) Test_UploadOvf_progress_works(check *C) {
+	catalogName := TestCreateCatalog + "3"
+	itemName := "item3"
+	setupCatalog(vcd, check, catalogName)
+
+	catalog, org := findCatalog(vcd, check, catalogName)
+	AddToCleanupList(itemName, "catalogItem", vcd.org.Org.Name, "Test_UploadOvf")
+
+	//execute
```
hodgesrm: Unnecessary comment.  (Same applies to verify below.)

govcd/uploadtask.go
```diff
@@ -0,0 +1,33 @@
+package govcd
```
hodgesrm: Needs a proper Apache 2 header.  Please copy from another file.  Also any file you touch should have the copyright date 2018 added.

govcd/uploadtask.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+	"time"
+)
+
+type UploadTask struct {
+	uploadProgress *float64
+	*Task
+}
+
+func NewUploadTask(task *Task, uploadProgress *float64) *UploadTask {
```
hodgesrm: Please add a comment to document public APIs in this file.

util/tar.go
```diff
@@ -4,103 +4,123 @@ import (
...
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
 )
 
-func Unpack(tarFile string) ([]string, error) {
+const TmpDirPrefix = "govcd"
+
+//extract files to system tmp dir with name govcd+random number. Created folder with files isn't deleted.
```
hodgesrm: What is the output of this function?  Also 'extract' should be capitalized.

util/tar_test.go
```diff
@@ -0,0 +1,24 @@
+package util
+
+import (
+	"testing"
+)
+
+func TestSanitizedName(t *testing.T) {
```
hodgesrm: Please a comment describing what this test is checking.

govcd/catalog_test.go
```diff
@@ -84,3 +96,179 @@ func (vcd *TestVCD) Test_DeleteCatalog(check *C) {
...
+	catalog, err := org.FindCatalog(testCreateCatalog)
+	check.Assert(err, IsNil)
+	return catalog, org
+}
+
+func setupCatalog(vcd *TestVCD, check *C, testCreateCatalog string) {
+	org := getOrg(vcd, check)
+
+	// creating catalog
+	adminCatalog, err := org.CreateCatalog(testCreateCatalog, TestCreateCatalogDesc, true)
```
dataclouder: If this test is using an AdminCatalog, it should check whether `skipAdminTests` is set, and skip the test if it is.

govcd/catalog.go
```diff
@@ -125,7 +128,7 @@ type Envelope struct {
 	} `xml:"References>File"`
 }
 
-// If catalogitem is a valid CatalogItem and the call succeds,
+// If catalog item is a valid CatalogItem and the call succeeds,
 // then the function returns a CatalogItem. If the item does not
 // exist, then it returns an empty CatalogItem. If The call fails
```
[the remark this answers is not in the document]
vbauzysvmware: fixed

govcd/catalog.go
```diff
@@ -163,37 +166,47 @@ func (cat *Catalog) FindCatalogItem(catalogitem string) (CatalogItem, error) {
 
 // uploads an ova file to a catalog. This method only uploads bits to vCD spool area.
```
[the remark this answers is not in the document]
vbauzysvmware: fixed

govcd/catalog.go
```diff
@@ -163,37 +166,47 @@ func (cat *Catalog) FindCatalogItem(catalogitem string) (CatalogItem, error) {
...
 // Returns errors if any occur during upload from vCD or upload process.
```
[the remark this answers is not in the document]
vbauzysvmware: Added

govcd/catalog.go
```diff
@@ -163,37 +166,47 @@ func (cat *Catalog) FindCatalogItem(catalogitem string) (CatalogItem, error) {
...
 	//	3. Start uploading bits to the transfer folder
 	//	4. Wait on the import task to finish on vCD side -> task success = upload complete
 
+	if *cat == (Catalog{}) {
+		return UploadTask{}, errors.New("catalog can not be empty or nil")
+	}
+
+	for _, catalogItemName := range getExistingCatalogItems(cat) {
+		if catalogItemName == itemName {
+			return UploadTask{}, fmt.Errorf("catalog item by name: '%s' already exist. Upload with different name", itemName)
```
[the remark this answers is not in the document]
vbauzysvmware: Fixed

govcd/catalog.go
```diff
@@ -202,58 +215,90 @@ func (cat *Catalog) UploadOvf(ovaFileName, itemName, description string, chunkSi
...
-		return Task{}, err
+		return UploadTask{}, err
 	}
 
+	var uploadProgress float64
+	callBack := func(bytesUploaded, totalSize int64) {
+		uploadProgress = (float64(bytesUploaded) / float64(totalSize)) * 100
+	}
+
+	go uploadFiles(cat.client, vappTemplate, &ovfFileDesc, tempPath, filesAbsPaths, uploadPieceSize, callBack, tmpDir)
```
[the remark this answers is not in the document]
vbauzysvmware: Added

govcd/catalog.go
```diff
@@ -331,15 +381,20 @@ func queryVappTemplate(client *Client, vappTemplateUrl *url.URL) (*types.VAppTem
...
+		if "error" == task.Status && newItemName == task.Owner.Name {
+			util.Logger.Printf("[Error] %#v", task.Error)
+			return vappTemplateParsed, fmt.Errorf("Error in vcd returned error code: %d, error: %s and message: %s ", task.Error.MajorErrorCode, task.Error.MinorErrorCode, task.Error.Message)
+		}
+	}
+
 	return vappTemplateParsed, nil
 }
 
 // Uploads ovf description file from unarchived provided ova file. As result vCD will generate temporary upload links which has to be queried later.
```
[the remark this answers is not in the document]
vbauzysvmware: Fixed

govcd/catalog.go
vbauzysvmware: Fixed

govcd/catalog.go
vbauzysvmware: Added.

govcd/catalog_test.go
vbauzysvmware: Added.

govcd/catalog_test.go
vbauzysvmware: Removed.

govcd/uploadtask.go
vbauzysvmware: Added

govcd/uploadtask.go
vbauzysvmware: Added.

util/tar.go
vbauzysvmware: Done.

util/tar_test.go
vbauzysvmware: Added

govcd/catalog_test.go
```diff
@@ -5,7 +5,19 @@
...
+	"log"
+	"os"
+	"path"
+	"runtime"
+	"strings"
+)
+
+const (
+	OvaPath        = "/test_vapp_template.ova"
+	OvaChunkedPath = "/template_with_custom_chunk_size.ova"
```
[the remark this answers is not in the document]
vbauzysvmware: 1. I would agree with https://medium.com/@matryer/5-simple-tips-and-tricks-for-writing-unit-tests-in-golang-619653f90742 and move all test to different package - in that away we can avoid more clutter in future and separate stuff. And solve issue mentioned by you as binary files would be in different package. Otherwise we create issue that test exist but they fail for others if they won't checkout special folder with binary files. 

2. Yes, I can move constants to api_vcd_test.go

govcd/catalog_test.go
```diff
@@ -84,3 +96,179 @@ func (vcd *TestVCD) Test_DeleteCatalog(check *C) {
 	check.Assert(catalog, Equals, Catalog{})
 
 }
+
+func (vcd *TestVCD) Test_UploadOvf(check *C) {
+	checkUploadOvf(vcd, check, getCurrentPath()+OvaPath, TestCreateCatalog, "item1")
```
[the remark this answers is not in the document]
vbauzysvmware: ok, I missed this - will correct it.

govcd/catalog_test.go
```diff
@@ -84,3 +96,179 @@ func (vcd *TestVCD) Test_DeleteCatalog(check *C) {
...
+}
+
+func (vcd *TestVCD) Test_UploadOvf_chuncked(check *C) {
+	checkUploadOvf(vcd, check, getCurrentPath()+OvaChunkedPath, TestCreateCatalog+"2", "item2")
+}
+
+func (vcd *TestVCD) Test_UploadOvf_progress_works(check *C) {
+	catalogName := TestCreateCatalog + "3"
+	itemName := "item3"
```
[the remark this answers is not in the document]
vbauzysvmware: It's test independence question. Test can't depend on other tests result (e.g. what id did with catalog in previous test).

govcd/catalog_test.go
```diff
@@ -84,3 +96,179 @@ func (vcd *TestVCD) Test_DeleteCatalog(check *C) {
...
+	check.Assert(catalog.Catalog.CatalogItems[0].CatalogItem[0].Name, Equals, itemName)
+}
+
+func (vcd *TestVCD) Test_UploadOvf_ShowProgress_works(check *C) {
+	catalogName := TestCreateCatalog + "4"
+	itemName := "item4"
+	setupCatalog(vcd, check, catalogName)
+
+	catalog, org := findCatalog(vcd, check, catalogName)
+	AddToCleanupList(itemName, "catalogItem", vcd.org.Org.Name, "Test_UploadOvf")
```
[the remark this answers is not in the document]
vbauzysvmware: I checked and think we can avoid it as catalog cleanup allows to delete catalog with items. So removing AddToCleanupList should be sufficient.

govcd/catalog_test.go
vbauzysvmware: Will add.

govcd/catalog_test.go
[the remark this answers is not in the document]
dataclouder: We have already tests that depend on user defined artifacts. This test should be no different. It may pass with the artifacts that we recommend, but we must see how it behaves with user defined items.

govcd/catalog_test.go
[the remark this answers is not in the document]
dataclouder: If more catalogs are needed, we should name them according to their purposes. 
In general, we should make tests that work with the material provided by users (i.e the catalog defined in the configuration file).
We should create new catalogs only when testing something relating to catalog creation or if we run an operation that requires two catalogs. Otherwise, we should always use the catalog given by users.

govcd/catalog_test.go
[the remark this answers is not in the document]
dataclouder: No. All the tests are removing their artifacts. The cleanup list is the safety net to remove things when the test fails and aborts prematurely. 
We should keep it and update the cleanup function accordingly.

govcd/catalog_test.go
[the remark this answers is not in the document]
vbauzysvmware: Move binaries to test-resources. Ova paths are configured if empty, tests are skipped.

govcd/catalog_test.go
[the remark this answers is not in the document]
vbauzysvmware: Fixed

govcd/catalog_test.go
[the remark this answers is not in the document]
vbauzysvmware: CatalogItem cleanup added.

govcd/catalog_test.go
vbauzysvmware: this isn't needed cause tests rung fine with org user or admin.

govcd/catalog_test.go
```diff
@@ -84,3 +89,181 @@ func (vcd *TestVCD) Test_DeleteCatalog(check *C) {
...
+
+func getOrg(vcd *TestVCD, check *C) AdminOrg {
+	// Fetching organization
+	org, err := GetAdminOrgByName(vcd.client, vcd.org.Org.Name)
+	check.Assert(org, Not(Equals), AdminOrg{})
+	check.Assert(err, IsNil)
+	return org
+}
+
+func skipThenOvaPathMissing(vcd *TestVCD, check *C) {
```
dataclouder: This should be called `skipWhenOvaPathMissing`

govcd/catalog_test.go
```diff
@@ -84,3 +89,181 @@ func (vcd *TestVCD) Test_DeleteCatalog(check *C) {
...
+		}
+	}
+	return count
+}
+
+func checkUploadOvf(vcd *TestVCD, check *C, ovaFileName, catalogName, itemName string) {
+	//setupCatalog(vcd, check, testCreateCatalog)
+
+	catalog, org := findCatalog(vcd, check, vcd.config.VCD.Catalog.Name)
+	AddToCleanupList(itemName, "catalogItem", vcd.org.Org.Name+"|"+vcd.config.VCD.Catalog.Name, "Test_UploadOvf")
```
dataclouder: The call to `AddToCleanupList` should only happen *after* the successful creation of an item.

govcd/catalog.go
```diff
@@ -202,58 +223,94 @@ func (cat *Catalog) UploadOvf(ovaFileName, itemName, description string, chunkSi
 	for _, filePath := range filesAbsPaths {
```
dataclouder: We are assuming here that the extracted file contains a .ovf.
If this is not true, the procedure hangs. This check needs to happen before the item is created.

govcd/api_vcd_test.go
```diff
@@ -293,7 +300,38 @@ func (vcd *TestVCD) removeLeftoverEntities(entity CleanupEntity) {
...
+						return
+					}
+					err = catalogItemApi.Delete()
+					if err != nil {
+						vcd.infoCleanup(notDeletedMsg, entity.EntityType, entity.Name, err)
+					}
+				}
+			}
+		}
+		vcd.infoCleanup(removedMsg, entity.EntityType, entity.Name, entity.CreatedBy)
```
dataclouder: Something is missing in this procedure. When I use a non existing OVA in the test, it still prints `Removed catalogItem '<ITEM_NAME>' created by Test_UploadOvf`.

govcd/catalog.go
```diff
@@ -161,99 +165,148 @@ func (cat *Catalog) FindCatalogItem(catalogitem string) (CatalogItem, error) {
...
+	util.Logger.Printf("[TRACE] Upload finished and task for vcd import created. \n")
+
+	return *uploadTask, nil
 }
 
-func uploadFiles(client *Client, vappTemplate *types.VAppTemplate, ovfFileDesc *Envelope, tempPath string, filesAbsPaths []string) error {
+// Upload files for vCD created upload links. Different approach then vmdk file are
+// chunked (e.g. test.vmdk.000000000, test.vmdk.000000001 or test.vmdk). vmdk files are chunked if
+// in description file attribute ChunkSize is not zero.
+func uploadFiles(client *Client, vappTemplate *types.VAppTemplate, ovfFileDesc *Envelope, tempPath string, filesAbsPaths []string, uploadPieceSize int64, callBack func(bytesUpload, totalSize int64)) error {
```
dataclouder: This function has too many positional arguments.
Please add documentation about the meaning of every argument. Consider doing the same for every function that has more than three args.

Issue #99, opened 2018-09-27
logging of http request may empty input data


dataclouder:
The logging of HTTP requests in api.go may empty the data when the body is not a `*bytes.Buffer`.

Most of the requests in the API are created using a `*bytes.Buffer`, but some aren't. When the body is some other kind of `io.Reader`, copying the contents will empty the buffer, and the request will fail.

To fix it, we will check whether the body type is `*bytes.Buffer` and make the copy only in that case.

PR #100, opened 2018-09-27
Fixes Issue #99 error in logging of HTTP requests

dataclouder:
Logging of HTTP requests was removing data when the body object was
not what we thought it was.
There is now a type check to ensure that we don't empty the buffer
when it is not safe.

Related Issue: #99 

Also added unwanted items to .gitignore

Signed-off-by: Giuseppe Maxia

### DIFF
--- a/govcd/api.go
+++ b/govcd/api.go
@@ -6,16 +6,16 @@
 package govcd
 
 import (
+	"bytes"
 	"encoding/xml"
 	"fmt"
+	"github.com/vmware/go-vcloud-director/types/v56"
+	"github.com/vmware/go-vcloud-director/util"
 	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
-
-	"bytes"
-	types "github.com/vmware/go-vcloud-director/types/v56"
-	"github.com/vmware/go-vcloud-director/util"
+	"reflect"
 )
 
 // Client provides a client to vCloud Director, values can be populated automatically using the Authenticate method.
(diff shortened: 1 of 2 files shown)


govcd/api.go
```diff
@@ -59,9 +58,19 @@ func (cli *Client) NewRequest(params map[string]string, method string, reqUrl ur
...
+			var ibody interface{}
+			ibody = body
+			bbody, ok := ibody.(*bytes.Buffer)
+			// If the inner object is a bytes.Buffer, we get a safe copy of the data.
+			// If it is really just an io.Reader, we don't, as the copy would empty the reader
+			if ok {
+				payload = bbody.String()
+			} else {
+				// With this content, we'll know that the payload is not really empty, but
+				// it was unavailable due to the bidy type.
```
lvirbalas: Typo in the "bidy".

govcd/api.go
```diff
@@ -59,9 +58,19 @@ func (cli *Client) NewRequest(params map[string]string, method string, reqUrl ur
...
+				payload = "<N/A>"
```
lvirbalas: "N/A" can be interpreted as Not Available. Let's be more specific and tell the real issue. For example, "Other Type". With reflection you could even print which type it holds.

govcd/api.go
```diff
@@ -59,9 +58,19 @@ func (cli *Client) NewRequest(params map[string]string, method string, reqUrl ur
 		// to the processing function.
 		payload := ""
 		if req.ContentLength > 0 {
-			buf := new(bytes.Buffer)
-			buf.ReadFrom(body)
-			payload = buf.String()
+			// We try to convert body to a *bytes.Buffer
+			var ibody interface{}
+			ibody = body
+			bbody, ok := ibody.(*bytes.Buffer)
```
lvirbalas: You could use a reflection here instead to check the type.

govcd/api.go
dataclouder: "Other type" may confuse who is reading. I just want to say that it is not available. What about:
"<Not retrieved for logging>" ?

govcd/api.go
dataclouder: I could. The advantage would be that I could log the exact type so that I will have data for future improvements. Do we need to?

dataclouder: Addressed all the requests

PR #101, opened 2018-10-01
Update or add copyright to all code files

dataclouder:
* Add copyright in files that were without
* Update year for existing copyright
* Fix minor error in types.go
* Add a script to detect lack of copyright in new code files

Signed-off-by: Giuseppe Maxia
Note: copyright in tar.go will be updated as part of PR #98

### DIFF
--- a/scripts/copyright_check.sh
+++ b/scripts/copyright_check.sh
@@ -1,6 +1,11 @@
 #!/bin/bash
+# This script will find code files that don't have a copyright notice
+
+# This check will find files with a copyright but not for the current year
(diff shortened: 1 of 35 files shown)


scripts/copyright_check.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
lvirbalas: Need a description/comment about this script at the top.

dataclouder: Added description to copyright check script.

PR #113, opened 2018-10-23
Catalog upload fixes

vbauzysvmware:
Added function to support ability to get vCD task progress and fixed multi disk ova's upload progress calculations.

### DIFF
--- a/govcd/catalog.go
+++ b/govcd/catalog.go
@@ -146,7 +146,7 @@ func (cat *Catalog) FindCatalogItem(catalogitem string) (CatalogItem, error) {
 
 				resp, err := checkResp(cat.client.Http.Do(req))
 				if err != nil {
-					return CatalogItem{}, fmt.Errorf("error retrieving catalog: %s", err)
+					return CatalogItem{}, fmt.Errorf("error retreiving catalog: %s", err)
 				}
 
 				cat := NewCatalogItem(cat.client)
@@ -279,6 +279,7 @@ func (cat *Catalog) UploadOvf(ovaFileName, itemName, description string, uploadP
 // uploadPieceSize - size of chunks in which the file will be uploaded to the catalog.
 // callBack a function with signature //function(bytesUpload, totalSize) to let the caller monitor progress of the upload operation.
 func uploadFiles(client *Client, vappTemplate *types.VAppTemplate, ovfFileDesc *Envelope, tempPath string, filesAbsPaths []string, uploadPieceSize int64, callBack func(bytesUpload, totalSize int64)) error {
+	var uploadedBytes int64
 	for _, item := range vappTemplate.Files.File {
 		if item.BytesTransferred == 0 {
 			number, err := getFileFromDescription(item.Name, ovfFileDesc)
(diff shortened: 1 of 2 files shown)


govcd/catalog.go
```diff
@@ -254,7 +254,7 @@ func (cat *Catalog) UploadOvf(ovaFileName, itemName, description string, uploadP
 	var task Task
 	for _, item := range vappTemplate.Tasks.Task {
 		task, err = createTaskForVcdImport(cat.client, item.HREF)
-		if err != nil {
+		if err != nil || task.Task.Status == "error" {
```
dataclouder: If the error is nil but the task reports an error, we are hiding the error message from the user.
There should be a separate check for the task status and return accordingly.

govcd/catalog.go
```diff
@@ -501,8 +514,10 @@ func getExistingCatalogItems(catalog *Catalog) (catalogItemNames []string) {
 // uploadedBytes - how much of file already uploaded
 // fileSizeToUpload - how much bytes will be uploaded
 // uploadPieceSize - size of chunks in which the file will be uploaded to the catalog.
+// uploadedBytesForCallback all uploaded bytes if multi disk in ova
+// allFilesSize overall sum of size if multi disk in ova
 // callBack a function with signature //function(bytesUpload, totalSize) to let the caller monitor progress of the upload operation.
-func uploadFile(client *Client, uploadLink, filePath string, uploadedBytes, fileSizeToUpload int64, uploadPieceSize int64, callBack func(bytesUpload, totalSize int64)) (int64, error) {
+func uploadFile(client *Client, uploadLink, filePath string, uploadedBytes, fileSizeToUpload int64, uploadPieceSize int64, uploadedBytesForCallback, allFilesSize int64, callBack func(bytesUpload, totalSize int64)) (int64, error) {
```
dataclouder: `uploadFile` and `uploadPartFile` have definitely too many arguments to pass by position. This makes reading the code unnecessarily difficult.
It would be better to create a structure to pass, using explicit field names. It would improve readability and facilitate debugging.

govcd/catalog.go
vbauzysvmware: Good spot - tx, fixed.

govcd/catalog.go
vbauzysvmware: Changed.

govcd/catalog.go
```diff
@@ -146,7 +159,7 @@ func (cat *Catalog) FindCatalogItem(catalogitem string) (CatalogItem, error) {
 
 				resp, err := checkResp(cat.client.Http.Do(req))
 				if err != nil {
-					return CatalogItem{}, fmt.Errorf("error retrieving catalog: %s", err)
+					return CatalogItem{}, fmt.Errorf("error retreiving catalog: %s", err)
```
lvirbalas: Spelling error, should be "retrieving", i.e. as in previous commit :)

govcd/catalog.go
```diff
@@ -279,6 +296,7 @@ func (cat *Catalog) UploadOvf(ovaFileName, itemName, description string, uploadP
 // uploadPieceSize - size of chunks in which the file will be uploaded to the catalog.
```
lvirbalas: Please consider renaming to `uploadChunkSize` in all places

vbauzysvmware: I renamed it to piece due that we have already chunkedOVA which means that vmkds inside are splitted. And upload splitting is call pieceSize to misinterpret with chunkedOVA.

govcd/catalog.go
vbauzysvmware: done.

govcd/catalog.go
```diff
@@ -46,6 +46,19 @@ type Catalog struct {
 	client  *Client
 }
 
+// uploadLink - vCD created temporary upload link
+// uploadedBytes - how much of file already uploaded
+// fileSizeToUpload - how much bytes will be uploaded
+// uploadPieceSize - size of chunks in which the file will be uploaded to the catalog.
```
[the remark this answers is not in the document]
vbauzysvmware: Comment Regarding naming in lower comment response.

govcd/catalog.go
lvirbalas: Ah! Got it!

PR #114, opened 2018-10-23
Add comments and change network creation signature

dataclouder:
These are additional fixes as requested in comments to PR #112 

* Change Network creation function signature
* Add comments for less obvious tests and operations

Signed-off-by: Giuseppe Maxia

### DIFF
--- a/govcd/orgvdcnetwork.go
+++ b/govcd/orgvdcnetwork.go
@@ -109,11 +109,11 @@ func RemoveOrgVdcNetworkIfExists(vdc Vdc, networkName string) error {
 	return nil
 }
 
-// A wrapper call around CreateOrgVDCNetWork.
+// A wrapper call around CreateOrgVDCNetwork.
 // Creates a network and then uses the associated task to monitor its configuration
 func (vdc *Vdc) CreateOrgVDCNetworkWait(networkConfig *types.OrgVDCNetwork) error {
 
-	task, err := vdc.CreateOrgVDCNetWork(networkConfig)
+	task, err := vdc.CreateOrgVDCNetwork(networkConfig)
 	if err != nil {
 		return fmt.Errorf("error creating the network: %s", err)
 	}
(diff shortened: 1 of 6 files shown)


govcd/task.go
```diff
@@ -79,14 +79,23 @@ func (task *Task) WaitInspectTaskCompletion(inspectionFunc InspectionFunc, delay
 			return fmt.Errorf("error retrieving task: %s", err)
 		}
 
+		// If an inspection function is provided, we pass information about the task processing:
+		// * The task itself
```
lvirbalas: Nit: convert to lower case "The" -> "the"

govcd/orgvdcnetwork.go
```diff
@@ -184,7 +186,7 @@ func (vdc *Vdc) CreateOrgVDCNetWorkBasic(networkConfig *types.OrgVDCNetwork) (Ta
 			for _, taskItem := range orgVDCNetwork.OrgVDCNetwork.Tasks.Task {
 				return Task{taskItem, vdc.client}, nil
 			}
-			return Task{}, fmt.Errorf("[CreateOrgVDCNetWorkBasic] No suitable task found")
+			return Task{}, fmt.Errorf("[CreateOrgVDCNetWork] No suitable task found")
```
lvirbalas: Knowing function name here is useful, but if someone changes the name of the function (one more time), the function name in the error message may be left out :) Please see if using runtime reflection to get the function name could be a stable alternative (reflections are not always straightforward).

vbauzysvmware: This could be easy solved by: https://github.com/gogap/aop

dataclouder: I have added a function that returns the current function name, so that this message will always return the current function. BTW, the current function name is misspelled: the "w" should be lowercase, and so this change comes at the right moment.

govcd/task.go
dataclouder: fixed

Issue #14, opened 2018-08-02
VDC CRUD implementation


auppunda:
Similar to org and catalogs, we need to be able to create, read, update, delete vdcs

lvirbalas:
@auppunda I believe you have started coding this effort locally. Please post a small update where you stand and what challenges, if any, are you hitting.

auppunda:
Still sorting out xml issues with the create. I already have delete vdcs and get vdcs. Do not have update.

lvirbalas:
Now tracked in #117

Issue #18, opened 2018-08-06
User class implementation


auppunda:
User class needs to be created, with create, read, update, delete methods

PR #117, opened 2018-11-07
Create and delete for VDCs

megalord:
## Description

Add methods for creating a VDC in an (admin) org and deleting a VDC.

Related issue: (#14)

I'm ultimately trying to add support for managing VDCs to the [vcd terraform provider](https://github.com/terraform-providers/terraform-provider-vcd), focusing on the `vmware-takes-ownership` branch that pulls in this repo as a dependency.  I have a few questions on some code here that I don't fully follow and have left those comments inline.


lvirbalas:
Hi @megalord 

First of all, thank you for joining the project with a contribution!

Please carefully review the CONTRIBUTING.md - there are some important items that are missing in this PR currently. The first one that catches the eye is the lack of:
> Include unit tests for new features and update tests for old ones. Refer to the testing guide for more details.

Please see TESTING.md for more details. I'm looking forward to seeing you amending this PR! Creating VDCs is a needed feature indeed.

lvirbalas:
p.s. Nit: correct way of capitalising is "VDC" not "vDC" (I know, it's confusing). I updated the issue and PR name.

govcd/org.go
```diff
@@ -233,6 +233,31 @@ func (adminOrg *AdminOrg) GetVdcByName(vdcname string) (Vdc, error) {
 	return Vdc{}, nil
 }
 
+// CreateVdc creates a vDC with the given params under the given organization.
```
lvirbalas: Nit: correct capitalisation is "VDC".

govcd/org.go
```diff
@@ -233,6 +233,31 @@ func (adminOrg *AdminOrg) GetVdcByName(vdcname string) (Vdc, error) {
...
+// Returns a creation task.
+// API Documentation: https://code.vmware.com/apis/220/vcloud#/doc/doc/operations/POST-CreateVdcParams.html
+func (org *AdminOrg) CreateVdc(vdc *types.CreateVdcParams) (Task, error) {
```
vbauzysvmware: I recommended to add logging for trace/debug, info levels.

vbauzysvmware: Create function should have validation for mandatory values.

govcd/vdc.go
```diff
@@ -131,6 +132,38 @@ func (vdc *Vdc) Refresh() error {
 	return nil
 }
 
+// Deletes the vdc, returning an error of the vCD call fails.
+// API Documentation: https://code.vmware.com/apis/220/vcloud#/doc/doc/operations/DELETE-Vdc.html
+func (vdc *Vdc) Delete(force bool, recursive bool) error {
```
vbauzysvmware: I recommended to add logging for trace/debug, info levels.

govcd/vdc.go
```diff
@@ -131,6 +132,38 @@ func (vdc *Vdc) Refresh() error {
...
+
+	if vdc.Vdc.HREF == "" {
+		return fmt.Errorf("cannot refresh, Object is empty")
+	}
+
+	vdcUrl, _ := url.ParseRequestURI(vdc.Vdc.HREF)
```
vbauzysvmware: We shouldn't omit low possibility errors.

govcd/vdc.go
```diff
@@ -131,6 +132,38 @@ func (vdc *Vdc) Refresh() error {
...
+		return fmt.Errorf("error deleting vdc: %s", err)
+	}
+	task := NewTask(vdc.client)
+	if err = decodeBody(resp, task.Task); err != nil {
+		return fmt.Errorf("error decoding task response: %s", err)
+	}
+	if task.Task.Status == "error" {
+		return fmt.Errorf("vdc not properly destroyed")
+	}
+	err = task.WaitTaskCompletion()
```
vbauzysvmware: If VDC with a lot of resources takes time we should give task handling to client (api user).

types/v56/types.go
```diff
@@ -373,6 +373,49 @@ type Vdc struct {
 	VMQuota            int                   `xml:"VmQuota"`
 }
 
+// VdcStorageProfile represents the parameters to create a storage profile in an organization vDC.
+// Type: VdcStorageProfileParamsType
+// Namespace: http://www.vmware.com/vcloud/v1.5
+// Description: Represents the parameters to create a storage profile in an organization vDC.
+// Since: 5.1
+// https://code.vmware.com/apis/220/vcloud#/doc/doc/types/VdcStorageProfileParamsType.html
+type VdcStorageProfile struct {
```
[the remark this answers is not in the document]
megalord: I agree it isn't obvious, but isn't that just a result of the API definitions?  The API defines [VdcStorageProfilesType](https://code.vmware.com/apis/220/vcloud#/doc/doc/types/VdcStorageProfilesType.html) and [VdcStorageProfileType](https://code.vmware.com/apis/220/vcloud#/doc/doc/types/VdcStorageProfileType.html) differently, so it seems that design is just being followed here.  I don't mind deviating if you think it's best; maybe we could call this `VdcStorageProfileDefinition` or something since it actually has the data elements (rather than just being a reference).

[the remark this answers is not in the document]
megalord: Also perhaps I'm in the wrong mindset of trying to follow the API too closely.  I see one of your other comments for changing `CreateVdcParams` to `VdcConfiguration`, so it seems that's not a goal here?

govcd/vdc.go
```diff
@@ -131,6 +132,38 @@ func (vdc *Vdc) Refresh() error {
 	return nil
 }
 
+// Deletes the vdc, returning an error of the vCD call fails.
+// API Documentation: https://code.vmware.com/apis/220/vcloud#/doc/doc/operations/DELETE-Vdc.html
+func (vdc *Vdc) Delete(force bool, recursive bool) error {
+
+	if vdc.Vdc.HREF == "" {
+		return fmt.Errorf("cannot refresh, Object is empty")
```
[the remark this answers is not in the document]
megalord: Yep! Sorry

govcd/org.go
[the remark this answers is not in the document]
megalord: Because we changed the type name to `VdcConfiguration`, I opted for `vdcConfiguration` instead of `vdcDefinition`.  Is that ok?

megalord:
Thanks for the feedback y'all!  I addressed most of the feedback and added a few questions.

summary of the changes:
* trace logging at the beginning of the function calls
* type name changes and addition of `AdminVdc` type - that way `CreateVdc` returns more than a `Task`
* validation of the `VdcConfiguration` type passed to `CreateVdc`
* new test called `Test_CreateVdc` - I had to add some config values, but because the library doesn't have support for those types, I had to use their UUIDs instead of their names (like the rest of the config).  Should I be approaching that differently?

types/v56/types.go
```diff
@@ -409,12 +452,12 @@ type Task struct {
...
 // Since: 0.9
 type CapacityWithUsage struct {
+	Units     string `xml:"Units"`
 	Allocated int64  `xml:"Allocated,omitempty"`
 	Limit     int64  `xml:"Limit,omitempty"`
-	Overhead  int64  `xml:"Overhead,omitempty"`
 	Reserved  int64  `xml:"Reserved,omitempty"`
-	Units     string `xml:"Units"`
 	Used      int64  `xml:"Used,omitempty"`
+	Overhead  int64  `xml:"Overhead,omitempty"`
```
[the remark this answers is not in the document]
vbauzysvmware: XSD shema provided with documentation requires elements to be in specific order.

govcd/org.go
```diff
@@ -233,6 +233,31 @@ func (adminOrg *AdminOrg) GetVdcByName(vdcname string) (Vdc, error) {
...
+	resp, err := checkResp(org.client.Http.Do(req))
+	if err != nil {
+		return Task{}, fmt.Errorf("error instantiating a new VDC: %s", err)
+	}
+
+	task := NewTask(org.client)
+	if err = decodeBody(resp, task.Task); err != nil {
+		return Task{}, fmt.Errorf("error decoding task response: %s", err)
+	}
+	return *task, nil
```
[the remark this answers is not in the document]
vbauzysvmware: Current implementation isn't exactly correct. Main function returns task and wrapper function is helper function if api user don't want to handle waiting for himself.

vbauzysvmware:
Your changes are very good. I see one only issue left from my review part - is that createVdc should return task.

govcd/org.go
```diff
@@ -233,6 +235,100 @@ func (adminOrg *AdminOrg) GetVdcByName(vdcname string) (Vdc, error) {
...
+
+	adminVdc := NewAdminVdc(org.client)
+	if err = decodeBody(resp, adminVdc.AdminVdc); err != nil {
+		return AdminVdc{}, fmt.Errorf("error decoding admin VDC response: %s", err)
+	}
+	return *adminVdc, nil
+}
+
+// Creates the vdc and waits for the asynchronous task to complete.
+func (org *AdminOrg) CreateVdcWait(vdcDefinition *types.VdcConfiguration) (AdminVdc, error) {
```
dataclouder: The type `AdminVdc` is not visible without being qualified with the `types` package.

dataclouder: This function should return an `error` and a `Task`.

govcd/api_vcd_test.go
```diff
@@ -49,8 +50,12 @@ type TestConfig struct {
 		SysOrg   string `yaml:"sysOrg"`
 	}
 	VCD struct {
-		Org     string `yaml:"org"`
-		Vdc     string `yaml:"vdc"`
+		Org         string `yaml:"org"`
+		Vdc         string `yaml:"vdc"`
+		ProviderVdc struct {
+			Id               string `yaml:"id"`
```
dataclouder: What are these `Id` and `StorageProfileId`?
Foe all the other tests, we refer to entities by name. Could you change it so that we know what a tester should provide to try this new feature?
An example and some comments would be also useful.

lvirbalas:
@megalord would you be open to adjust the test to use entity names for the provider VDC and storage profile in an effort to make the test easier to configure?

megalord:
Hey, sorry for the delay y'all.  I'm open to it, yes, but I'm not sure how it would be immediately feasible.  The docs say that the `href` attribute is required on the `ProviderVdcReference` element, and the `href` I've been testing with is of the form `<hostname>/api/admin/providervdc/<uuid>`, which aligns with the documented call [here](https://vdc-download.vmware.com/vmwb-repository/dcr-public/ca48e1bb-282b-4fdc-b827-649b819249ed/55142cf1-5bb8-4ab1-8d09-b84f717af5ec/doc/doc/operations/GET-ProviderVdc.html) (the main docs site appears to be having CORS issues atm).  I know the Provider VDC does have a name, as returned in that api call, but I don't see a way to query by it.

Do you see a path forward on how to use the entity names?

megalord:
Ok I spent awhile poking around the api and found a way to get those `href`s by name:
```
/api/query?type=providerVdcStorageProfile&filter=(name==<name>)
/api/query?type=providerVdc&filter=(name==<name>)
```
I see there is a `Query()` method that can make this call, but it seems I'll need to define new types for `VMWProviderVdcRecord` and `ProviderVdcStorageProfileRecord` and add them to [`QueryResultRecordsType`](https://github.com/vmware/go-vcloud-director/blob/288b40347ca1da41589b2f342482eea76e604ac9/types/v56/types.go#L1905-L1910).  Does that seem like the best way to go about it?

vbauzysvmware:
Yes, query service works nicely and returns needed data. I can help on that if you need it. Also some changes are done, so you can merge latest code. And yes - to parse data, you have to add new types and references to them.

> Ok I spent awhile poking around the api and found a way to get those `href`s by name:
> 
> ```
> /api/query?type=providerVdcStorageProfile&filter=(name==<name>)
> /api/query?type=providerVdc&filter=(name==<name>)
> ```
> 
> I see there is a `Query()` method that can make this call, but it seems I'll need to define new types for `VMWProviderVdcRecord` and `ProviderVdcStorageProfileRecord` and add them to [`QueryResultRecordsType`](https://github.com/vmware/go-vcloud-director/blob/288b40347ca1da41589b2f342482eea76e604ac9/types/v56/types.go#L1905-L1910). Does that seem like the best way to go about it?

megalord:
Glad you mentioned that!  I was hitting a wall with
```Bad request: Expression name%3D%3Dprovider is not a comparison expression.```
, then saw the new `QueryWithNotEncodedParams()` method, thanks!

The latest commit has those test changes with the new Record types.

File: govcd/api_vcd_test.go [hunk 12]
```diff
@@ -63,8 +64,12 @@ type TestConfig struct {
 		SysOrg   string `yaml:"sysOrg"`
 	}
 	VCD struct {
-		Org     string `yaml:"org"`
-		Vdc     string `yaml:"vdc"`
+		Org         string `yaml:"org"`
+		Vdc         string `yaml:"vdc"`
+		ProviderVdc struct {
+			Name           string `yaml:"name"`
```
dataclouder: We need to add at least a network pool to create an usable VDC.
We could either add a specific network pool to the test fields, or search for the available pools.
I don't know if we can easily use a network pool with the API at our disposal. 
What do you think?

govcd/org_test.go
```diff
@@ -174,6 +175,86 @@ func (vcd *TestVCD) Test_Admin_GetVdcByName(check *C) {
...
+			Default: true,
+			ProviderVdcStorageProfile: &types.Reference{
+				HREF: providerVdcStorageProfileHref,
+			},
+		},
+		ProviderVdcReference: &types.Reference{
+			HREF: providerVdcHref,
+		},
+	}
+	task, err := adminOrg.CreateVdc(vdcConfiguration)
```
dataclouder: The VDC creation test should do the following:
* test all three allocation models, either re-creating the same VDC or creating three different ones;
* Use both thin provisioning and fast provisioning (we need the tests to use little resources)
* Check that the resulting VDC is enabled and has the relevant elements according to the previous points.

govcd/org_test.go
```diff
@@ -174,6 +175,86 @@ func (vcd *TestVCD) Test_Admin_GetVdcByName(check *C) {
...
+	check.Assert(task, Equals, Task{})
+	check.Assert(err, Not(IsNil))
+	check.Assert(err.Error(), Equals, "VdcConfiguration missing required field: ComputeCapacity[0].Memory.Units")
+	vdcConfiguration.ComputeCapacity[0].Memory.Units = "MB"
+
+	err = adminOrg.CreateVdcWait(vdcConfiguration)
+	check.Assert(err, IsNil)
+
```
dataclouder: All tests, after creating an entity, need to add the entity to the cleanup list (using function `AddtoCleanupList`), and there must be a cleanup section for VDC inside `removeLeftoverEntities` (currently, there is an empty place holder for VDC). The purpose of this method is to remove leftovers after failed operations.

govcd/org_test.go
```diff
@@ -174,6 +175,86 @@ func (vcd *TestVCD) Test_Admin_GetVdcByName(check *C) {
...
+// Fails if the config file name doesn't match with the found vdc, or
+// if the invalid vdc is found by the function.  Also tests an vdc
+// that doesn't exist. Asserts an error if the function finds it or
+// if the error is not nil.
+func (vcd *TestVCD) Test_CreateVdc(check *C) {
+	if vcd.skipAdminTests {
+		check.Skip("Configuration org != 'Sysyem'")
+	}
+
+	results, err := vcd.client.QueryWithNotEncodedParams(nil, map[string]string{
```
dataclouder: Before creating a VDC, we should check whether the VDC exists already. If yes, we should either delete it and create a new one, or exit gracefully with a meaningful error message.

types/v56/types.go
```diff
@@ -373,6 +373,68 @@ type Vdc struct {
...
+type AdminVdc struct {
+	Vdc
+
+	ResourceGuaranteedMemory float64    `xml:"ResourceGuaranteedMemory,omitempty"`
+	ResourceGuaranteedCpu    float64    `xml:"ResourceGuaranteedCpu,omitempty"`
+	VCpuInMhz                int64      `xml:"VCpuInMhz,omitempty"`
+	IsThinProvision          bool       `xml:"IsThinProvision,omitempty"`
+	NetworkPoolReference     *Reference `xml:"NetworkPoolReference,omitempty"`
+	ProviderVdcReference     *Reference `xml:"ProviderVdcReference"`
+	UsesFastProvioning       bool       `xml:"UsesFastProvioning,omitempty"`
```
dataclouder: Typo: `UsesFastProvioning` should be `UsesFastProvisioning`

govcd/api_vcd_test.go
megalord: I don't mind adding another parameter here for the network pool.  Is it ok to follow the same pattern using the query api to run a lookup by name?

govcd/org_test.go
megalord: Sweet thanks - I added code to the switch statement for the VDC case.

govcd/org_test.go
dataclouder: Yes. Let's test what is allowed.

govcd/org_test.go
```diff
@@ -174,6 +175,119 @@ func (vcd *TestVCD) Test_Admin_GetVdcByName(check *C) {
...
+		}
+		task, err := adminOrg.CreateVdc(vdcConfiguration)
+		check.Assert(task, Equals, Task{})
+		check.Assert(err, Not(IsNil))
+		check.Assert(err.Error(), Equals, "VdcConfiguration missing required field: ComputeCapacity[0].Memory.Units")
+		vdcConfiguration.ComputeCapacity[0].Memory.Units = "MB"
+
+		err = adminOrg.CreateVdcWait(vdcConfiguration)
+		check.Assert(err, IsNil)
+
```
dataclouder: Thanks for your updates. We're almost there. A few requests for the creation test:
1. As mentioned above, the test must check whether the VDC already exists before creating it. As it is, if the VDC exists, we get an ugly error message. Given that the test creates a VDC that is supposed to be deleted soon, I think a safe course of action would be to delete it and then proceed with the creation. 
2. The addition to the cleanup list works fine, but it should not be the main method for deletion. It is there as a safety net to take over when the test function exits prematurely. Instead, we should have an explicit deletion, and the test for its success (`err` equals nil, and a search for the VDC returns an empty result)

megalord: Dang sorry about that, I wrote and tested the check for the VDC already existing before adding all the allocation models.  I moved it inside the loop and ran the tests with and without the VDCs already existing.

Gotcha, I misinterpreted what you meant and have added the call to `vdc.DeleteWait` back into the test.  I've also added a refresh on the org after deletion to ensure an empty result.

Thank you for your patience!

dataclouder:
P.S. While waiting for approval on this one, you may want to consider one of the issues #18, #129, and #130 to complete the VDC automation!

lvirbalas:
@megalord Thank you for your perseverance.

vbauzysvmware:
Thank you, very nice ;-)

megalord:
Sweet!  I'm working on the provider changes [here](https://github.com/megalord/terraform-provider-vcd/tree/vdc), but I need to spend more time on it before raising a pull request.  I appreciate y'all's help!

PR #118, opened 2018-11-09
Upgrade API version from 5.5 to 27.0

vbauzysvmware:
This changes to Support Vulcan's vCD API version (with compatibility of vCD 8.20+)
Changes
- update hard coded parts
- change API to 27.0
- test updates
- update documentation with issues

Note that newer API has issue creating organization:
Organization creation has two bugs https://bugzilla.eng.vmware.com/show_bug.cgi?id=2177355, https://bugzilla.eng.vmware.com/show_bug.cgi?id=2228936 which requires, that creating organization elements are in incorrect order, at least one element(DelayAfterPowerOnSeconds, DeployedVMQuota, StoredVmQuota, UseServerBootSequence, getVdcQuota) is needed then providing generalOrgSettings and VAppLeaseSettings/VAppLeaseSettings provided fully or not at all.

### DIFF
--- a/govcd/api_vcd.go
+++ b/govcd/api_vcd.go
@@ -89,7 +89,7 @@ func NewVCDClient(vcdEndpoint url.URL, insecure bool) *VCDClient {
 
 	return &VCDClient{
 		Client: Client{
-			APIVersion: "27.0",
+			APIVersion: "27.0", // supported by vCD 8.20, 9.0, 9.1, 9.5
 			VCDHREF:    vcdEndpoint,
 			Http: http.Client{
 				Transport: &http.Transport{
--- a/govcd/api_vcd.go
+++ b/govcd/api_vcd.go
@@ -70,7 +70,7 @@ func (vdcCli *VCDClient) vcdauthorize(user, pass, org string) error {
 	// Set Basic Authentication Header
 	req.SetBasicAuth(user+"@"+org, pass)
 	// Add the Accept header for vCA
-	req.Header.Add("Accept", "application/*+xml;version=5.5")
+	req.Header.Add("Accept", "application/*+xml;version="+vdcCli.Client.APIVersion)
 	resp, err := checkResp(vdcCli.Client.Http.Do(req))
 	if err != nil {
 		return err
@@ -89,7 +89,7 @@ func NewVCDClient(vcdEndpoint url.URL, insecure bool) *VCDClient {
 
 	return &VCDClient{
 		Client: Client{
-			APIVersion: "5.5",
+			APIVersion: "27.0",
 			VCDHREF:    vcdEndpoint,
 			Http: http.Client{
 				Transport: &http.Transport{
@@ -126,7 +126,7 @@ func (vdcCli *VCDClient) Disconnect() error {
 	}
 	req := vdcCli.Client.NewRequest(map[string]string{}, "DELETE", vdcCli.sessionHREF, nil)
 	// Add the Accept header for vCA
-	req.Header.Add("Accept", "application/xml;version=5.5")
+	req.Header.Add("Accept", "application/xml;version="+vdcCli.Client.APIVersion)
 	// Set Authorization Header
 	req.Header.Add(vdcCli.Client.VCDAuthHeader, vdcCli.Client.VCDToken)
 	if _, err := checkResp(vdcCli.Client.Http.Do(req)); err != nil {
--- a/govcd/org_test.go
+++ b/govcd/org_test.go
@@ -7,6 +7,7 @@ package govcd
 import (
 	types "github.com/vmware/go-vcloud-director/types/v56"
 	. "gopkg.in/check.v1"
+	"time"
 )
 
 // Tests Refresh for Org by updating the org and then asserting if the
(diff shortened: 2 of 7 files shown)


govcd/system_test.go
```diff
@@ -88,6 +95,7 @@ func (vcd *TestVCD) Test_CreateOrg(check *C) {
 	err = org.Delete(true, true)
 	check.Assert(err, IsNil)
 	// Check if org still exists
+	time.Sleep(10 * time.Second)
```
lvirbalas: This is a comment regarding all time.Sleep(...) calls in this PR. Even though they are only in tests, we don't want to block code with specific time lengths. This can lead to (a) slower test runs when you we are waiting longer than we should or (b) race conditions when we are waiting not long enough.

What about replacing these sleeps with a handy wait-for function? It should loop by trying to get the needed resource until it's available. For looping, 1 second interval should be a good middle ground.

govcd/system.go
```diff
@@ -18,6 +18,9 @@ import (
 // settings parameter. The settings variable is defined in types.go.
 // Method will fail unless user has an admin token.
 // API Documentation: https://code.vmware.com/apis/220/vcloud#/doc/doc/operations/POST-CreateOrganization.html
+// Organization creation has two bugs https://bugzilla.eng.vmware.com/show_bug.cgi?id=2177355, https://bugzilla.eng.vmware.com/show_bug.cgi?id=2228936 which requires,
+// that creating organization elements are in incorrect order, at least one element(DelayAfterPowerOnSeconds, DeployedVMQuota, StoredVmQuota, UseServerBootSequence, getVdcQuota)
+// is needed then providing generalOrgSettings and VAppLeaseSettings/VAppLeaseSettings provided fully.
```
lvirbalas: Please rephrase this comment to be clearer and answer the following questions:
- What is the gist of these bugs?
- Which "one element" is needed? It's hard to understand as it stands right now.
- How these two bugs have been worked-around by your code below to make it work?
- The links are useful only if you're connected to internal network. Please remove the links, but keep the issue numbers (BZ 2177355 and BZ 2228936).

govcd/system_test.go
vbauzysvmware: Fixed

govcd/system.go
vbauzysvmware: Comment updated with Giuseppe's help.

vbauzysvmware: And Organization creation works. I verified with terraform provider too.

govcd/org_test.go
```diff
@@ -80,8 +80,14 @@ func (vcd *TestVCD) Test_DeleteOrg(check *C) {
...
-	time.Sleep(3 * time.Second)
-	org, err = GetAdminOrgByName(vcd.client, TestDeleteOrg)
+	for i := 0; i < 30; i++ {
+		org, err = GetAdminOrgByName(vcd.client, TestDeleteOrg)
+		if org == (AdminOrg{}) {
+			break
+		} else {
+			time.Sleep(1 * time.Second)
+		}
+	}
```
lvirbalas: There are three occurrences of the same loop in this commit. Please think of a good way to get rid of these duplications. I mean something between the lines of:
- What about introducing a utility function in `system.go` which would wait for the org for the given amount of seconds. Something like `WaitAdminOrgByName(vcd.client, TestDeleteOrg, 30)`.
- Or what about introducing a task here, if it takes so long to be created?
- Etc.

vbauzysvmware: This is test, so no task here. We using old api part where no task returned and changes to Organization API not this PR part. I introduced sleep just I spotted that it randomly fails when delete wasn't fast at some time.

lvirbalas: OK, I now realise that you are actually waiting for Delete to completely _remove_ the Org.

govcd/api_vcd.go
```diff
@@ -89,7 +89,7 @@ func NewVCDClient(vcdEndpoint url.URL, insecure bool) *VCDClient {
 
 	return &VCDClient{
 		Client: Client{
-			APIVersion: "5.5",
+			APIVersion: "27.0",
```
dataclouder: Is it possible to make this version dynamic or depending on a global variable?

govcd/org_test.go
```diff
@@ -120,7 +128,14 @@ func (vcd *TestVCD) Test_UpdateOrg(check *C) {
 	err = org.Delete(true, true)
 	check.Assert(err, IsNil)
 	// Check if org still exists
-	org, err = GetAdminOrgByName(vcd.client, TestUpdateOrg)
+	for i := 0; i < 30; i++ {
```
dataclouder: I suggest replacing this repeated loop with a function named `doesOrgExist`, which may also be reused by other tests.

govcd/api_vcd.go
dataclouder: If this is the intention, then let's put the supported API version in a constant with a telling name.

govcd/org_test.go
vbauzysvmware: Fixed ;-)

govcd/api_vcd.go
vbauzysvmware: What is you suggestion for name?

dataclouder: Something like `maximumSupportedVersion`.

lvirbalas: `APIVersion` name is correct here. It just means our client supports API v27.0 - it is _not_ minimum or maximum supported version. What we could do to be more clear is to add a comment like `APIVersion: "27.0", // vCD 8.20`

vbauzysvmware:
Verified with 8.20 version too. One fixed added.

PR #124, opened 2018-12-06
Add ability to cancel task

vbauzysvmware:
Small change to improve api
Signed-off-by: Vaidotas Bauzys <vbauzys@vmware.com>

### DIFF
--- a/govcd/catalog.go
+++ b/govcd/catalog.go
@@ -288,7 +288,7 @@ func uploadFiles(client *Client, vappTemplate *types.VAppTemplate, ovfFileDesc *
 			number, err := getFileFromDescription(item.Name, ovfFileDesc)
 			if err != nil {
 				util.Logger.Printf("[Error] Error uploading files: %#v", err)
-				uploadError = &err
+				*uploadError = err
 				return err
 			}
 			if ovfFileDesc.File[number].ChunkSize != 0 {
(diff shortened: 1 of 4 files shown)


govcd/task.go
```diff
@@ -140,3 +141,19 @@ func (task *Task) GetTaskProgress() (string, error) {
 
 	return strconv.Itoa(task.Task.Progress), nil
 }
+
+func (task *Task) CancelTask() error {
+	cancelTaskURL, err := url.ParseRequestURI(task.Task.HREF + "/action/cancel")
+	if err != nil {
+		util.Logger.Printf("[Error] Error cancellins task %v: %s", cancelTaskURL.String(), err)
```
lvirbalas: Spelling error in all error messages: "cancellins" -> "cancelling".

govcd/catalog.go
```diff
@@ -663,9 +663,11 @@ func removeCatalogItemOnError(client *Client, vappTemplateLink *url.URL, itemNam
...
 		}
 
-		for _, task := range vAppTemplate.Tasks.Task {
-			if itemName == task.Owner.Name {
-				err = cancelTask(client, task.HREF)
+		for _, taskItem := range vAppTemplate.Tasks.Task {
+			if itemName == taskItem.Owner.Name {
+				task := NewTask(client)
+				task.Task = taskItem
+				err = task.CancelTask()
```
lvirbalas: Why are we creating a new task and then calling cancel on it as opposed to calling cancel on the taskItem straight? Same question for the change in media.go file.

vbauzysvmware: TaskItem is type represting xml, and Task is "object".

Issue #131, opened 2018-12-14
ChangeStorageProfile test fails if it runs after AttachedVMDisk test


dataclouder:
When we run [`Test_ChangeStorageProfile`](https://github.com/vmware/go-vcloud-director/blob/a9f437f0ff51215d5df391585f5ecbfc3a1e8c40/govcd/vapp_test.go#L174) on its own, or when we skip all the independent disk tests, the test succeeds.
However, if we run the full suite (and the disk tests run before the storage profile test), we get this failure:
```
START: vapp_test.go:174: TestVCD.Test_ChangeStorageProfile
vapp_test.go:182:
    check.Assert(err, IsNil)
... value *errors.errorString = &errors.errorString{s:"error customizing VM: API Error: 400: [ bdf2b9b5-2bef-443e-81d9-2958632b83ca ] 
Cannot use multiple storage profiles in a fast-provisioned VDC \"datacloudvdc\" for VM \"Photon OS\"."} 
("error customizing VM: API Error: 400: [ bdf2b9b5-2bef-443e-81d9-2958632b83ca ] 
Cannot use multiple storage profiles in a fast-provisioned VDC \"datacloudvdc\" for VM \"Photon OS\".")

FAIL: vapp_test.go:174: TestVCD.Test_ChangeStorageProfile
```
It is puzzling: if the error is legitimate, the test should fail always. Needs more investigation.

How to repeat:
```
cd govcd
export GOCACHE=off
go test -timeout=0 -check.vv -check.f '(Test_AttachedVMDisk|Test_ChangeStorage)' .
```

Related to PR #104.
cc @ty2

PR #133, opened 2018-12-19
Bug fixes

dataclouder:
Fix Issue #131 `ChangeStorageProfile` fails if it runs after `AttachedVMDisk`
    This failure occurs because the disk test leaves a disk attached to a VM.
    The fix detaches the disk at the end of the test.

Change snake_case function names to camelCase in vm_test.go

Add checks for configuration items in VDC test.
    Without these checks, the test was failing with unhelpful messages
    when the configuration file did not have the needed section filled.

Fix assertion for error in `Test_GetOrgByName`
    The assertion was checking for a nil error, but when we explicitly
    search for a non-existing item, we expect the error to be set.

Fix several misspelling occurrences of the word "System" in skip messages

Add missing error handling in `ChangeStorageProfile`

### DIFF
--- a/govcd/disk_test.go
+++ b/govcd/disk_test.go
@@ -348,6 +348,9 @@ func (vcd *TestVCD) detachIndependentDisk(disk Disk) error {
 				HREF: disk.Disk.HREF,
 			},
 		})
+		if err != nil {
+			return err
+		}
 		err = task.WaitTaskCompletion()
 		if err != nil {
 			return err
--- a/govcd/disk_test.go
+++ b/govcd/disk_test.go
@@ -254,8 +254,8 @@ func (vcd *TestVCD) Test_AttachedVMDisk(check *C) {
 	}
 
 	// Find VM
-	vapp := vcd.find_first_vapp()
-	vmType, vmName := vcd.find_first_vm(vapp)
+	vapp := vcd.findFirstVapp()
+	vmType, vmName := vcd.findFirstVm(vapp)
 	if vmName == "" {
 		check.Skip("skipping test because no VM is found")
 	}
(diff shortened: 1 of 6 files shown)


govcd/disk_test.go
```diff
@@ -321,6 +321,39 @@ func (vcd *TestVCD) Test_AttachedVMDisk(check *C) {
...
+			return err
+		}
+
+		// Detach the disk from VM
+		task, err := vm.DetachDisk(&types.DiskAttachOrDetachParams{
+			Disk: &types.Reference{
+				HREF: disk.Disk.HREF,
+			},
+		})
+		err = task.WaitTaskCompletion()
```
vbauzysvmware: Missing error handling from DetachDisk

govcd/vapp.go
```diff
@@ -610,22 +610,28 @@ func (vapp *VApp) ChangeStorageProfile(name string) (Task, error) {
...
 
-	if vapp.VApp.Children == nil {
+	if vapp.VApp.Children == nil || len(vapp.VApp.Children.VM) == 0 {
 		return Task{}, fmt.Errorf("vApp doesn't contain any children, aborting customization")
 	}
 
 	vdc, err := vapp.getParentVDC()
-	storageprofileref, err := vdc.FindStorageProfileReference(name)
+	if err != nil {
+		return Task{}, fmt.Errorf("error retrieving parent VDC for vapp %s", vapp.VApp.Name)
```
lvirbalas: "for vapp" -> "for vApp"

govcd/vapp.go
```diff
@@ -610,22 +610,28 @@ func (vapp *VApp) ChangeStorageProfile(name string) (Task, error) {
...
+	}
+	storageProfileRef, err := vdc.FindStorageProfileReference(name)
+	if err != nil {
+		return Task{}, fmt.Errorf("error retrieving storage profile %s for vapp %s", name, vapp.VApp.Name)
```
lvirbalas: "for vapp" -> "for vApp"

govcd/vapp.go
```diff
@@ -610,22 +610,28 @@ func (vapp *VApp) ChangeStorageProfile(name string) (Task, error) {
...
-		StorageProfile: &storageprofileref,
+		StorageProfile: &storageProfileRef,
 		Xmlns:          "http://www.vmware.com/vcloud/v1.5",
 	}
 
-	output, err := xml.MarshalIndent(newprofile, "  ", "    ")
+	output, err := xml.MarshalIndent(newProfile, "  ", "    ")
 	if err != nil {
-		fmt.Printf("error: %v\n", err)
+		return Task{}, fmt.Errorf("error encoding storage profile change metadata for vapp %s", vapp.VApp.Name)
```
lvirbalas: "for vapp" -> "for vApp"

govcd/org_test.go
```diff
@@ -191,9 +191,18 @@ func (vcd *TestVCD) Test_Admin_GetVdcByName(check *C) {
...
 	}
 
+	if vcd.config.VCD.ProviderVdc.Name == "" {
+		check.Skip("No Provider VDC name given for VDC tests")
+	}
+	if vcd.config.VCD.ProviderVdc.StorageProfile == "" {
+		check.Skip("No Storage Profile given for VDC tests")
+	}
+	if vcd.config.VCD.ProviderVdc.NetworkPool == "" {
+		check.Skip("No Network Pool given for VDC tests")
```
lvirbalas: Shouldn't all three messages above say "for Org tests" instead of "for VDC tests"?

dataclouder: No. It's the VDC creation test that gets skipped. This test is in `org_test.go` because we have the creation cases in the parent's entity test file. For example, the Org creation test is in `system_test.go`.

govcd/disk_test.go
dataclouder: Done. Also added the same error handling in the cleanup procedure

PR #135, opened 2018-12-27
Bug fixing 2018 12

dataclouder:
## Add workaround for disk update bug

The disk update has a problem when using a disk size that is less than one MB. Even if the disk size is the same or bigger than the original disk, when the size is less than one MB the update fails with an error message saying that the update size is less than the original size.
To circumvent this error, I have introduced a check that requires at least 1 MB for the disk size, both for creation and for updates.

##  Add option to skip logging of custom tags

Since there are tags that are very chatty (e.g. SupportedVersions,VAppTemplate) it is possible to set a list of those we want to skip, using either an internal variable or and environment variable.

It is also possible to set a list of functions that we want to log (rather than logging everything).

### DIFF
--- a/govcd/disk.go
+++ b/govcd/disk.go
@@ -29,6 +29,8 @@ func NewDisk(cli *Client) *Disk {
 	}
 }
 
+const MinimumDiskSize int = 1048576
+
 // Create an independent disk in VDC
 // Reference: vCloud API Programming Guide for Service Providers vCloud API 30.0 PDF Page 102 - 103,
 // https://vdc-download.vmware.com/vmwb-repository/dcr-public/1b6cf07d-adb3-4dba-8c47-9c1c92b04857/
@@ -43,8 +45,8 @@ func (vdc *Vdc) CreateDisk(diskCreateParams *types.DiskCreateParams) (Task, erro
 		return Task{}, fmt.Errorf("disk name is required")
 	}
 
-	if diskCreateParams.Disk.Size <= 0 {
-		return Task{}, fmt.Errorf("disk size should be greater than or equal to 1KB")
+	if diskCreateParams.Disk.Size < MinimumDiskSize {
+		return Task{}, fmt.Errorf("disk size should be greater than or equal to 1Mb")
 	}
 
 	var err error
@@ -104,6 +106,7 @@ func (vdc *Vdc) CreateDisk(diskCreateParams *types.DiskCreateParams) (Task, erro
 	task := NewTask(vdc.client)
 	task.Task = disk.Disk.Tasks.Task[0]
 
+	util.Logger.Printf("[TRACE] AFTER CREATE DISK\n %s\n", prettyDisk(*disk.Disk))
 	// Return the disk
 	return *task, nil
 }
@@ -129,8 +132,8 @@ func (d *Disk) Update(newDiskInfo *types.Disk) (Task, error) {
 		return Task{}, fmt.Errorf("disk name is required")
 	}
 
-	if newDiskInfo.Size <= 0 {
-		return Task{}, fmt.Errorf("disk size should be greater than or equal to 1KB")
+	if newDiskInfo.Size < MinimumDiskSize {
+		return Task{}, fmt.Errorf("disk size should be greater than or equal to 1Mb")
 	}
 
 	// Verify the independent disk is not connected to any VM
@@ -180,6 +183,7 @@ func (d *Disk) Update(newDiskInfo *types.Disk) (Task, error) {
 	if err != nil {
 		return Task{}, fmt.Errorf("error xml.Marshal: %s", err)
 	}
+	util.Logger.Printf("[TRACE] BEFORE UPDATE DISK\n %s\n", prettyDisk(*d.Disk))
 
 	// Send request
 	reqPayload := bytes.NewBufferString(xml.Header + string(xmlPayload))
@@ -267,7 +271,7 @@ func (d *Disk) Delete() (Task, error) {
 
 // Refresh the disk information by disk href
 func (d *Disk) Refresh() error {
-	util.Logger.Printf("[TRACE] Disk refersh, HREF: %s\n", d.Disk.HREF)
+	util.Logger.Printf("[TRACE] Disk refresh, HREF: %s\n", d.Disk.HREF)
 
 	disk, err := FindDiskByHREF(d.client, d.Disk.HREF)
 	if err != nil {
--- a/govcd/disk.go
+++ b/govcd/disk.go
@@ -29,7 +29,9 @@ func NewDisk(cli *Client) *Disk {
 	}
 }
 
-const MinimumDiskSize int = 1048576
+// While theoretically we can use smaller amounts, there is an issue when updating
+// disks with size < 1MB
+const MinimumDiskSize int = 1048576 // = 1Mb
 
 // Create an independent disk in VDC
 // Reference: vCloud API Programming Guide for Service Providers vCloud API 30.0 PDF Page 102 - 103,
--- a/govcd/sample_govcd_test_config.json
+++ b/govcd/sample_govcd_test_config.json
@@ -38,7 +38,7 @@
 		"logFileName": "go-vcloud-director.log",
 		"logHttpRequests": true,
 		"skipResponseTags": "SupportedVersions,VAppTemplate",
-		"includeFunctions": "FindVAppByName",
+		"apiLogFunctions": "FindVAppByName",
 		"logHttpResponses": true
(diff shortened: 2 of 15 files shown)


govcd/sample_govcd_test_config.yaml
```diff
@@ -70,10 +70,10 @@ vcd:
     disk:
       #
       # Disk size (bytes) for create disk, skip disk tests if it is less than or equal to 0
-      size: 1024
+      size: 1048576
       #
       # Disk size (bytes) for update disk, skip some disk tests if it is less than or equal to 0
-      sizeForUpdate: 10240
+      sizeForUpdate: 1048576
```
lvirbalas: How did the tests pass previously with disk size of less than an MB?

util/logging.go
```diff
@@ -188,14 +194,21 @@ func ProcessResponseOutput(caller string, resp *http.Response, result string) {
 	if !LogHttpResponse {
 		return
 	}
+	outText := result
+	if SkipVersionsResponse {
+		re := regexp.MustCompile(`<SupportedVersions`)
```
lvirbalas: Please add a comment how this regexp works.

dataclouder:
> Can there be a case where result contains not only version output? In this case, we would end up skipping everything. If it’s possible, then we need to only remove the version tag contents, but leave everything else intact.

[Theoretically not](https://code.vmware.com/apis/220/vcloud#/doc/doc/operations/GET-SupportedVersions.html)
However, to dissipate any doubts, I changed the implementation in such a way that we perform a regular expression based replacement instead of a blind one, so that only the text between the initial and final label will be removed.

govcd/sample_govcd_test_config.json
```diff
@@ -29,14 +29,15 @@
 		"externalIp": "10.150.10.10",
 		"internalIp": "192.168.1.10",
     "disk": {
-      "size": 1024,
-      "sizeForUpdate": 10240
+      "size": 1048576,
```
vbauzysvmware: Comment what this 1048576 means would be good. That is Megabyte in bytes.

govcd/sample_govcd_test_config.yaml
vbauzysvmware: Comment what is size 1048576 means would be good

govcd/sample_govcd_test_config.json
```diff
@@ -29,14 +29,15 @@
...
+      "sizeForUpdate": 1048576
     }
 	},
 	"logging": {
 		"enabled": true,
 		"logFileName": "go-vcloud-director.log",
 		"logHttpRequests": true,
+		"skipResponseTags": "SupportedVersions,VAppTemplate",
```
vbauzysvmware: Didn't we miss here includeFunctionList ?

util/LOGGING.md
```diff
@@ -56,6 +56,18 @@ During the request and response processing, any password or authentication token
...
+It is also possible to skip the output of the some tags (such as the result of `/versions` request,) which are quite large using 
+
+```go
+util.SkipTagList = "SupportedVersions,ovf:License"
+```
+
+For an even more dedicated log, you can define from which function names you want the logs, using
+
+```go
+util.IncludeVersionList = "FindVAppByName,GetAdminOrgByName"
```
vbauzysvmware: Why here is called "IncludeVersionList" ?

util/LOGGING.md
```diff
@@ -78,4 +88,6 @@ Variable                    | Corresponding environment var
 `LogOnScreen`               | `GOVCD_LOG_ON_SCREEN`
 `LogHttpRequest`            | `GOVCD_LOG_SKIP_HTTP_REQ`
 `LogHttpResponse`           | `GOVCD_LOG_SKIP_HTTP_RESP`
+`SkipTagList`               | `GOVCD_LOG_SKIP_TAG_LIST`
+`IncludeFunctionList`       | `GOVCD_LOG_INCLUDE_FUNCTION_LIST`
```
vbauzysvmware: I am not so sure about IncludeFunctionList - at least for me include means add aditionally what it is happening.
I would say better naming would be "OnlyFucntionList"

govcd/sample_govcd_test_config.json
dataclouder: The descriptions are only in the .yaml file

govcd/sample_govcd_test_config.json
dataclouder: Added

util/LOGGING.md
dataclouder: Indeed. And the variable is private. Changed to use the `Set*` functions

util/LOGGING.md
dataclouder: I use `include` as opposed to `skip`. I don't like using `only` in this context because it is not a verb.
However, if there is consensus, I will change the name to `OnlyFunctionList`.

vbauzysvmware: 'ShowOnlyFunctionList' then maybe. I am open for better naming.

util/logging.go
```diff
@@ -36,6 +37,12 @@ const (
 
 	// Name of the environment variable that enables logging of HTTP responses
 	envLogSkipHttpResp = "GOVCD_LOG_SKIP_HTTP_RESP"
+
+	// Name of the environment variable with a custom list of of responses to skip from logging
+	envLogSkipTagList = "GOVCD_LOG_SKIP_TAG_LIST"
+
+	// Name of the environment variable with a custom list of of functions to include in the logging
+	envLogIncludeFunctionList = "GOVCD_LOG_INCLUDE_FUNCTION_LIST"
```
lvirbalas: Suggestion:
- `GOVCD_LOG_SKIP_TAG_LIST` -> `GOVCD_LOG_SKIP_TAGS`
- `GOVCD_LOG_INCLUDE_FUNCTION_LIST` -> `GOVCD_LOG_FUNCTIONS`
(notice the plural forms too). Open question is whether we go with  "_LOG" or "_API_LOG".

govcd/disk.go
```diff
@@ -43,8 +45,8 @@ func (vdc *Vdc) CreateDisk(diskCreateParams *types.DiskCreateParams) (Task, erro
 		return Task{}, fmt.Errorf("disk name is required")
 	}
 
-	if diskCreateParams.Disk.Size <= 0 {
-		return Task{}, fmt.Errorf("disk size should be greater than or equal to 1KB")
+	if diskCreateParams.Disk.Size < MinimumDiskSize {
```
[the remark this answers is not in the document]
dataclouder: added

util/LOGGING.md
dataclouder: fixed using Linas's suggestion

govcd/sample_govcd_test_config.yaml
```diff
@@ -90,6 +92,15 @@ logging:
...
     # Defines whether we log the responses in HTTP operations
     logHttpResponses: true
+    #
+    # Comma-separated list of custom tags to skip from the logs
+    skipResponseTags: SupportedVersions,VAppTemplate
+    #
+    # Comma-separated list of functions from where we log the API calls.
+    # When this is set, we only log API requests and responses if the name
+    # of the function matches any of the names in this list.
+    includeFunctions: FindVAppByName,GetAdminOrgByName
```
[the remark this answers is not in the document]
dataclouder: done as suggested

util/LOGGING.md
```diff
@@ -78,4 +88,6 @@ Variable                    | Corresponding environment var
 `LogOnScreen`               | `GOVCD_LOG_ON_SCREEN`
 `LogHttpRequest`            | `GOVCD_LOG_SKIP_HTTP_REQ`
 `LogHttpResponse`           | `GOVCD_LOG_SKIP_HTTP_RESP`
+`SetSkipTags`               | `GOVCD_LOG_SKIP_TAG_LIST`
+`SetIncludeFunctions`       | `GOVCD_LOG_INCLUDE_FUNCTION_LIST`
```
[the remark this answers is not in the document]
dataclouder: done

util/logging.go
dataclouder: done

dataclouder:
Where should I write the description of the copyright check? Is [CONTRIBUTING.md](https://github.com/vmware/go-vcloud-director/blob/master/CONTRIBUTING.md) a good place?

lvirbalas:
> Where should I write the description of the copyright check? Is CONTRIBUTING.md a good place?

I meant just put a comment `# ...` into copyright_check.sh itself (somewhere towards the top), as the algorithm is now quite advanced.

PR #137, opened 2019-01-03
Fix bug in checking if sys admin

vbauzysvmware:
Issue found in terraform provider. Fix removes case sensitive check against sys org.


vbauzysvmware:
@dataclouder Seems Travis has issue with copyright check ?

dataclouder:
> @dataclouder Seems Travis has issue with copyright check ?

Yes. The fix is ion the way. As soon as I merge PR #135, it will behave again

vbauzysvmware:
@dataclouder still a few of them failing ^^

PR #138, opened 2019-01-04
Update copyright year for files recently modified

dataclouder:
This change is needed to satisfy the automated copyright check.
I've removed the check from Travis, but the script can be used for our own pre-committing checks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ TEST?=./...
 GOFMT_FILES?=$$(find . -name '*.go' | grep -v vendor)
 
 #default: fmt test testrace vet
-default: fmtcheck vet build copyright
+default: fmtcheck vet build
 
 # test runs the test suite and vets the code
 test: get-deps fmtcheck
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -2,7 +2,7 @@ TEST?=./...
 GOFMT_FILES?=$$(find . -name '*.go' | grep -v vendor)
 
 #default: fmt test testrace vet
-default: fmtcheck vet build copyright
+default: fmtcheck vet build
```
lvirbalas: Why did you remove the check, considering that you recently enhanced the copyright check to be smarter and only check recently changed files?

dataclouder: The check [failed unexpectedly](https://travis-ci.org/vmware/go-vcloud-director/builds/475189768) on a file that was not modified recently. I need to understand why. 
I think that my algorithm works on the assumption of a centralized source of truth, but fails on a truly distributed repository.

PR #144, opened 2019-01-11
Fix/revert delete catalog func cause it wasn't working for not admin user

vbauzysvmware:
We missed that previous fix changes causes failure for non admin user. We uses link from catalog entity, but delete api is accessible on /admin/ path only. I added comment for splicing number with example.

### DIFF
--- a/govcd/catalog.go
+++ b/govcd/catalog.go
@@ -82,7 +82,10 @@ func (catalog *Catalog) Delete(force, recursive bool) error {
 
 // slice only numeric part from ID:"urn:vcloud:catalog:97384890-180c-4563-b9b7-0dc50a2430b0"
 func getEntityNumericId(catalogId string) string {
-	return catalogId[19:]
+	if catalogId != "" && (len(catalogId) > 19) {
+		return catalogId[19:]
+	}
+	return ""
 }
 
 // Deletes the Catalog, returning an error if the vCD call fails.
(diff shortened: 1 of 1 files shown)


govcd/catalog.go
```diff
@@ -62,17 +62,17 @@ func NewAdminCatalog(client *Client) *AdminCatalog {
...
 // Link to API call: https://code.vmware.com/apis/220/vcloud#/doc/doc/operations/DELETE-Catalog.html
 func (catalog *Catalog) Delete(force, recursive bool) error {
-	adminCatalogHREF, err := url.ParseRequestURI(catalog.Catalog.HREF)
-	if err != nil {
-		return fmt.Errorf("error parsing admin catalog's href: %v", err)
-	}
+
+	adminCatalogHREF := catalog.client.VCDHREF
+	// slice only numeric part from ID:"urn:vcloud:catalog:97384890-180c-4563-b9b7-0dc50a2430b0"
+	adminCatalogHREF.Path += "/admin/catalog/" + catalog.Catalog.ID[19:]
```
lvirbalas: You have replaced the "[19:]" with some function in one of the previous PRs. Would be nice to have it here as well.

vbauzysvmware: I am reverting back previous implementation as it wasn't good enough, it worked only for admin user.

PR #146, opened 2019-01-16
Fix check for enabling Dhcp

vbauzysvmware:
Ref: https://github.com/terraform-providers/terraform-provider-vcd/issues/78

Check wasn't working cause entity always existed. Changed check against entities details.

Signed-off-by: Vaidotas Bauzys <vbauzys@vmware.com>


vbauzysvmware:
If I got the question - 34 line.

Issue #140, opened 2019-01-08
Test for edge gateway VPN missing task checking


dataclouder:
The test for edge gateway VPN  in [edgegateway_test.go](https://github.com/vmware/go-vcloud-director/blob/fe7558bda51923652bfda0ca570ea9d2edcecc78/govcd/edgegateway_test.go#L123)
is missing a few checks:

* The returning task is skipped
* The task keeps running in the background, but it is never checked.

The result is that we eventually get a semi-hidden error when we retrieve the edge gateway, and see that there is a task error. However, since the retrieval operation returns a status code `200`, the error gets overlooked, unless we analyze the logs.

The test needs to be rewritten with a proper VPN configuration. When properly checked, the test fails as follows:

```
START: edgegateway_test.go:92: TestVCD.Test_AddIpsecVPN
com.vmware.vcloud.api.presentation.service.BadRequestException: Cannot update edge gateway "gwdatacloud"
	at com.vmware.vcloud.networking.model.edge.ModelBasedGatewayManager.handleUpdateException(ModelBasedGatewayManager.java:253)
	at com.vmware.vcloud.networking.model.edge.ModelBasedGatewayManager.configureGatewayServicesAsync(ModelBasedGatewayManager.java:247)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.vmware.vcloud.common.service.OsgiServiceReferenceFactoryBean$3.invoke(OsgiServiceReferenceFactoryBean.java:276)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:179)
	at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:213)
	at com.sun.proxy.$Proxy305.configureGatewayServicesAsync(Unknown Source)
	at com.vmware.vcloud.management.system.NetworkServiceImpl.configureGatewayServicesTask(NetworkServiceImpl.java:3300)
	at com.vmware.vcloud.management.system.NetworkServiceImpl.executeTask(NetworkServiceImpl.java:2585)
[...]

PR #147, opened 2019-01-16
Fix Issue#140: incomplete VPN setup in EdgeGW test

dataclouder:
The test `Test_AddIpsecVPN` had two problems (issue #140):
* incomplete definition for VPN (only one side of the tunnel was
mentioned)
* Not checking the task returned by the VPN creation.
Therefore, the test was failing but nobody noticed.

This fix fills the VPN definition with the minimum needed fields
and monitors the result of the task to completion.

It also adds the necessary fields to the test configuration file

IMPORTANT: **DO NOT MERGE** too soon.

While this PR fixes a problem, it also introduces a different one. When this code is merged, the test will actually create a VPN. But the VPN is not deleted, because the Govcd library has not implemented the VPN deletion yet. A side effect of this successful operation is that now the test `Test_1to1Mappings` has a conflict, and fails. We can start reviewing this change, but this change cannot get in without a function that removes the VPN.

### DIFF
--- a/govcd/edgegateway.go
+++ b/govcd/edgegateway.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"os"
 	"regexp"
 	"strings"
 	"time"
@@ -725,6 +726,10 @@ func (eGW *EdgeGateway) AddIpsecVPN(ipsecVPNConfig *types.EdgeGatewayServiceConf
 	resp, err := checkResp(eGW.client.Http.Do(req))
 	if err != nil {
 		return Task{}, fmt.Errorf("error reconfiguring Edge Gateway: %s", err)
+	}
+
+	if os.Getenv("GOVCD_DEBUG") != "" {
+		util.Logger.Printf("Edge Gateway Service Configuration: %s\n", prettyEdgeGatewayServiceConfiguration(ipsecVPNConfig))
 	}
 
 	task := NewTask(eGW.client)
--- a/govcd/edgegateway.go
+++ b/govcd/edgegateway.go
@@ -737,3 +737,18 @@ func (eGW *EdgeGateway) AddIpsecVPN(ipsecVPNConfig *types.EdgeGatewayServiceConf
 	return *task, nil
 
 }
+
+// Removes an Edge Gateway VPN, by passing an empty configuration
+func (eGW *EdgeGateway) RemoveIpsecVPN() (Task, error) {
+	err := eGW.Refresh()
+	if err != nil {
+		fmt.Printf("error: %v\n", err)
+	}
+	ipsecVPNConfig := &types.EdgeGatewayServiceConfiguration{
+		Xmlns: "http://www.vmware.com/vcloud/v1.5",
+		GatewayIpsecVpnService: &types.GatewayIpsecVpnService{
+			IsEnabled: false,
+		},
+	}
+	return eGW.AddIpsecVPN(ipsecVPNConfig)
+}
--- a/govcd/edgegateway_test.go
+++ b/govcd/edgegateway_test.go
@@ -5,8 +5,10 @@
 package govcd
 
 import (
+	"fmt"
 	types "github.com/vmware/go-vcloud-director/types/v56"
 	. "gopkg.in/check.v1"
+	"os"
 )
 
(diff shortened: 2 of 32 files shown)


govcd/edgegateway_test.go
```diff
@@ -5,8 +5,10 @@
 package govcd
 
 import (
+	"fmt"
 	types "github.com/vmware/go-vcloud-director/types/v56"
```
vbauzysvmware: Please remove "types" and align imports according standard :-)

govcd/monitor.go
```diff
@@ -130,6 +130,14 @@ func prettyTask(task *types.Task) string {
 	return ""
 }
 
+func prettyEdgeGatewayServiceConfiguration(conf *types.EdgeGatewayServiceConfiguration) string {
```
vbauzysvmware: Please add comment that this returns json as itsn't mentioned in func name.

govcd/sample_govcd_test_config.json
```diff
@@ -27,11 +27,13 @@
...
 		"edgeGateway": "myedgegw",
 		"externalIp": "10.150.10.10",
+		"externalNetmask": "255.255.224.0",
 		"internalIp": "192.168.1.10",
-    "disk": {
-      "size": 1048576,
-      "sizeForUpdate": 1048576
-    }
+		"internalNetmask": "255.255.255.0",
+        "disk": {
```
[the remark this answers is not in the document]
dataclouder: Not "ran away", but properly indented now.

govcd/sample_govcd_test_config.json
```diff
@@ -27,11 +27,13 @@
 		},
 		"edgeGateway": "myedgegw",
 		"externalIp": "10.150.10.10",
+		"externalNetmask": "255.255.224.0",
 		"internalIp": "192.168.1.10",
-    "disk": {
-      "size": 1048576,
-      "sizeForUpdate": 1048576
-    }
+		"internalNetmask": "255.255.255.0",
```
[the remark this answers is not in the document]
dataclouder: I don't know how to get defaults that are good for all servers we use, or any servers that users may want to try. I am open to suggestions if such defaults exist.

dataclouder:
Added VPN removal function

govcd/monitor.go
dataclouder: Done

govcd/edgegateway.go
```diff
@@ -737,3 +737,18 @@ func (eGW *EdgeGateway) AddIpsecVPN(ipsecVPNConfig *types.EdgeGatewayServiceConf
 	return *task, nil
 
 }
+
+// Removes an Edge Gateway VPN, by passing an empty configuration
```
vbauzysvmware: Link to documentation would be good if available.

govcd/edgegateway_test.go
```diff
@@ -147,7 +147,13 @@ func (vcd *TestVCD) Test_AddIpsecVPN(check *C) {
...
 
+	// Configures VPN service
 	task, err := edge.AddIpsecVPN(ipsecVPNConfig)
 	err = task.WaitTaskCompletion()
 	check.Assert(err, IsNil)
+
+	// Removes VPN service
+	task, err = edge.RemoveIpsecVPN()
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
```
vbauzysvmware: Would be good to verify that Ipsec is removed querying edge configuration.

govcd/edgegateway.go
dataclouder: The documentation is thin about this operation. I'm open to suggestions.

govcd/edgegateway_test.go
dataclouder: Done

govcd/edgegateway.go
dataclouder: Nope.

govcd/sample_govcd_test_config.json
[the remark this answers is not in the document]
lvirbalas: Unless I'm missing the context here, but if we're talking about internal network, any default will work. Even the above "255.255.255.0" is fine.
We can leave this configuration in, but I'm concerned we're adding parameters which users don't really need to care about. We should strive to make test config simpler.

PR #151, opened 2019-02-05
Add check for vApp template to be valid

vbauzysvmware:
Ref: https://github.com/terraform-providers/terraform-provider-vcd/issues/157

In issue real error is that vApp template returned something but not fully what was expected and govcd panicked. According user issue was that vapp template(OVA) wasn't uploaded successfully. I couldn't recreate same error but got different `null pointer` exception when ova is partly uploaded. I added check that before using vApp template to check it status. Status `8` is parsed from responses - didn't find explanations in documentations. When vApp template was with error then status was `0`.

### DIFF
--- a/govcd/vapp.go
+++ b/govcd/vapp.go
@@ -100,6 +100,14 @@ func (vapp *VApp) Refresh() error {
 // acceptAllEulas - setting allows to automatically accept or not Eulas.
 func (vapp *VApp) AddVM(orgVdcNetworks []*types.OrgVDCNetwork, vappNetworkName string, vappTemplate VAppTemplate, name string, acceptAllEulas bool) (Task, error) {
 
+	if vappTemplate == (VAppTemplate{}) || vappTemplate.VAppTemplate == nil {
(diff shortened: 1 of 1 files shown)


govcd/vapp.go
```diff
@@ -100,6 +100,14 @@ func (vapp *VApp) Refresh() error {
 // acceptAllEulas - setting allows to automatically accept or not Eulas.
 func (vapp *VApp) AddVM(orgVdcNetworks []*types.OrgVDCNetwork, vappNetworkName string, vappTemplate VAppTemplate, name string, acceptAllEulas bool) (Task, error) {
 
+	if vappTemplate == (VAppTemplate{}) || vappTemplate.VAppTemplate == nil {
+		return Task{}, fmt.Errorf("vApp Template can not be empty")
+	}
+
+	if vappTemplate.VAppTemplate.Status != 8 { // status Ready
```
lvirbalas: Would be really good to confirm all available status codes and put them into comment.

vbauzysvmware: Comments added.

PR #152, opened 2019-02-06
Refactor of network functions

dn1s:
## Refactor of vapp AddVM() function and networking functions

https://github.com/terraform-providers/terraform-provider-vcd/issues/118

- c6b8031
- d330e38
- 38496ae
- 5827d65

### DIFF
--- a/govcd/vapp.go
+++ b/govcd/vapp.go
@@ -133,8 +133,8 @@ func (vapp *VApp) AddVM(networks []map[string]interface{}, vappTemplate VAppTemp
 				HREF: vappTemplate.VAppTemplate.Children.VM[0].HREF,
 				Name: name,
 			},
-			VMGeneralParams: &types.VMGeneralParams {
-				Name: name,
+			VMGeneralParams: &types.VMGeneralParams{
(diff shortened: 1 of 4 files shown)


vbauzysvmware:
Thank you for you contribution! I am glad to have active users :) And also I missed description regarding PR  from you regarding what's the idea behind refactoring, what you have spotted and what you improved. It help us to be inline with Idea.

dn1s:
My thinking here is we should try to do as much as we can in one api call (obviously without screwing up) so we can reduce the amount of api calls which change parameters on vapp/vm because each of them is taking up atleast 15 seconds in our environment. Additionally this pr is "required" for my https://github.com/dn1s/terraform-provider-vcd/tree/add_multi_nics_support branch for the terraform provider.

dn1s:
> Please fix the conflicts.
> Then, the new feature should have an explicit test that checks the added functionality.

Gonna add it later this day!

dn1s:
I'm gonna close this pr and try to achive the same thing with functions we have already (unfortunately this will be slower I guess because of more api call).

PR #156, opened 2019-02-07
Add vapp network dhcp

vbauzysvmware:
Reference: [terraform-providers/terraform-provider-vcd#97](https://github.com/terraform-providers/terraform-provider-vcd/issues/97)

Existing functionality needed to be enhance to support Terraform. 
* Fixed validation
* Added ability to ignore sending values, needs explicitly set that DHCP service should be configured - this way we can ignore default values passed down from other services/users and DHCP service configured but no enabled.

### DIFF
--- a/govcd/vapp.go
+++ b/govcd/vapp.go
@@ -1116,7 +1116,7 @@ func validateNetworkConfigSettings(networkSettings *VappNetworkSettings) error {
 	}
 
 	if networkSettings.Gateway == "" {
-		return errors.New("network gateway ip is missing")
+		return errors.New("network gateway IP is missing")
 	}
 
 	if networkSettings.NetMask == "" {
@@ -1155,7 +1155,7 @@ func (vapp *VApp) RemoveIsolatedNetwork(networkName string) (Task, error) {
 	}
 
 	if !isNetworkFound {
-		return Task{}, fmt.Errorf("network to remove: %s, isn't found", networkName)
+		return Task{}, fmt.Errorf("network to remove %s, wasn't found", networkName)
 	}
 
 	return updateNetworkConfigurations(vapp, networkConfigurations)
--- a/govcd/vapp.go
+++ b/govcd/vapp.go
@@ -34,14 +34,18 @@ func (vdcCli *VCDClient) NewVApp(client *Client) VApp {
 
 // struct type used to pass information for vApp network creation
 type VappNetworkSettings struct {
-	Name             string
-	Gateway          string
-	NetMask          string
-	DNS1             string
-	DNS2             string
-	DNSSuffix        string
-	GuestVLANAllowed bool
-	IPRange          []*types.IPRange
+	Name                 string
+	Gateway              string
+	NetMask              string
+	DNS1                 string
+	DNS2                 string
+	DNSSuffix            string
+	GuestVLANAllowed     bool
+	StaticIPRanges       []*types.IPRange
+	DHCPIsEnabled        bool
+	DHCPMaxLeaseTime     int
+	DHCPDefaultLeaseTime int
+	DHCPIPRange          *types.IPRange
 }
 
 // Returns the vdc where the vapp resides in.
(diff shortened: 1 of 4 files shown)


types/v56/types.go
```diff
@@ -257,13 +257,13 @@ type NetworkConfigSection struct {
 // Since: 0.9
 type NetworkConnection struct {
 	Network                 string `xml:"network,attr"`                      // Name of the network to which this NIC is connected.
-	NetworkConnectionIndex  int    `xml:"NetworkConnectionIndex"`            // Virtual slot number associated with this NIC. First slot number is 0.
 	NeedsCustomization      bool   `xml:"needsCustomization,attr,omitempty"` // True if this NIC needs customization.
-	ExternalIPAddress       string `xml:"ExternalIpAddress,omitempty"`       // If the network to which this NIC connects provides NAT services, the external address assigned to this NIC appears here.
```
lvirbalas: Why this has been removed?

govcd/vapp.go
```diff
@@ -43,6 +43,7 @@ type VappNetworkSettings struct {
 	DNSSuffix            string
 	GuestVLANAllowed     bool
 	StaticIPRanges       []*types.IPRange
+	AddDHCPService       bool
```
lvirbalas: 1. What is the use case of having `AddDHCPService` set to `false`?
2. Doesn't vApp network always have a DHCP service? It may be not configured, but isn't it always there?
3. We had issue with DHCP being not enabled which you fixed. Does it still work correctly with this PR? https://github.com/terraform-providers/terraform-provider-vcd/issues/78

govcd/vapp.go
```diff
@@ -1098,16 +1103,27 @@ func (vapp *VApp) AddIsolatedNetwork(newIsolatedNetworkSettings *VappNetworkSett
 		return Task{}, err
 	}
 
+	// for case when range is one ip address
```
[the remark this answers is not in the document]
vbauzysvmware: DHCP pool can consist of one IP address. Api requires start and end addresses. You can input just one IP address in UI.

types/v56/types.go
vbauzysvmware: It isn't, I shuffled to be inline with strict xml cause otherwise serialized messages isn't accepted by vCD.

govcd/vapp.go
vbauzysvmware: 1. We need this due GO limitations with default property initialization and we can't check if they are nil. So not send half empty xml DHCP service configuration I need explicitly state that DHCP info was added and marshalling should include these properties. 
2. According responses I got three states - not configured, configured disabled, configured enabled
3. This code part isn't used with edge gtw

govcd/vapp.go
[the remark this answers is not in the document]
vbauzysvmware: Improved comment

govcd/vapp.go
dataclouder: >     1. we can't check if they are nil.
Please elaborate this point.

vbauzysvmware: We can't identify when value wasn't provided or when it is init value for e.g. default int value will be 0 or user will provide value 0.

dataclouder: Why the DHCP properties can't be like `DHCPIPRange *types.IPRange`, which can be nil ?

vbauzysvmware: If you think it's better I could change struct to more levels and then it should work also.

dataclouder: It is not a matter of personal preferences here. In this library we use this approach several times. Using a `struct` pointer simplifies the marshaling, and you don't need to have an explicit `AddDHCPService`

lvirbalas: Nice suggestion!

vbauzysvmware:
Note: Current changes in functionalities are improvement/fixes. No new functions. And this old code doesn't have any tests before. I may try write some tests, but not sure it's quality.

govcd/vapp.go
vbauzysvmware: Changed code.

govcd/vm.go
```diff
@@ -241,50 +241,52 @@ func (vm *VM) ChangeNetworkConfig(networks []map[string]interface{}, ip string)
...
-			ipAllocationMode = "MANUAL"
-			// TODO: Check a valid IP has been given
-			ipAddress = network["ip"].(string)
-		} else if ip == "" {
-			ipAllocationMode = "DHCP"
+	networkSection, err := vm.GetNetworkConnectionSection()
+
+	for _, network := range networks {
+		for index, networkConnection := range networkSection.NetworkConnection {
+			if networkConnection.Network == network["orgnetwork"] {
```
lvirbalas: Is `network["orgnetwork"]` the name, ID or ... of the network? Really need a comment to mention what it represents.

govcd/vm.go
```diff
@@ -241,50 +241,52 @@ func (vm *VM) ChangeNetworkConfig(networks []map[string]interface{}, ip string)
...
-		} else if network["ip"].(string) != "" {
-			ipAllocationMode = "MANUAL"
-			// TODO: Check a valid IP has been given
-			ipAddress = network["ip"].(string)
-		} else if ip == "" {
-			ipAllocationMode = "DHCP"
+	networkSection, err := vm.GetNetworkConnectionSection()
+
+	for _, network := range networks {
+		for index, networkConnection := range networkSection.NetworkConnection {
```
lvirbalas: This double loop is tricky to read. The fact that `networks` is a map doesn't help either. Please add a human-oriented comment above that explains what is happening here.

govcd/vm.go
vbauzysvmware: Network name

govcd/vm.go
vbauzysvmware: Added requested comment

govcd/vm.go
vbauzysvmware: comment added

PR #158, opened 2019-02-12
Switch to go modules

dataclouder:
# Switch to go modules

* Remove vendor
* Remove vendor references in scripts
* Enable go modules
*  Fix Go formatting (it's changed slightly in Go 1.11)
* Update Travis and Jenkins scripts
* Update Travis to use modules explicitly


vbauzysvmware:
What about cache cleaning solution?

Issue #129, opened 2018-12-13
implement external network operations


dataclouder:
To make a VDC immediately useful, we need the ability of creating/deleting/updating an external network.

References: 
* [External Network type](https://code.vmware.com/apis/287/vcloud#/doc/doc/types/ExternalNetworkType.html)

lvirbalas:
Thank you for this issue, @dataclouder 
Please post a suggested Terraform resource definition too.

megalord:
I'll take a shot at this one!

PR #159, opened 2019-02-19
Create and delete for external networks

megalord:
## Description

Add methods for creating and deleting an external network.

Related issue: (#129 )

I still need to add a negative test on the create just to make sure it returns the correct error.  I can't validate the fields on the created network without implementing a GET method for it too, which I'm happy to do if you like.  If so I could use guidance on what to do with the existing method (`GetExternalNetworkByName`) since they'd have a similar purpose.

### DIFF
--- a/govcd/extension.go
+++ b/govcd/extension.go
@@ -13,7 +13,7 @@ import (
 	"github.com/vmware/go-vcloud-director/v2/util"
 )
 
-// Deprecated - use GetExternalNetworkByName2
+// Deprecated - use GetExternalNetwork
 func GetExternalNetworkByName(vcdClient *VCDClient, networkName string) (*types.ExternalNetworkReference, error) {
 	externalNetwork := NewExternalNetwork(&vcdClient.Client)
 	err := externalNetwork.GetByName(networkName)
@@ -27,7 +27,7 @@ func GetExternalNetworkByName(vcdClient *VCDClient, networkName string) (*types.
 	}, nil
 }
 
-func GetExternalNetworkByName2(vcdClient *VCDClient, networkName string) (*ExternalNetwork, error) {
+func GetExternalNetwork(vcdClient *VCDClient, networkName string) (*ExternalNetwork, error) {
 	externalNetwork := NewExternalNetwork(&vcdClient.Client)
 	err := externalNetwork.GetByName(networkName)
 	return externalNetwork, err
--- a/govcd/extension.go
+++ b/govcd/extension.go
@@ -13,6 +13,7 @@ import (
 	"github.com/vmware/go-vcloud-director/v2/util"
 )
 
+// Deprecated - use GetExternalNetworkByName2
 func GetExternalNetworkByName(vcdClient *VCDClient, networkName string) (*types.ExternalNetworkReference, error) {
 	externalNetwork := NewExternalNetwork(&vcdClient.Client)
 	err := externalNetwork.GetByName(networkName)
@@ -24,6 +25,12 @@ func GetExternalNetworkByName(vcdClient *VCDClient, networkName string) (*types.
 		Type: externalNetwork.ExternalNetwork.Type,
 		Name: externalNetwork.ExternalNetwork.Name,
 	}, nil
+}
+
+func GetExternalNetworkByName2(vcdClient *VCDClient, networkName string) (*ExternalNetwork, error) {
+	externalNetwork := NewExternalNetwork(&vcdClient.Client)
+	err := externalNetwork.GetByName(networkName)
+	return externalNetwork, err
 }
 
 func CreateExternalNetwork(vcdClient *VCDClient, externalNetwork *types.ExternalNetwork) (Task, error) {
--- a/govcd/extension.go
+++ b/govcd/extension.go
@@ -9,44 +9,22 @@ import (
 	"encoding/xml"
 	"errors"
 	"fmt"
-	"net/url"
 
 	"github.com/vmware/go-vcloud-director/v2/types/v56"
 	"github.com/vmware/go-vcloud-director/v2/util"
 )
 
 func GetExternalNetworkByName(vcdClient *VCDClient, networkName string) (*types.ExternalNetworkReference, error) {
-	extNetworkRefs := &types.ExternalNetworkReferences{}
-
-	extNetworkHREF, err := getExternalNetworkHref(vcdClient)
+	externalNetwork := NewExternalNetwork(&vcdClient.Client)
+	err := externalNetwork.GetByName(networkName)
 	if err != nil {
 		return &types.ExternalNetworkReference{}, err
 	}
-
-	extNetworkURL, err := url.ParseRequestURI(extNetworkHREF)
-	if err != nil {
-		return &types.ExternalNetworkReference{}, err
-	}
-
-	req := vcdClient.Client.NewRequest(map[string]string{}, "GET", *extNetworkURL, nil)
-	resp, err := checkResp(vcdClient.Client.Http.Do(req))
-	if err != nil {
-		util.Logger.Printf("[TRACE] error retrieving external networks: %s", err)
-		return &types.ExternalNetworkReference{}, fmt.Errorf("error retrieving external networks: %s", err)
-	}
-
-	if err = decodeBody(resp, extNetworkRefs); err != nil {
-		util.Logger.Printf("[TRACE] error retrieving  external networks: %s", err)
-		return &types.ExternalNetworkReference{}, fmt.Errorf("error decoding extension  external networks: %s", err)
-	}
-
-	for _, netRef := range extNetworkRefs.ExternalNetworkReference {
-		if netRef.Name == networkName {
-			return netRef, nil
-		}
-	}
-
-	return &types.ExternalNetworkReference{}, nil
+	return &types.ExternalNetworkReference{
+		HREF: externalNetwork.ExternalNetwork.HREF,
+		Type: externalNetwork.ExternalNetwork.Type,
+		Name: externalNetwork.ExternalNetwork.Name,
+	}, nil
 }
 
 func CreateExternalNetwork(vcdClient *VCDClient, externalNetwork *types.ExternalNetwork) (Task, error) {
@@ -82,8 +60,8 @@ func CreateExternalNetwork(vcdClient *VCDClient, externalNetwork *types.External
 	return *task, nil
 }
 
-func getExternalNetworkHref(vcdClient *VCDClient) (string, error) {
-	extensions, err := getExtension(vcdClient)
+func getExternalNetworkHref(client *Client) (string, error) {
+	extensions, err := getExtension(client)
 	if err != nil {
 		return "", err
 	}
@@ -97,13 +75,13 @@ func getExternalNetworkHref(vcdClient *VCDClient) (string, error) {
 	return "", errors.New("external network link isn't found")
 }
 
-func getExtension(vcdClient *VCDClient) (*types.Extension, error) {
+func getExtension(client *Client) (*types.Extension, error) {
 	extensions := &types.Extension{}
 
-	extensionHREF := vcdClient.Client.VCDHREF
+	extensionHREF := client.VCDHREF
 	extensionHREF.Path += "/admin/extension/"
-	req := vcdClient.Client.NewRequest(map[string]string{}, "GET", extensionHREF, nil)
-	resp, err := checkResp(vcdClient.Client.Http.Do(req))
+	req := client.NewRequest(map[string]string{}, "GET", extensionHREF, nil)
+	resp, err := checkResp(client.Http.Do(req))
 	if err != nil {
 		util.Logger.Printf("[TRACE] error retrieving extension: %s", err)
 		return extensions, fmt.Errorf("error retrieving extension: %s", err)
--- a/govcd/extension.go
+++ b/govcd/extension.go
@@ -5,6 +5,8 @@
 package govcd
 
 import (
+	"bytes"
+	"encoding/xml"
 	"errors"
 	"fmt"
 	"net/url"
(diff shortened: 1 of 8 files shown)


govcd/extension.go
```diff
@@ -47,6 +49,39 @@ func GetExternalNetworkByName(vcdClient *VCDClient, networkName string) (*types.
...
+	util.Logger.Printf("[TRACE] CreateExternalNetwork - xml payload: %s\n", xmlData)
+	externalnetsHREF := vcdClient.Client.VCDHREF
+	externalnetsHREF.Path += "/admin/extension/externalnets"
+	req := vcdClient.Client.NewRequest(map[string]string{}, "POST", externalnetsHREF, xmlData)
+	req.Header.Add("Content-Type", "application/vnd.vmware.admin.vmwexternalnet+xml")
+	resp, err := checkResp(vcdClient.Client.Http.Do(req))
+	if err != nil {
+		util.Logger.Printf("[TRACE] error instantiating a new ExternalNetwork: %s", err)
+		return Task{}, fmt.Errorf("error instantiating a new ExternalNetwork: %s", err)
+	}
```
vbauzysvmware: Line separation missing - hard to read this block

govcd/extension.go
```diff
@@ -47,6 +49,39 @@ func GetExternalNetworkByName(vcdClient *VCDClient, networkName string) (*types.
...
+
+	externalNetwork = new(types.ExternalNetwork)
+	if err = decodeBody(resp, externalNetwork); err != nil {
+		util.Logger.Printf("[TRACE] error decoding admin extension externalnets response: %s", err)
+		return Task{}, fmt.Errorf("error decoding admin extension externalnets response: %s", err)
+	}
+
+	task := NewTask(&vcdClient.Client)
+	task.Task = externalNetwork.Tasks.Task[0]
```
vbauzysvmware: Doesn't work directly parsing the task as we do in other places? for e.g.
```
	task := NewTask(eGW.client)

	if err = decodeBody(resp, task.Task); err != nil {
		return Task{}, fmt.Errorf("error decoding Task response: %s", err)
	}
```

govcd/extension_test.go
```diff
@@ -28,3 +30,82 @@ func (vcd *TestVCD) Test_GetExternalNetwork(check *C) {
 	check.Assert(externalNetwork.Name, Equals, networkName)
 	check.Assert(externalNetwork.Type, Equals, expectedType)
 }
+
+func (vcd *TestVCD) Test_CreateExternalNetwork(check *C) {
+	if vcd.skipAdminTests {
+		check.Skip("Configuration org != 'Sysyem'")
```
vbauzysvmware: Mistype in `System`

govcd/extension_test.go
```diff
@@ -28,3 +30,82 @@ func (vcd *TestVCD) Test_GetExternalNetwork(check *C) {
...
+	}
+	networkName := vcd.config.VCD.ExternalNetwork
```
vbauzysvmware: Why we need configured network name for test which test creation of ext network?

govcd/extension_test.go
```diff
@@ -28,3 +30,82 @@ func (vcd *TestVCD) Test_GetExternalNetwork(check *C) {
...
+	if networkName == "" {
+		check.Skip("No external network provided")
+	}
+
+	externalNetworkRef, err := GetExternalNetworkByName(vcd.client, networkName)
```
vbauzysvmware: Why we delete existing one external network?

govcd/extension_test.go
```diff
@@ -28,3 +30,82 @@ func (vcd *TestVCD) Test_GetExternalNetwork(check *C) {
...
+	if *externalNetworkRef != (types.ExternalNetworkReference{}) {
+		externalNetwork := NewExternalNetwork(&vcd.client.Client)
+		externalNetwork.ExternalNetwork = &types.ExternalNetwork{
+			HREF: externalNetworkRef.HREF,
+		}
+		err = externalNetwork.DeleteWait()
+		check.Assert(err, IsNil)
+	}
+
+	results, err := vcd.client.QueryWithNotEncodedParams(nil, map[string]string{
```
vbauzysvmware: Recommend to extract to specific function. We have similar for queering media, vappTemplate

govcd/externalnetwork.go
```diff
@@ -0,0 +1,88 @@
...
+	if err = decodeBody(resp, externalNetwork.ExternalNetwork); err != nil {
+		util.Logger.Printf("[TRACE] error decoding extension external network: %s", err)
+		return fmt.Errorf("error decoding extension external network: %s", err)
+	}
+	return nil
+}
+
+func validateExternalNetwork(externalNetwork *types.ExternalNetwork) error {
+	if externalNetwork.Name == "" {
+		return errors.New("VdcConfiguration missing required field: Name")
```
vbauzysvmware: Please update `VdcConfiguration` to External Nework

govcd/externalnetwork.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	if externalNetwork.Xmlns == "" {
+		return errors.New("VdcConfiguration missing required field: Xmlns")
```
vbauzysvmware: Please update `VdcConfiguration` to External Nework

types/v56/types.go
```diff
@@ -2140,6 +2158,33 @@ type ExternalNetworkReference struct {
 	Name string `xml:"name,attr,omitempty"`
 }
 
+type VimObjectRef struct {
```
vbauzysvmware: Please add comments and links to doc's for new structs.

govcd/extension_test.go
megalord: Ahh right, I forgot all the names for new resources are in `govdc/api_vcd_test.go` :+1:

govcd/extension_test.go
megalord: I got the idea from feedback on my previous pull request: https://github.com/vmware/go-vcloud-director/pull/117#discussion_r240193783

But it doesn't look like that change made it all the way through, so faulty memory on my part.  I'll remove this.

govcd/extension.go
```diff
@@ -47,6 +49,39 @@ func GetExternalNetworkByName(vcdClient *VCDClient, networkName string) (*types.
 	return &types.ExternalNetworkReference{}, nil
 }
 
+func CreateExternalNetwork(vcdClient *VCDClient, externalNetwork *types.ExternalNetwork) (Task, error) {
```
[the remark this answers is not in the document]
megalord: Correct, I don't see a technical reason that prevents us from adding it to the `AdminOrg` struct.

I'm still quite green to vmware concepts in general, but I don't yet see the connection between an org and an external network.  Other methods on the struct like `CreateCatalog` and `CreateVdc` operate on child resources of the org, but I'm not sure the same applies here.

I do agree that there is opportunity for improved convenience - I find the pattern of
```
type <Type> struct {
	<Type> *types.<Type>
	client   *Client
}
```
a bit cumbersome.  My first thought is to attach more methods to the `VCDClient`.

Am I getting too ahead of myself?  Ultimately I don't mind adding it to the `AdminOrg`.

govcd/extension.go
megalord: It does!

govcd/extension_test.go
megalord: I wasn't quite sure where to put it, but it is in `extension.go` now.

govcd/externalnetwork.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"errors"
+	"fmt"
+	"net/url"
+
+	"github.com/vmware/go-vcloud-director/types/v56"
+	"github.com/vmware/go-vcloud-director/util"
+)
+
+type ExternalNetwork struct {
```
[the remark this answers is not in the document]
megalord: I took a shot at this in a separate commit from the other changes.  I think I understood what you're after but let me know otherwise!

megalord:
Also FYI, I'm purposefully waiting to rebase this until the `vmware-stash-while-master-pr-open` branch of the provider is merged, because I'm juggling about five branches between the two repos (working ahead on a complete provider build).  Integrating the module changes while they are still in flight is too difficult to keep up with.

govcd/extension.go
vbauzysvmware: So I think it can be improved by decoding `Task` directly and avoid the not needed checks and getting from array.

govcd/extension_test.go
vbauzysvmware: I think in that case and in this case - right answer is "exit gracefully with a meaningful error message.". We don't delete what we don't expect to find I would say :)

govcd/extension.go
```diff
@@ -116,3 +116,15 @@ func getExtension(vcdClient *VCDClient) (*types.Extension, error) {
 
 	return extensions, nil
 }
+
+func queryVirtualCenters(vdcCli *VCDClient, filter string) ([]*types.QueryResultVirtualCenterRecordType, error) {
```
vbauzysvmware: This function should be public.

vbauzysvmware:
Yep, I agree what we had is not straight forward. Until weren't owners we couldn't do much. At the moment we stabilize  `vmware-stash-while-master-pr-open` - merge to master and then create or rename `vmware-stash-while-master-pr-open` to make explicitly clear which branch is mainstream.

govcd/extension.go
lvirbalas: Please also put a comment above public function explaining what it does.

govcd/externalnetwork.go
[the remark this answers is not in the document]
Didainius: Yep, thank you. I think it will be more convenient for future uses.

govcd/extension.go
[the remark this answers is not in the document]
Didainius: I'am a bit raw myself with this project and I agree with the struct type being a bit strange, but we can't break it (at least until 3.0) :)
And you're right. There's nothing an `organization administrator` can do with external networks. I made wrong assumption about `AdminOrg` myself.

govcd/extension.go
megalord: Agreed! I have already made that change in one of the newer commits.

vbauzysvmware:
We have agreement that PR are against `master` so no  `dev` branches at the moment.

govcd/extension.go
```diff
@@ -5,50 +5,63 @@
...
 	if err != nil {
-		util.Logger.Printf("[TRACE] error retrieving external networks: %s", err)
-		return &types.ExternalNetworkReference{}, fmt.Errorf("error retrieving external networks: %s", err)
+		return Task{}, fmt.Errorf("error marshalling xml: %s", err)
 	}
 
-	if err = decodeBody(resp, extNetworkRefs); err != nil {
-		util.Logger.Printf("[TRACE] error retrieving  external networks: %s", err)
-		return &types.ExternalNetworkReference{}, fmt.Errorf("error decoding extension  external networks: %s", err)
+	xmlData := bytes.NewBufferString(xml.Header + string(output))
```
vbauzysvmware: No need logging here, cause request and payload will be logged automatically.

govcd/extension.go
```diff
@@ -5,50 +5,63 @@
...
+	util.Logger.Printf("[TRACE] CreateExternalNetwork - xml payload: %s\n", xmlData)
+	externalnetsHREF := vcdClient.Client.VCDHREF
```
vbauzysvmware: Nit: 
```suggestion
	externalNetHREF := vcdClient.Client.VCDHREF
```

govcd/extension.go
```diff
@@ -5,50 +5,63 @@
...
+		return Task{}, fmt.Errorf("error decoding admin extension externalnets response: %s", err)
 	}
 
-	return &types.ExternalNetworkReference{}, nil
+	return *task, nil
 }
 
-func getExternalNetworkHref(vcdClient *VCDClient) (string, error) {
-	extensions, err := getExtension(vcdClient)
+func getExternalNetworkHref(client *Client) (string, error) {
```
vbauzysvmware: This function should be moved to externalNetwork.go as it used there.

govcd/extension.go
```diff
@@ -81,3 +94,16 @@ func getExtension(vcdClient *VCDClient) (*types.Extension, error) {
 
 	return extensions, nil
 }
+
+// Find a list of Virtual Centers matching the filter parameter.
+func QueryVirtualCenters(vdcCli *VCDClient, filter string) ([]*types.QueryResultVirtualCenterRecordType, error) {
```
vbauzysvmware: Please, move this function to org.go

govcd/extension_test.go
```diff
@@ -24,7 +25,69 @@ func (vcd *TestVCD) Test_GetExternalNetwork(check *C) {
...
+	check.Assert(err, IsNil)
+	AddToCleanupList(externalNetwork.Name, "externalNetwork", "", "Test_CreateExternalNetwork")
+	check.Assert(task, Not(Equals), Task{})
+
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
+
+	externalNetworkRef, err := GetExternalNetworkByName(vcd.client, TestCreateExternalNetwork)
+	check.Assert(err, IsNil)
+	check.Assert(externalNetworkRef.Name, Equals, TestCreateExternalNetwork)
```
vbauzysvmware: Test missing to check all other properties if they are set correctly.

govcd/externalnetwork.go
```diff
@@ -0,0 +1,122 @@
...
+	if err = decodeBody(resp, externalNetwork.ExternalNetwork); err != nil {
+		util.Logger.Printf("[TRACE] error decoding extension external network: %s", err)
+		return fmt.Errorf("error decoding extension external network: %s", err)
+	}
+	return nil
+}
+
+func validateExternalNetwork(externalNetwork *types.ExternalNetwork) error {
+	if externalNetwork.Name == "" {
+		return errors.New("External Network missing required field: Name")
```
vbauzysvmware: Nit: error should start with lower case

govcd/externalnetwork.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	if externalNetwork.Xmlns == "" {
+		return errors.New("External Network missing required field: Xmlns")
```
vbauzysvmware: Nit: error should start with lower case

govcd/extension.go
```diff
@@ -5,50 +5,63 @@
...
+	"encoding/xml"
 	"errors"
 	"fmt"
-	"net/url"
 
 	"github.com/vmware/go-vcloud-director/v2/types/v56"
 	"github.com/vmware/go-vcloud-director/v2/util"
 )
 
 func GetExternalNetworkByName(vcdClient *VCDClient, networkName string) (*types.ExternalNetworkReference, error) {
```
vbauzysvmware: I would like that we follow our current API behaviour and return external network struct vs `ExternalNetworkReference`.

megalord: Ok just to make sure we're on the same page - you're asking me for a breaking change here, right?  Specifically, someone using this library would bump the version of this lib in their app and the code may not compile due to a type mismatch.

Is there a versioning strategy for this repo?  I couldn't find anything in the markdown files.  As a generalization though, I'll refer to the golang FAQ:

> Packages intended for public use should try to maintain backwards compatibility as they evolve.

and

> If different functionality is required, add a new name instead of changing an old one.

https://golang.org/doc/faq#get_version

vbauzysvmware: Well if this function existed - we have to support it due backwards compatibility. Solution is easy, deprecate this function and let's create wrapping function which will return network struct.

vbauzysvmware: P.s. Regarding versioning - we follow https://semver.org/

megalord: Ok, so maybe I'm not quite following what you mean, but that's pretty much how the code looks now.
```go
	externalNetwork := NewExternalNetwork(&vcdClient.Client)
	err := externalNetwork.GetByName(networkName)
```
Would you prefer I put those two lines in new method?

megalord: Ok, so maybe I'm not quite following what you mean, but that's pretty much how the code looks now.
```go
	externalNetwork := NewExternalNetwork(&vcdClient.Client)
	err := externalNetwork.GetByName(networkName)
```
Would you prefer I put those two lines in new method?

vbauzysvmware: In terraform you used a little different approach
````
externalNetwork := govcd.NewExternalNetwork(&vcdClient.VCDClient.Client)
externalNetwork.ExternalNetwork.HREF = externalNetworkRef.HREF
err = externalNetwork.Refresh()
```
Aim is to avoid such constructions, cause user need good knowledge of govcd for that. So answer is yes - wrapper could just has a few lines which allows to return right construct with externalNetwork which can be used directly with API functions.

megalord: Sure sounds good.  I'm not too particular on the name, so let me know if you'd like to change it.

vbauzysvmware: Well, this one is hard :) - Let's do it simple -> "getExternalNetwork"

megalord: Sweet can do.  I think it needs to be a public method, no?

vbauzysvmware: Yes, definitely :) ant let's add `// DEPRECATED please use getExternalNetwork function instead` for old one

vbauzysvmware:
Work continue on https://github.com/vmware/go-vcloud-director/pull/185

PR #162, opened 2019-02-28
Preparation for multi nic support

dn1s:
Removes redundant ip param from ChangeNetworkConfig() as you can see here how the function is used in https://github.com/terraform-providers/terraform-provider-vcd/blob/master/vcd/resource_vcd_vapp_vm.go:

```
			networks := []map[string]interface{}{map[string]interface{}{
				"ip":         d.Get("ip").(string),
				"is_primary": true,
				"orgnetwork": netName,
			}}
			task, err := vm.ChangeNetworkConfig(networks, d.Get("ip").(string))
```

Related issue: https://github.com/terraform-providers/terraform-provider-vcd/issues/118

@vbauzysvmware 
@dataclouder 
@lvirbalas


lvirbalas:
@dn1s please join project's Slack channel #vcd-terraform-dev in vmwarecode.slack.com workspace, as it would be good to discuss some of the changes you're proposing in more depth.

dn1s:
Sure but you would need to invite me can't join by myself ;)

dn1s:
@lvirbalas Sure but you would need to invite me can't join by myself ;)

vbauzysvmware:
Is this PR still valid ass #118 has much more changes required here?, IMHO

govcd/vm.go
```diff
@@ -252,22 +253,19 @@ func (vm *VM) ChangeNetworkConfig(networks []map[string]interface{}, ip string)
 				ipAddress := "Any"
 
 				// TODO: Review current behaviour of using DHCP when left blank
-				if ip == "dhcp" || network["ip"].(string) == "dhcp" {
+				if network["ip"].(string) == "dhcp" {
```
Didainius: This is pretty confusing to compute on the brain.
I've tried to automated input/output testing for this piece of code and it seems to me that the output's are different which may break someone's (not necessarily terraform provider) functionality.
Here's the sample code: https://gist.github.com/Didainius/6c8b8cdc75d57d3af9bc16190b742c7b
And it's output in the bottom (you can notice if `ip` is empty you get different results).

Can you verify if it's my mistake or difference in outputs.

dn1s: I guess you did not supply somthing for network["ip_allocation_mode"].(string) when you left ip 
 and network["ip"].(string) empty right?

dn1s:
> Is this PR still valid ass #118 has much more changes required here?, IMHO

Don't really understand what you mean with that. Could you please point to some code which I had to change?
For me these changes work and I'm using 27.0 api.

govcd/vm.go
dn1s: Well to have the exact same behaviour as before I would need to check if network["ip_allocation_mode"].(string) is empty and if it is apply dhcp otherwise apply network["ip_allocation_mode"].(string).

vbauzysvmware:
Sorry, I provided wrong link. I mean if this inline with changes needed for https://github.com/terraform-providers/terraform-provider-vcd/pull/177

vbauzysvmware:
Could also add comment for function `ChangeNetworkConfig` it's expects maps of values.

vbauzysvmware:
I also think we should use this opportunity and make function better with 'func (vm *VM) ChangeNetworkConfig(networks []map[string]interface{})'  -> `func (vm *VM) ChangeNetworkConfig(networks []NetworkSettins)`. This would allow use struct and explicitly define expected values.  To back compatible change current method just to be `wrapper` of new function signature.

govcd/vm.go
Didainius: Looks like you're right. I'd like that function to still be 100% compatible with previous. Looking at codebase for terraform it didn't set `network["ip_allocation_mode"].(string)` before triggering this function so while I'm pretty fresh to this codebase myself, I believe it may break things for someone. Another way could be what vbauzysvmware mentioned below - introduce a new function and wrap it in this one for backwards compatibility.

dn1s: Behavior is fixed.

govcd/vapp.go
```diff
@@ -168,7 +166,7 @@ func (vapp *VApp) AddVM(orgVdcNetworks []*types.OrgVDCNetwork, vappNetworkName s
 			&types.NetworkConnection{
 				Network:                 vappNetworkName,
 				NetworkConnectionIndex:  len(orgVdcNetworks),
-				IsConnected:             true,
+				IsConnected:             false,
```
Didainius: Same question as above:
>  Why was this changed? Wouldn't it break behaviour from previous version?

govcd/vapp.go
```diff
@@ -136,10 +136,8 @@ func (vapp *VApp) AddVM(orgVdcNetworks []*types.OrgVDCNetwork, vappNetworkName s
 			},
 			InstantiationParams: &types.InstantiationParams{
 				NetworkConnectionSection: &types.NetworkConnectionSection{
-					Type:                          vappTemplate.VAppTemplate.Children.VM[0].NetworkConnectionSection.Type,
```
Didainius: Can this break functionality if someone depends on it using the SDK?

govcd/vm.go
```diff
@@ -247,44 +309,45 @@ func (vm *VM) ChangeCPUCountWithCore(virtualCpuCount int, coresPerSocket *int) (
...
+			// changes network config only if network name matches with provided one and if order is correct (this is needed
+			// for example if you attach NICs from same network)
+			if networkConnection.Network == network["orgnetwork"] && index == nicSlot {
 				// Determine what type of address is requested for the vApp
 				ipAllocationMode := "NONE"
 				ipAddress := "Any"
 
 				// TODO: Review current behaviour of using DHCP when left blank
-				if ip == "dhcp" || network["ip"].(string) == "dhcp" {
+				if network["ip"].(string) == "dhcp" {
```
Didainius: Can you doublecheck my testcode piece again. I still get some difference and 
https://play.golang.org/p/Wrf169V7Znd

dn1s: In my opinion this behavior was broken and now it's fixed. If you add an actual ip to your test you will see how this improves the code. ONLY if ip is valid and not dhcp allocated or none it will get to MANUAL otherwise just use DHCP as fallback
```
==== Before ====
dhcp 		Any 		DHCP
allocated 		Any 		POOL
none 		Any 		NONE
 		Any 		DHCP
RANDOM 		RANDOM 		MANUAL
127.0.0.1 		127.0.0.1 		MANUAL
==== After2 ====
dhcp 		Any 		DHCP
allocated 		Any 		POOL
none 		Any 		NONE
 		Any 		
RANDOM 		Any 		DHCP
127.0.0.1 		127.0.0.1 		MANUAL
```

dn1s: The only thing I'd consider changing is:
```
else {
		ipAllocationMode = network["ip_allocation_mode"].(string)
	}
```
could be:
```
else {
		ipAllocationMode = "DHCP"
	}
if network["ip_allocation_mode"].(string) {
		ipAllocationMode = network["ip_allocation_mode"].(string)
}
```

govcd/vapp.go
dn1s: Reverted.

govcd/vapp.go
```diff
@@ -151,7 +149,7 @@ func (vapp *VApp) AddVM(orgVdcNetworks []*types.OrgVDCNetwork, vappNetworkName s
 			&types.NetworkConnection{
 				Network:                 orgVdcNetwork.Name,
 				NetworkConnectionIndex:  index,
-				IsConnected:             true,
+				IsConnected:             false,
```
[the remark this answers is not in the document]
dn1s: Reverted.

govcd/vapp.go
dn1s: Might break it for someone eventually. But it also fixes #213 and #26

govcd/vm.go
```diff
@@ -247,44 +309,45 @@ func (vm *VM) ChangeCPUCountWithCore(virtualCpuCount int, coresPerSocket *int) (
 
 }
 
-func (vm *VM) ChangeNetworkConfig(networks []map[string]interface{}, ip string) (Task, error) {
+func (vm *VM) ChangeNetworkConfig(networks []map[string]interface{}) (Task, error) {
```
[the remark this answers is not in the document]
dn1s: I could also not remove ip param and just not use it. Would be the same.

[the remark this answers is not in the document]
Didainius: Ok, or even that way and add a comment on top that the ip is redundant. We will clean up such things as we come to 3.0, but it's too early for 3.0 now.

govcd/vapp.go
Didainius: Ok. Fair enough. Could you put a line near some notes in changelog that it is changed and a short why. That gives someone perspective.

govcd/vm.go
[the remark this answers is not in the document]
lvirbalas: I concur, we are working towards a 2.2.0, so cannot afford backward-incompatible breakages. Thanks.

Didainius:
Work continues in #188

Issue #139, opened 2019-01-07
Add Semantic Versioning tags to repository


dataclouder:
The current tag for the repository is `v1.0.0`, which is outdated.

As we have introduced new functionalities and some incompatibilities, we need to tag the current repo as `v2.0.0`, as the basis for terraform-provider-vcd.

After the tag is established, it must be updated at every change:

* If there are no API changes (i.e. bug fixing, code cleanup, etc), we increase the revision number (e.g. from `v2.0.5` to `v2.0.6`).

* If we introduce a new feature, without breaking compatibility with the previous version, we increment the minor number (e.g. from `v2.0.6` to `v2.1.0`).

* If we break compatibility, meaning that one or more **public** identifiers or signatures are changed, we need to bump up the major number (e.g. from `v2.1.0` to `v3.0.0`).

These tags are needed to simplify the import of modules from `terraform-provider-vcd`, which should be upgraded to use Go 1.11 and switch from using the `vendor` folder to leveraging [`go modules`](https://github.com/golang/go/wiki/Modules).

lvirbalas:
> If we introduce a new feature, without breaking compatibility with the previous version, we increment the minor number (e.g. from v2.0.6 to v2.1.0).

Bumping this up with each commit seems overwhelming, given the amount of work being put in.

> If we introduce a new feature, without breaking compatibility with the previous version, we increment the minor number (e.g. from v2.0.6 to v2.1.0).

Same comment as above, just as a sample we may introduce ten new features pretty quickly. Would this mean we end up having 2.10.0?

> If we break compatibility, meaning that one or more public identifiers or signatures are changed, we need to bump up the major number (e.g. from v2.1.0 to v3.0.0).

In general, I agree, but I disagree that this should happen instantly after commit. We should "accumulate" some changes and only then increase the major number, otherwise we'll end up with v42.0.0 in no time.

dataclouder:
> We should  "accumulate" some changes and only then increase the major number, otherwise we'll end up with v42.0.0 in no time.

This is not how the system works. A major number increase happens only on incompatible changes, which we should try to avoid. The recommendation of the Go community on the usage of modules is to change the import path for an incompatible change (from `v.1.x` to `v2.0`). Therefore, a `v42.0.0` is very unlikely, while a perfectly acceptable accumulation of changes could result in `v2.42.0`, meaning that this tag can be used by someone who is only using functions introduced in `v2.0.0`.

[...]

PR #163, opened 2019-03-01
Switch module to v2

Didainius:
Switching project module to `v2` to match semver tags. This should then allow to migrate `terraform-provider-vcd` for modules support. It does three things:

* Changes module name definition in `go.mod`.
* Adds  all the imports from  `v2`.
* Adds Go 1.12.x for Travis testing as it is already released.

It should be tagged as `v2.1.0-alpha.2` or similar after merge to move forwards with `terraform-provider-vcd` module support.

@lvirbalas @vbauzysvmware @dataclouder 

Related issues: ( #139, https://github.com/terraform-providers/terraform-provider-vcd/issues/162 )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: go
 
 go:
-    - "1.11.x"
     - "1.12.x"
 
 sudo: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: go
 
 go:
     - "1.11.x"
+    - "1.12.x"
 
 sudo: false
 install: false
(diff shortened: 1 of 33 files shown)


Didainius:
Sorry for early PR. I'm working to introduce another PR on terraform-provider-vcd which should demonstrate how this whole puzzle works.

Didainius:
> Please check if everything is fine with formatting. I recall that Giuseppe showed a several places where `fmt` change code. Otherwise build complain

Giuseppe cleaned it up here in his previous commit 3f06ee3.

`gofmt` now doesn't complain on 1.11 and 1.12 (does complain on 1.10)
Now this opens wider conversation if or not `make fmtcheck` should be there or not, because in short go team does not freeze gofmt at the moment and working with different versions of go may cause `fmtcheck` complain. More there - https://github.com/golang/go/issues/22695#issuecomment-344483632, but this excerpt is important:

> Yes, there is a second half to this, which is that local presubmit checks are fine (and should use the installed gofmt) but post-submit CI failures are typically not. It should not be the case that updating the gofmt on the CI machine starts making some kind of automated "gofmt-ed" test fail. Our position is don't write tests like that. Treat running gofmt over your source as best-effort during submit but not something to be reverified automatically after submit.

vbauzysvmware:
I am ok with mandatory check, as keeps are code formatting correct. Simple changes after version, dodn't impact much. One or line code changes, doesn't concern.

vbauzysvmware:
Is vendor folder recreated?

PR #165, opened 2019-03-06
Add new api for independent disk query by name.

vbauzysvmware:
Ref: https://github.com/terraform-providers/terraform-provider-vcd/issues/73

Add new API to query by disk name. Need to implement terraform independent resource.

### DIFF
--- a/govcd/disk.go
+++ b/govcd/disk.go
@@ -22,10 +22,24 @@ type Disk struct {
 	client *Client
 }
 
+// Independent disk query record
+type DiskRecord struct {
+	Disk   *types.DiskRecordType
+	client *Client
+}
+
 // Init independent disk struct
 func NewDisk(cli *Client) *Disk {
 	return &Disk{
 		Disk:   new(types.Disk),
+		client: cli,
+	}
+}
+
+// create instance with reference to types.DiskRecordType
+func NewDiskRecord(cli *Client) *DiskRecord {
+	return &DiskRecord{
+		Disk:   new(types.DiskRecordType),
 		client: cli,
 	}
 }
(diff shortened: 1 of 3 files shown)


govcd/disk.go
```diff
@@ -380,3 +394,32 @@ func FindDiskByHREF(client *Client, href string) (*Disk, error) {
 	// Return the disk
 	return disk, nil
 }
+
+// Find vm using vApp name and VM name. Returns VMRecord query return type
+func (vdc *Vdc) QueryDisk(diskName string) (DiskRecord, error) {
+
```
Didainius: Validate if diskName is not empty as it's done in `func (vdc *Vdc) QueryVM(vappName, vmName string)`

govcd/disk.go
```diff
@@ -380,3 +394,32 @@ func FindDiskByHREF(client *Client, href string) (*Disk, error) {
 	// Return the disk
 	return disk, nil
 }
+
+// Find vm using vApp name and VM name. Returns VMRecord query return type
```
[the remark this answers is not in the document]
vbauzysvmware: Fixed

govcd/disk.go
vbauzysvmware: Fixed

govcd/disk.go
```diff
@@ -30,6 +36,14 @@ func NewDisk(cli *Client) *Disk {
 	}
 }
 
+// create instance with reference to types.DiskRecordType
```
lvirbalas: Nit: begin comments with capital letter.

vbauzysvmware: Fixed

PR #167, opened 2019-03-06
Add ability to provide cores then configuring CPUs

vbauzysvmware:
Ref: terraform-providers/terraform-provider-vcd#174

Add ability to provide core number when change CPU configuration.

### DIFF
--- a/govcd/vapp.go
+++ b/govcd/vapp.go
@@ -594,11 +594,18 @@ func (vapp *VApp) GetNetworkConnectionSection() (*types.NetworkConnectionSection
 	return networkConnectionSection, nil
 }
 
-func (vapp *VApp) ChangeCPUcount(size int) (Task, error) {
-	return vapp.ChangeCPUCountWithCore(size, nil)
+// Sets number of available virtual logical processors
+// (i.e. CPUs x cores per socket)
+// https://communities.vmware.com/thread/576209
+func (vapp *VApp) ChangeCPUcount(virtualCpuCount int) (Task, error) {
+	return vapp.ChangeCPUCountWithCore(virtualCpuCount, nil)
 }
 
-func (vapp *VApp) ChangeCPUCountWithCore(size int, cores *int) (Task, error) {
+// Sets number of available virtual logical processors
+// (i.e. CPUs x cores per socket) and cores per socket.
+// Socket count is a result of: virtual logical processors/cores per socket
+// https://communities.vmware.com/thread/576209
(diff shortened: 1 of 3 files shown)


vbauzysvmware:
Our code didn't assign any cores - > didn't send using API. I think vCD automatically set some value or take from vappTempate. So you can get isue when let's say vappTemplate has 4 cpu 4 cores and you adjust with property to 1 CPU and VCD throws error something likes this: "Virtual machine &quot;Photon OS-IOKU&quot; cannot support 1 CPUs using 2 cores per socket. CPU count must be a multiple of cores per socket" - I am not sure how this cpu/cores algorithm requirements are, but this code should allow user to adjust both of them and provide validate changes with property values. Terraform change will be needed to.

Didainius:
Not directly related to this one, but I see references to `VM[0]`. Doesn't this mean that if someone sets parameters on vApp level we will actually only adjust the first VM inside vApp and others will stay as they are?

vbauzysvmware:
Yes you are right, we have a few places with links to [0] which has to be fixed. As so some of resources and limited to one recource and with `count ` >1 doesn't work. In this case we need to change overall - CPU change is has to be done for vm and not for vapp.

Didainius:
True. Probably we can deprecate together with 1VM within vApp method.

vbauzysvmware:
Write down a BUG for spotted thing https://github.com/vmware/go-vcloud-director/issues/168

vbauzysvmware:
My point is that this change allows user to config cpu/cores - not only cpus. Current code haven't and didn't add any validation to check what values you pass and if they right in regards cpu/cores formula. It's for user to decide. We can add some comment and link to documentation if you found any regarding formula. In future we can think about validation functionality, but it will be tricky cause you have to check total system cpu/core count and current VM cpu/core count to be sure that values passed are right. Currently user will receive error from vCD response if cpu/core values aren't valid.

lvirbalas:
No need for validation (we may shoot us in the foot - let's keep that for underlying vSphere), but please add that comment and merge.

spengilley:
> Our code didn't assign any cores - > didn't send using API. I think vCD automatically set some value or take from vappTempate. So you can get isue when let's say vappTemplate has 4 cpu 4 cores and you adjust with property to 1 CPU and VCD throws error something likes this: "Virtual machine "Photon OS-IOKU" cannot support 1 CPUs using 2 cores per socket. CPU count must be a multiple of cores per socket" - I am not sure how this cpu/cores algorithm requirements are, but this code should allow user to adjust both of them and provide validate changes with property values. Terraform change will be needed to.

That's exactly what happened. I had a vAppTemplate with 2 cores.

PR #170, opened 2019-03-20
Fix admin in url

vbauzysvmware:
Ref: https://github.com/terraform-providers/terraform-provider-vcd/issues/167

Issue is with links like `http://admin.url.com/api/admin/org`. Govcd code has to jump from admin to normal (or vice versa)user API calls and has functionality which removes `/admin` part from url. So if general URL part also has `admin` it's removed together and server isn't found anymore. 
Fix are done through code which more strict regarding what url part to remove.


vbauzysvmware:
@Didainius  Could you run test with your env if everything is fine?

Didainius:
> @Didainius Could you run test with your env if everything is fine?

go-vcloud-director:
```
OK: 88 passed
PASS
ok      github.com/vmware/go-vcloud-director/v2/govcd   1407.169s
```

terraform-provider-vcd:
```
ok      github.com/terraform-providers/terraform-provider-vcd/v2/vcd    1735.700s
```

vbauzysvmware:
Could you elaborate ? - As far as I understand only possible thing is extract from changed function two lines of code and create new function and write test for them. And yet we almost will test if standart go functionality `strings.Split` works fine.

vbauzysvmware:
> > @Didainius Could you run test with your env if everything is fine?
> 
> go-vcloud-director:
> 
> ```
> OK: 88 passed
> PASS
> ok      github.com/vmware/go-vcloud-director/v2/govcd   1407.169s
> ```
> 
> terraform-provider-vcd:
> 
> ```
> ok      github.com/terraform-providers/terraform-provider-vcd/v2/vcd    1735.700s
> ```

Great! Thank you

PR #171, opened 2019-03-20
Add cpu cores support for VM and deprecated func for vApp

vbauzysvmware:
Ref: terraform-providers/terraform-provider-vcd#174

Add ability to provide core number when change CPU configuration. 
* vApp function deprecated
* add missing functions for VM

### DIFF
--- a/govcd/vm.go
+++ b/govcd/vm.go
@@ -177,6 +177,7 @@ func (vm *VM) PowerOff() (Task, error) {
 
 // Sets number of available virtual logical processors
 // (i.e. CPUs x cores per socket)
+// Cpu cores count is inherited from template.
 // https://communities.vmware.com/thread/576209
 func (vm *VM) ChangeCPUCount(virtualCpuCount int) (Task, error) {
 	return vm.ChangeCPUCountWithCore(virtualCpuCount, nil)
--- a/govcd/vm.go
+++ b/govcd/vm.go
@@ -175,7 +175,18 @@ func (vm *VM) PowerOff() (Task, error) {
 
 }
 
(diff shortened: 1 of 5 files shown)


vbauzysvmware:
It's old code. I don't know. I just changed that makes it deprecated to remove code from vApp which is not illogic.

govcd/vm.go
```diff
@@ -175,7 +175,18 @@ func (vm *VM) PowerOff() (Task, error) {
 
 }
 
-func (vm *VM) ChangeCPUcount(size int) (Task, error) {
+// Sets number of available virtual logical processors
+// (i.e. CPUs x cores per socket)
```
Didainius: Probably worth mentioning that number of cores is inherited from template, using this function.

govcd/vm_test.go
```diff
@@ -632,3 +643,60 @@ func (vcd *TestVCD) Test_AnswerVmQuestion(check *C) {
...
+				check.Assert(item.CoresPerSocket, Equals, cores)
+				check.Assert(item.VirtualQuantity, Equals, cpuCount)
+				foundItem = true
+				break
+			}
+		}
+		check.Assert(foundItem, Equals, true)
+	}
+
+	// return tu previous value
```
Didainius: Typo
```suggestion
	// return to previous value
```

govcd/vm.go
vbauzysvmware: added

govcd/vm_test.go
vbauzysvmware: fixed

vbauzysvmware:
Notes added/

PR #174, opened 2019-04-08
Introduce vCD API version checks

### LINKED ISSUE #173: Introduce version detection for client
In some cases different vCD versions behave differently and to protect in some cases we need to determine vCD version in the client to behave conditionally.

One of such examples is independent disk description used in Test `Test_QueryDisk`:
On 8.2: 
```
START: disk_test.go:571: TestVCD.Test_QueryDisk
Running: TestVCD.Test_QueryDisk
disk_test.go:621:
    check.Assert(diskRecord.Disk.Description, Equals, diskCreateParamsDisk.Description)
... obtained string = ""
... expected string = "TestQueryDisk"

FAIL: disk_test.go:571: TestVCD.Test_QueryDisk
```
On vCD ~~9.x~~ 9.5 it works fine.

The general idea is to determine vCD version on authentication and store it in client so that it can be used throughout code base.

Didainius:
Closes #173

As we already spotted (#173) sometimes even the same API version on vCD has some differences which caused test suite failures. This PR introduces ability to check on API supported versions using Hashicorp's `go-version` package in a friendly manner: "> 27", "=27", ">= 27", etc. It does not directly allow to match vCD product versions (9.x.x), because there is no API to get them.

Exported API introduces two methods:
*  ~~`APIMaxVerIs`~~`APIVCDMaxVersionIs` (allows to compare against vCD Maximum supported version and is useful to understand which vCD version is running https://code.vmware.com/doc/preview?id=8072) 
* ~~`APICurVerIs`~~`APIClientVersionIs` (allows to compare against currently used API version)

Also introduced a configuration option `WithAPIVersion(version string) VCDClientOption` which allows to easily override API version used


govcd/api_vcd_versions.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+type VersionInfos []VersionInfo
+
+type SupportedVersions struct {
+	VersionInfos `xml:"VersionInfo"`
+}
+
+// APIMaxVerIs allows to compare against maximum vCD supported API version. It will always
+// return false until the client has not done Authenticate(). Can be useful to validate
```
lvirbalas: Why do we need to authenticate first? Aren't versions returned even without authenticating?

govcd/api_vcd_versions.go
```diff
@@ -0,0 +1,169 @@
...
+	return isSupported
+}
+
+// APICurVerIs allows to compare against currently used API version. Can be useful to validate
+// if a certain feature can be used or not.
+//
+// Format: ">= 27.0, < 32.0", ">= 30.0", "= 27.0"
+//
+// vCD version mapping to API version support https://code.vmware.com/doc/preview?id=8072
+func (vdcCli *VCDClient) APICurVerIs(versionConstraint string) bool {
```
lvirbalas: `APIMaxVerIs` is clear, but I find purpose of `APICurVerIs` confusing. A deeper description may be helpful for someone who is seeing these functions first time.

> // APICurVerIs allows to compare against currently used API version
Used by what? Isn't `APIMaxVerIs` the currently used one?

Didainius: I see this is not clear. Trying to rephrase.

* `APICurVerIs` compares against  currently used API version `VCDClient.Client.APIVersion` variable (the API versions which client is set to use).

* `APIMaxVerIs` compares against maximum supported version from `/api/versions` (not necessarily the currently used one). This allows to check what is the maximum API version that vCD instance supports and can be used to guess vCD product version. Like API `31.0` support was first introduced in vCD 9.5 (as per https://code.vmware.com/doc/preview?id=8072). Therefore`if APIMaxVerIs(">= 31.0")` implies that you have vCD 9.5 or newer running inside.

govcd/api_vcd_versions.go
Didainius: They are, but I wasn't willing to do additional call. This rings some bells and I will think how to make them work unauthenticated while storing this info to avoid multiple calls.

govcd/api_vcd_versions.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// APIMaxVerIs allows to compare against maximum vCD supported API version. It will always
+// return false until the client has not done Authenticate(). Can be useful to validate
+// what vCD version is actually running.
+//
+// Format: ">= 27.0, < 32.0", ">= 30.0", "= 27.0"
+//
+// vCD version mapping to API version support https://code.vmware.com/doc/preview?id=8072
+func (vdcCli *VCDClient) APIMaxVerIs(versionConstraint string) bool {
```
dataclouder: It should be `vcdCli`, not `vdcCli`.
This error is present in many methods through the library. It would be nice to fix all of them.

govcd/api_vcd_versions.go
lvirbalas: Your suggested descriptions are indeed clearer! Now, I'm a proponent of making things clear from function names where possible, so here's a suggestion growing on yours. What about renaming the function as follows:
`APICurVerIs` -> `APICliVerIs `

dataclouder: Even better `APIClientVersionIs`, as `Cli` may not always convey the right idea. Let's make it clear that it's the client version and not anything else, such as a `CLI` (command line interface).

govcd/api_vcd_versions.go
Didainius: Ok. It seemed confusing and now I know why. I will fix it either in the end of this PR (to avoid many altered files in PR) or in a separate one.

govcd/api_vcd_versions.go
lvirbalas: I like this suggestion (and the commit that just came with it), but then let's make the other function name consistent in style as well:

- `APIMaxVerIs` -> `APIVCDVersionIs` ?
- `APICurVerIs` -> `APIClientVersionIs` (done)

Didainius: Still I'd like some word like `Max` or similar to mention that we're comparing against maximum/biggest version as vCD returns more than supported version. `APIVCDMaxVersionIs`?

lvirbalas: Fantastic! It turns out to be even of same length :))

`APIVCDMaxVersionIs`
`APIClientVersionIs`

govcd/api_vcd_versions.go
Didainius: I have changed this to work without authentication.

govcd/api_vcd_versions.go
vbauzysvmware: I agree due one more reason - using full words ant not shortened ones `Ver`

govcd/api_vcd.go
```diff
@@ -19,38 +19,30 @@ import (
...
 	}
-	defer resp.Body.Close()
 
-	supportedVersions := new(supportedVersions)
-	err = decodeBody(resp, supportedVersions)
-	if err != nil {
-		return fmt.Errorf("error decoding versions response: %s", err)
+	// find login address matching the API version
+	var neededVersion VersionInfo
+	for _, v := range vdcCli.supportedVersions.VersionInfos {
```
vbauzysvmware: Let's try to avoid one letter variables

govcd/api_vcd_versions.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	util.Logger.Printf("[TRACE] checking max API version against constraints '%s'", versionConstraint)
+	maxVersion, err := vdcCli.maxSupportedVersion()
+	if err != nil {
+		util.Logger.Printf("[ERROR] unable to find max supported version : %s", err)
+		return false
+	}
+
+	isSupported, err := vdcCli.apiVerMatchesConstraint(maxVersion, versionConstraint)
```
vbauzysvmware: I would like to use full words `Ver` -> `Version`

govcd/api_vcd_versions.go
```diff
@@ -0,0 +1,192 @@
...
+	vdcCli.supportedVersions = *suppVersions
+
+	return nil
+}
+
+// maxSupportedVersion parses supported version list and returns the highest version in string format.
+func (vdcCli *VCDClient) maxSupportedVersion() (string, error) {
+	versions := make([]*semver.Version, len(vdcCli.supportedVersions.VersionInfos))
+	for i, raw := range vdcCli.supportedVersions.VersionInfos {
+		v, _ := semver.NewVersion(raw.Version)
```
vbauzysvmware: Let's avoid using one letter variables

govcd/api_vcd_versions.go
```diff
@@ -0,0 +1,192 @@
...
+// vcdCheckSupportedVersion checks if there is at least one specified version exactly matching listed ones.
+// Format example "27.0"
+func (vdcCli *VCDClient) vcdCheckSupportedVersion(version string) (bool, error) {
+	return vdcCli.checkSupportedVersionConstraint(fmt.Sprintf("= %s", version))
+}
+
+// Checks if there is at least one specified version matching the list returned by vCD.
+// Constraint format can be in format ">= 27.0, < 32",">= 30" ,"= 27.0".
+func (vdcCli *VCDClient) checkSupportedVersionConstraint(versionConstraint string) (bool, error) {
+	for _, vi := range vdcCli.supportedVersions.VersionInfos {
```
vbauzysvmware: what this `vi` is?

govcd/api_vcd_versions.go
vbauzysvmware: If APIClientVersionIs is check against hardcoded value in of our client? why ist's fetching versions? - or I got something wrong?

govcd/api_vcd.go
Didainius: I can change these, but that's idiomatic Go for limited scope vars - https://github.com/golang/go/wiki/CodeReviewComments#variable-names

govcd/api_vcd_versions.go
Didainius: single 'version info'. Like 
`for 'version info' in range vdcCli.supportedVersions.VersionInfos`

govcd/api_vcd_versions.go
Didainius: Done

govcd/api_vcd_versions.go
Didainius: > If APIClientVersionIs is check against hardcoded value in of our client? why ist's fetching versions? - or I got something wrong?

Very good catch. Removed that.

govcd/api_vcd.go
dataclouder: We prefer to avoid them: when we inherited the library, we had to understand what the variables in the loops meant. We saw the same `v` used to refer to a vcd, a vdc, a version, a vapp, a vm, and it took us some time to bring the algorithms to the light. 
It's better to invest a few letters to make the code readable without effort.

Didainius: 2 vs 1 - agreed 👍

vbauzysvmware: Oh, Giuseppe was faster ;-)

govcd/api_vcd_versions.go
vbauzysvmware: Would be nice if name would be explicit :)

govcd/api_vcd_versions_test.go
```diff
@@ -0,0 +1,125 @@
...
+	// Minimum supported vCD 8.20 introduced API version 27.0
+	r := vcd.client.APIVCDMaxVersionIs(">= 27.0")
+	check.Assert(r, Equals, true)
+
+	mockVcd := getMockVcdWithAPIVersion("27.0")
+
+	var versionTests = []struct {
+		version      string
+		boolChecker  Checker
+		isSsupported bool
```
dataclouder: typo: `isSsupported` should be `isSupported`

PR #178, opened 2019-04-11
Go lint for a couple of files and comment not commented public functions

dn1s:
I recently started using vscode and it has a cool go-lint highlighting. Those are just a couple of files which I touched. Let me know if you like it that way or not :)
Also checkout vscode for yourself if you haven't yet.

### DIFF
--- a/govcd/vapp.go
+++ b/govcd/vapp.go
@@ -17,11 +17,13 @@ import (
 	"github.com/vmware/go-vcloud-director/v2/util"
 )
 
+// VApp Struct representing a vApp with it's client
 type VApp struct {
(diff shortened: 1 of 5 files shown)


lvirbalas:
Thank you for the intention, but in my opinion the auto-changed comments are less readable in this suggestion and overall the timing is not good for this change - we have bigger PRs cooking that add value and need to be merged first.

BTW, was this one auto-fixed (good catch)?
```
- return Task{}, fmt.Errorf("error: %v\n", err)
+ return Task{}, fmt.Errorf("error: %v", err)
```

PR #183, opened 2019-04-18
Add changes to correct NAT api function behaviour

vbauzysvmware:
Ref: https://github.com/terraform-providers/terraform-provider-vcd/pull/224

This small changes fixes issue with NAT api functions. New functions requires network which will be used to apply rule. Currently implementation would take first network form edge gtw.

* deprecate old functions
* add property to new functions to require network
* update unit tests

### DIFF
--- a/govcd/edgegateway.go
+++ b/govcd/edgegateway.go
@@ -181,7 +181,7 @@ func (eGW *EdgeGateway) AddNATRule(network *types.OrgVDCNetwork, natType, extern
 	return eGW.AddNATPortMappingWithUplink(network, natType, externalIP, "any", internalIP, "any", "any", "")
 }
 
-// Deprecated: Use eGW.AddNetworkNATMapping()
+// Deprecated: Use eGW.AddNATRule()
 func (eGW *EdgeGateway) AddNATMapping(natType, externalIP, internalIP string) (Task, error) {
 	return eGW.AddNATPortMapping(natType, externalIP, "any", internalIP, "any", "any", "")
 }
--- a/govcd/edgegateway.go
+++ b/govcd/edgegateway.go
@@ -177,14 +177,25 @@ func (eGW *EdgeGateway) RemoveNATPortMapping(natType, externalIP, externalPort s
 
 }
 
+func (eGW *EdgeGateway) AddNetworkNATMapping(network *types.OrgVDCNetwork, natType, externalIP, internalIP string) (Task, error) {
+	return eGW.AddNetworkNATPortMapping(network, natType, externalIP, "any", internalIP, "any", "any", "")
+}
+
+func (eGW *EdgeGateway) AddNetworkNATPortMapping(network *types.OrgVDCNetwork, natType, externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType string) (Task, error) {
+	return eGW.AddNATPortMappingWithUplink(network, natType, externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType)
+}
+
+// Deprecated: Use eGW.AddNetworkNATMapping()
 func (eGW *EdgeGateway) AddNATMapping(natType, externalIP, internalIP string) (Task, error) {
 	return eGW.AddNATPortMapping(natType, externalIP, "any", internalIP, "any", "any", "")
 }
 
+// Deprecated: Use eGW.AddNetworkNATPortMapping()
 func (eGW *EdgeGateway) AddNATPortMapping(natType, externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType string) (Task, error) {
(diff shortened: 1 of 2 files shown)


vbauzysvmware:
> Where is the new `AddNetworkNATPortMapping` function?

In 184 Line

lvirbalas:
> > Where is the new `AddNetworkNATPortMapping` function?
> 
> In 184 Line

Sorry, somehow missed that last evening. though now have a different question:

Existing function:
```
func (eGW *EdgeGateway) AddNATPortMappingWithUplink(network *types.OrgVDCNetwork, natType, externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType string) (Task, error)
```
New function:
```
func (eGW *EdgeGateway) AddNetworkNATPortMapping(network *types.OrgVDCNetwork, natType, externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType string) (Task, error)
```

What is the purpose of adding another function with same parameters but a bit different name?

vbauzysvmware:
> > > Where is the new `AddNetworkNATPortMapping` function?
> > 
> > 
> > In 184 Line
> 
> Sorry, somehow missed that last evening. though now have a different question:
> 
> Existing function:
> `func (eGW *EdgeGateway) AddNATPortMappingWithUplink(network *types.OrgVDCNetwork, natType, externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType string) (Task, error)`
> New function:
> `func (eGW *EdgeGateway) AddNetworkNATPortMapping(network *types.OrgVDCNetwork, natType, externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType string) (Task, error)`
> 
> What is the purpose of adding another function with same parameters but a bit different name?

Yep, good spot. Fixed

govcd/edgegateway.go
```diff
@@ -177,14 +177,21 @@ func (eGW *EdgeGateway) RemoveNATPortMapping(natType, externalIP, externalPort s
...
+func (eGW *EdgeGateway) AddNetworkNATMapping(network *types.OrgVDCNetwork, natType, externalIP, internalIP string) (Task, error) {
+	return eGW.AddNATPortMappingWithUplink(network, natType, externalIP, "any", internalIP, "any", "any", "")
+}
+
+// Deprecated: Use eGW.AddNetworkNATMapping()
 func (eGW *EdgeGateway) AddNATMapping(natType, externalIP, internalIP string) (Task, error) {
 	return eGW.AddNATPortMapping(natType, externalIP, "any", internalIP, "any", "any", "")
 }
 
+// Deprecated: Use eGW.AddNetworkNATPortMapping()
```
lvirbalas: I think this function doesn't exist.

govcd/edgegateway.go
```diff
@@ -177,14 +177,21 @@ func (eGW *EdgeGateway) RemoveNATPortMapping(natType, externalIP, externalPort s
 
 }
 
+func (eGW *EdgeGateway) AddNetworkNATMapping(network *types.OrgVDCNetwork, natType, externalIP, internalIP string) (Task, error) {
+	return eGW.AddNATPortMappingWithUplink(network, natType, externalIP, "any", internalIP, "any", "any", "")
```
lvirbalas: I'm a bit confused, why don't we just use the existing `AddNATPortMappingWithUplink` as opposed to defining the new `AddNetworkNATMapping`?

govcd/edgegateway.go
vbauzysvmware: Fixed

govcd/edgegateway.go
vbauzysvmware: Idea of wrapper is to have simple function against full bloated - as it hides - user don't need what to send to values which you don't use : (network, natType, externalIP, `"any"`, internalIP, `"any"`, `"any"`, `""`)

lvirbalas: Great (and my bad - on a hurry missed that this wrapper has a lot less params).

Last comment: what about renaming to ‘AddNATWithNetwork’, so we have all public NAT functions with same ‘AddNAT...’ beginning?

vbauzysvmware:
> No big comments I guess. As long as we still have compatibility - it's fine.
> Probably the test could check that the rule is actually created and has the specified attributes.
> Also this sparks a thought how do we start prepping a list of what can be removed for 3.0 or do we just rely on // Deprecated tag

Existing tests improved

govcd/edgegateway.go
vbauzysvmware: Well I don't like notation `withNetwork`  cause by default it is bug without network.
I changed to: `AddNATRule` - I hope it's better

lvirbalas: It’s great! But the comments still point to old name ;)

vbauzysvmware: Yep, fixed:)

PR #184, opened 2019-04-25
Add tags to all test files

dataclouder:
Test files are now all built conditionally using tags.
Running the tests without tags will trigger an error message
with a list of available tags (`cd govcd && go test -v .`).
The tags allow users to run only a subset of the tests, or all of them.
The `Makefile` uses the tags to run the same tests as before (tag `gocheck`),
and a few tags to run tests related to several features (`make
testcatalog`, `make testvapp`, make `testnetwork`).

The split by tags allows also running several sets of tests, one based
on go-check (our current default) and one based on regular Go tests
(which may not require a live vCD).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ default: fmtcheck vet build
 # test runs the test suite and vets the code
 test: fmtcheck
 	@echo "==> Running Tests"
-	cd govcd && go test -tags "functional" -timeout=45m -check.vv .
+	cd govcd && go test -tags "functional" -timeout=60m -check.vv .
 
 # testrace runs the race checker
 testrace:
@@ -24,9 +24,14 @@ testcatalog:
 testvapp:
 	cd govcd && go test -tags "vapp vm" -timeout 25m -check.vv .
 
-# tests only networking and edge gateway features
+# tests only edge gateway features
+testgateway:
+	cd govcd && go test -tags "gateway" -timeout 15m -check.vv .
+
+# tests only networking features
 testnetwork:
 	cd govcd && go test -tags "network" -timeout 15m -check.vv .
+
 
 # vet runs the Go source code static analysis tool `vet` to find
 # any common errors.
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ default: fmtcheck vet build
 # test runs the test suite and vets the code
 test: fmtcheck
 	@echo "==> Running Tests"
-	cd govcd && go test -timeout=45m -check.vv .
+	cd govcd && go test -tags "gocheck" -timeout=45m -check.vv .
 
 # testrace runs the race checker
 testrace:
@@ -14,7 +14,19 @@ testrace:
 
 # This will include tests guarded by build tag concurrent with race detector
 testconcurrent:
-	cd govcd && go test -race -tags "concurrent" -check.vv -check.f "Test.*Concurrent" .
+	cd govcd && go test -race -tags "api concurrent" -timeout 15m -check.vv -check.f "Test.*Concurrent" .
+
+# tests only catalog related features
+testcatalog:
+	cd govcd && go test -tags "catalog" -timeout 15m -check.vv .
+
+# tests only vapp and vm features
+testvapp:
+	cd govcd && go test -tags "vapp vm" -timeout 25m -check.vv .
+
+# tests only networking and edge gateway features
+testnetwork:
+	cd govcd && go test -tags "network" -timeout 15m -check.vv .
 
 # vet runs the Go source code static analysis tool `vet` to find
 # any common errors.
@@ -40,5 +52,5 @@ copyright:
 
 build:
 	@echo "==> Building govcd library"
-	cd govcd && go build . && go test -c .
+	cd govcd && go build . && go test -tags ALL -c .
 
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ default: fmtcheck vet build
 # test runs the test suite and vets the code
 test: fmtcheck
 	@echo "==> Running Tests"
-	cd govcd && go test -tags "gocheck" -timeout=45m -check.vv .
+	cd govcd && go test -tags "functional" -timeout=45m -check.vv .
 
 # testrace runs the race checker
 testrace:
--- a/TESTING.md
+++ b/TESTING.md
@@ -153,13 +153,13 @@ Tests that integrate in the functional suite use the tag `functional`. Using tha
 at once.
 We define as `functional` the tests that need a live vCD to run.
 
-Note that as of today (April 2019) we only have functional tests, but we plan to add more, which can be, for example,
+Note that as of today we only have functional tests, but we plan to add more, which can be, for example,
 `unit` tests (will test internal assumptions without using a vCD), or tests that take much time (such as performance,
 endurance, or memory leakage) or require repeated tests over a long period. For this reason, the set of tests that we
 want to run always is the functional suite. For everything else we need to decide whether it's safe or desirable to run
 them together or separately.
 
-The build tag line should also contain the tag `ALL` which will run all the tests, non only the functional suite.
+The build tag line should also contain the tag `ALL` which will run all the tests, not only the functional suite.
 This includes tests that use a different framework. At the moment, this is useful to run a global compilation test.
 Depending on which additional tests we will implement, we may change the dependency on the `ALL` tag if we detect
 clashes between frameworks.
--- a/TESTING.md
+++ b/TESTING.md
@@ -149,11 +149,28 @@ All tests need to have a build tag. The tag should be the first line of the file
 package govcd
 ```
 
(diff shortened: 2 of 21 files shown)


dataclouder:
I used `gocheck` because I want to keep the door open to use other frameworks. The intention is to use `gotest` for tests that are based on the standard go tests.
We are also used the `ALL` tag to run everything.
That said, if we have better names in mind, I am open to changing them.

dataclouder:
Changed main tags names as agreed.
Added tagging management notes to TESTING.md

dataclouder:
Also added a note in CHANGELOG

Makefile
```diff
@@ -6,15 +6,27 @@ default: fmtcheck vet build
 # test runs the test suite and vets the code
 test: fmtcheck
 	@echo "==> Running Tests"
-	cd govcd && go test -timeout=45m -check.vv .
+	cd govcd && go test -tags "functional" -timeout=45m -check.vv .
```
lvirbalas: `make test` should run all tests, not only functional.

TESTING.md
```diff
@@ -23,7 +23,7 @@ To run tests with go use these commands:
 
 ```bash
 cd govcd
-go test -check.v .
+go test -tags "functional" -check.v .
```
lvirbalas: It feels better to write up how to run all the tests here, because a newcomer may ignore some tests if he only runs functional.

TESTING.md
```diff
@@ -90,6 +139,29 @@ This entity is initialized when the test starts. You can assume that the variabl
...
+
+All tests need to have a build tag. The tag should be the first line of the file, followed by a blank line
+
+```go
+// +build functional featurename ALL
+
+package govcd
+```
+
+Tests that integrate in the existing suite use the tag `functional`. Using that tag, we can run all functional tests at once.
```
lvirbalas: The paragraph about this "existing suite" is confusing. If I'm a newcomer what is "existing"? It sort of implies there's some "non-existing" tests that one doesn't know of. It would be much more useful to describe, that functional tests are the ones that require vCD connection, while unit tests - don't. In other words, describe the types of tests as opposed to history.

I suggest we have three bulleted paragraphs here:
- That all test need to have either `functional` or `unit` (currently, the confusing/missing part)
- All should have `ALL` (already described)
- About the "feature" tags (already described)

govcd/api_vcd_test.go
```diff
@@ -625,3 +634,61 @@ func TestVCDClient_Authenticate(t *testing.T) {
 		t.Fatalf("Error authenticating: %v", err)
 	}
 }
+
+func (vcd *TestVCD) createTestVapp(name string) (VApp, error) {
```
lvirbalas: Curious, why this test moved from vApp test to API test?

Makefile
dataclouder: > `make test` should run all tests, not only functional.

This is intentional. For now, we only have tests that use the gocheck library. When we start using a different framework (of which we have already some samples in our personal repos), we don't want to mix them, because both frameworks have some "magic" that starts the tests, and we want to avoid possible conflicts.
The purpose of `make test` is to run the functional tests without `TF_ACC`, just to see if the tests runs.
If we want to check that the tests compile, running `make` will do it, as it will compile using the `ALL` tag

TESTING.md
dataclouder: As mentioned above, `functional` means "all the tests using the gocheck library".
One of the reason for this set of changes is that we want to run tests on different frameworks without fear of clashes. For now we don't have such tests, but it will be soon. When we do, we will add instructions about them.

Makefile
lvirbalas: What we need is an easy and safe way to run all the tests. How to do that?

govcd/api_vcd_test.go
dataclouder: This function is needed to create the initial vApp. 
If we leave it in `vapp_test.go`, every time we run a test we need to include vApp, even if we don't need it. With this change, we can isolate tests better. When vApp is not requested, the vApp creation gets skipped and the tests are faster.

Makefile
dataclouder: How about we get to that point when we have non `functional` tests?
By then, we will know what kind of tests we want to implement, how to enable them, and whether they can run together or separately. 
This framework has paved the way to add more tests safely.
If we design the next tests cleanly, we can run them with two separate commands, possibly in parallel.

dataclouder:
> 3.) Another question here: I assume you'll use this framework for parallelising test runs, please share your thoughts on how would that work.

Running tests in parallel should work by using separate processes, each using a different tag.
For example:

* Process 1 `go test -tags catalog -check.vv . `
* Process 2 `go test -tags 'vapp vdc vm' -check.vv . `
* Process 3 `go test -tags network -check.vv . `

Each process needs to use a different log name, or we should run the tests from three different directories.

TESTING.md
dataclouder: I have rephrased the paragraph to describe the scope and usage of the framework.

TESTING.md
```diff
@@ -90,6 +139,46 @@ This entity is initialized when the test starts. You can assume that the variabl
...
+at once.
+We define as `functional` the tests that need a live vCD to run.
+
+Note that as of today (April 2019) we only have functional tests, but we plan to add more, which can be, for example,
+`unit` tests (will test internal assumptions without using a vCD), or tests that take much time (such as performance,
+endurance, or memory leakage) or require repeated tests over a long period. For this reason, the set of tests that we
+want to run always is the functional suite. For everything else we need to decide whether it's safe or desirable to run
+them together or separately.
+
+The build tag line should also contain the tag `ALL` which will run all the tests, non only the functional suite.
```
lvirbalas: Typo: "non" -> "not"

TESTING.md
```diff
@@ -90,6 +139,46 @@ This entity is initialized when the test starts. You can assume that the variabl
...
+// +build functional featurename ALL
+
+package govcd
+```
+
+Tests that integrate in the functional suite use the tag `functional`. Using that tag, we can run all functional tests
+at once.
+We define as `functional` the tests that need a live vCD to run.
+
+Note that as of today (April 2019) we only have functional tests, but we plan to add more, which can be, for example,
```
lvirbalas: Please remove "(April 2019)" - it will get old quickly. "as of today" is enough ;)

TESTING.md
lvirbalas: @Didainius I believe you have some non-functional tests cooking? Would they work if executed together with the `functional` ones?

TESTING.md
dataclouder: fixed

TESTING.md
dataclouder: fixed

Didainius:
@lvirbalas yes, they would as they work on their own data structures and not rely on network connections at all:
https://github.com/Didainius/go-vcloud-director/blob/dainius-multinic/govcd/vm_test.go#L1109
https://github.com/Didainius/go-vcloud-director/blob/dainius-multinic/govcd/vm_test.go#L1009

Also some of the tests in API version check would be able to run like that as well. https://github.com/Didainius/go-vcloud-director/blob/dainius-multinic/govcd/api_vcd_versions_test.go

lvirbalas:
You are probably planning to merge these together with the multi-NIC PR? If that's the case, then this would mean that your tests would *not* get executed with this PR's change to `make test`.

dataclouder:
> You are probably planning to merge these together with the multi-NIC PR? If that's the case, then this would mean that your tests would _not_ get executed with this PR's change to `make test`.

These tests use the gocheck framework. 
They can run together with the functional tests, or separately, depending on how they are tagged.

Didainius:
I am conscious I will need to make adjustments after this PR is merged.

TESTING.md
```diff
@@ -41,6 +41,55 @@ cd govcd
...
+    api_test.go:59: # No tags were defined
+    api_test.go:46:
+        # -----------------------------------------------------
+        # Tags are required to run the tests
+        # -----------------------------------------------------
+
+        At least one of the following tags should be defined:
+
+           * ALL :       Runs all the tests
+           * functional: Runs all the tests that use check.v1
```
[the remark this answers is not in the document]
lvirbalas: Thanks for adding it!

TESTING.md
```diff
@@ -90,6 +139,46 @@ This entity is initialized when the test starts. You can assume that the variabl
...
+// +build functional featurename ALL
+
+package govcd
+```
+
+Tests that integrate in the functional suite use the tag `functional`. Using that tag, we can run all functional tests
+at once.
+We define as `functional` the tests that need a live vCD to run.
+
+Note that as of today we only have functional tests, but we plan to add more, which can be, for example,
```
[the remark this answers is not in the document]
lvirbalas: OK, we agreed that you will adjust these once Dainius merges his tests.

TESTING.md
lvirbalas: OK, we agreed that either @dataclouder or @Didainius will add these new tests to the the `make test` once they are ready.

PR #185, opened 2019-04-30
Add external network

vbauzysvmware:
Based by Megalord work: https://github.com/vmware/go-vcloud-director/pull/159

Related issue: (#129 )

Add methods for creating, geting and deleting an external network.

### DIFF
--- a/govcd/api_vcd_test.go
+++ b/govcd/api_vcd_test.go
@@ -53,6 +53,7 @@ const (
 	TestVMAttachOrDetachDisk      = "TestVMAttachOrDetachDisk"
 	TestVMAttachDisk              = "TestVMAttachDisk"
 	TestVMDetachDisk              = "TestVMDetachDisk"
+	TestCreateExternalNetwork     = "TestCreateExternalNetwork"
 )
 
 const (
@@ -88,13 +89,16 @@ type TestConfig struct {
 			SP1 string `yaml:"storageProfile1"`
 			SP2 string `yaml:"storageProfile2,omitempty"`
 		} `yaml:"storageProfile"`
-		ExternalIp      string `yaml:"externalIp,omitempty"`
-		ExternalNetmask string `yaml:"externalNetmask,omitempty"`
-		InternalIp      string `yaml:"internalIp,omitempty"`
-		InternalNetmask string `yaml:"internalNetmask,omitempty"`
-		EdgeGateway     string `yaml:"edgeGateway,omitempty"`
-		ExternalNetwork string `yaml:"externalNetwork,omitempty"`
-		Disk            struct {
+		ExternalIp                   string `yaml:"externalIp,omitempty"`
+		ExternalNetmask              string `yaml:"externalNetmask,omitempty"`
+		InternalIp                   string `yaml:"internalIp,omitempty"`
+		InternalNetmask              string `yaml:"internalNetmask,omitempty"`
+		EdgeGateway                  string `yaml:"edgeGateway,omitempty"`
+		ExternalNetwork              string `yaml:"externalNetwork,omitempty"`
+		ExternalNetworkPortGroup     string `yaml:"externalNetworkPortGroup,omitempty"`
+		ExternalNetworkPortGroupType string `yaml:"externalNetworkPortGroupType,omitempty"`
+		VimServer                    string `yaml:"vimServer,omitempty"`
+		Disk                         struct {
 			Size          int `yaml:"size,omitempty"`
 			SizeForUpdate int `yaml:"sizeForUpdate,omitempty"`
 		}
@@ -434,6 +438,19 @@ func (vcd *TestVCD) removeLeftoverEntities(entity CleanupEntity) {
 			return
 		}
 		err = RemoveOrgVdcNetworkIfExists(vdc, entity.Name)
+		if err == nil {
+			vcd.infoCleanup(removedMsg, entity.EntityType, entity.Name, entity.CreatedBy)
+		} else {
+			vcd.infoCleanup(notDeletedMsg, entity.EntityType, entity.Name, err)
+		}
+		return
+	case "externalNetwork":
+		externalNetwork, err := GetExternalNetwork(vcd.client, entity.Name)
+		if err != nil {
+			vcd.infoCleanup(notFoundMsg, "externalNetwork", entity.Name)
+			return
+		}
+		err = externalNetwork.DeleteWait()
 		if err == nil {
 			vcd.infoCleanup(removedMsg, entity.EntityType, entity.Name, entity.CreatedBy)
 		} else {
--- a/govcd/api_vcd_test.go
+++ b/govcd/api_vcd_test.go
@@ -54,6 +54,7 @@ const (
 	TestVMAttachDisk              = "TestVMAttachDisk"
 	TestVMDetachDisk              = "TestVMDetachDisk"
 	TestCreateExternalNetwork     = "TestCreateExternalNetwork"
+	TestDeleteExternalNetwork     = "TestDeleteExternalNetwork"
 )
 
 const (
--- a/govcd/extension.go
+++ b/govcd/extension.go
@@ -5,15 +5,15 @@
 package govcd
 
 import (
-	"errors"
 	"github.com/vmware/go-vcloud-director/v2/types/v56"
 	"net/http"
 )
 
+// DEPRECATED please use GetExternalNetwork function instead
 func GetExternalNetworkByName(vcdClient *VCDClient, networkName string) (*types.ExternalNetworkReference, error) {
 	extNetworkRefs := &types.ExternalNetworkReferences{}
 
-	extNetworkHREF, err := getExternalNetworkHref(vcdClient)
+	extNetworkHREF, err := getExternalNetworkHref(&vcdClient.Client)
 	if err != nil {
 		return &types.ExternalNetworkReference{}, err
 	}
@@ -33,28 +33,33 @@ func GetExternalNetworkByName(vcdClient *VCDClient, networkName string) (*types.
 	return &types.ExternalNetworkReference{}, nil
 }
 
-func getExternalNetworkHref(vcdClient *VCDClient) (string, error) {
-	extensions, err := getExtension(vcdClient)
-	if err != nil {
-		return "", err
-	}
-
-	for _, extensionLink := range extensions.Link {
-		if extensionLink.Type == "application/vnd.vmware.admin.vmwExternalNetworkReferences+xml" {
-			return extensionLink.HREF, nil
-		}
-	}
-
-	return "", errors.New("external network link isn't found")
+func GetExternalNetwork(vcdClient *VCDClient, networkName string) (*ExternalNetwork, error) {
+	externalNetwork := NewExternalNetwork(&vcdClient.Client)
+	err := externalNetwork.GetByName(networkName)
+	return externalNetwork, err
 }
 
-func getExtension(vcdClient *VCDClient) (*types.Extension, error) {
+func CreateExternalNetwork(vcdClient *VCDClient, externalNetwork *types.ExternalNetwork) (Task, error) {
+	err := validateExternalNetwork(externalNetwork)
+	if err != nil {
+		return Task{}, err
+	}
+
+	externalNetHREF := vcdClient.Client.VCDHREF
+	externalNetHREF.Path += "/admin/extension/externalnets"
+
+	// Return the task
+	return vcdClient.Client.ExecuteTaskRequest(externalNetHREF.String(), http.MethodPost,
+		types.MimeExternalNetwork, "error instantiating a new ExternalNetwork: %s", externalNetwork)
+}
+
+func getExtension(client *Client) (*types.Extension, error) {
 	extensions := &types.Extension{}
 
-	extensionHREF := vcdClient.Client.VCDHREF
+	extensionHREF := client.VCDHREF
 	extensionHREF.Path += "/admin/extension/"
 
-	_, err := vcdClient.Client.ExecuteRequest(extensionHREF.String(), http.MethodGet,
+	_, err := client.ExecuteRequest(extensionHREF.String(), http.MethodGet,
 		"", "error retrieving extension: %s", nil, extensions)
 
 	return extensions, err
--- a/govcd/extension_test.go
+++ b/govcd/extension_test.go
@@ -6,10 +6,10 @@ package govcd
 
 import (
 	"fmt"
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
 	. "gopkg.in/check.v1"
 	"net/url"
-
-	"github.com/vmware/go-vcloud-director/v2/types/v56"
+	"time"
 )
 
 // Retrieves an external network and checks that its contents are filled as expected
@@ -60,11 +60,12 @@ func (vcd *TestVCD) Test_CreateExternalNetwork(check *C) {
 		Configuration: &types.NetworkConfiguration{
 			Xmlns: types.XMLNamespaceVCloud,
 			IPScopes: &types.IPScopes{
-				IPScope: types.IPScope{
-					Gateway: "192.168.201.1",
-					Netmask: "255.255.255.0",
-					DNS1:    "192.168.202.253",
-					DNS2:    "192.168.202.254",
+				IPScope: []*types.IPScope{&types.IPScope{
+					Gateway:   "192.168.201.1",
+					Netmask:   "255.255.255.0",
+					DNS1:      "192.168.202.253",
+					DNS2:      "192.168.202.254",
+					DNSSuffix: "some.net",
 					IPRanges: &types.IPRanges{
 						IPRange: []*types.IPRange{
 							&types.IPRange{
@@ -74,7 +75,7 @@ func (vcd *TestVCD) Test_CreateExternalNetwork(check *C) {
 						},
 					},
 				},
-			},
+				}},
 			FenceMode: "isolated",
 		},
 		VimPortGroupRefs: &types.VimObjectRefs{
@@ -91,7 +92,6 @@ func (vcd *TestVCD) Test_CreateExternalNetwork(check *C) {
 	}
 	task, err := CreateExternalNetwork(vcd.client, externalNetwork)
 	check.Assert(err, IsNil)
-	AddToCleanupList(externalNetwork.Name, "externalNetwork", "", "Test_CreateExternalNetwork")
 	check.Assert(task, Not(Equals), Task{})
 
 	err = task.WaitTaskCompletion()
@@ -103,15 +103,35 @@ func (vcd *TestVCD) Test_CreateExternalNetwork(check *C) {
 	check.Assert(newExternalNetwork.ExternalNetwork.Name, Equals, TestCreateExternalNetwork)
 
 	ipScope := newExternalNetwork.ExternalNetwork.Configuration.IPScopes.IPScope
-	check.Assert(ipScope.Gateway, Equals, "192.168.201.1")
-	check.Assert(ipScope.Netmask, Equals, "255.255.255.0")
-	check.Assert(ipScope.DNS1, Equals, "192.168.202.253")
-	check.Assert(ipScope.DNS2, Equals, "192.168.202.254")
+	check.Assert(ipScope[0].Gateway, Equals, "192.168.201.1")
+	check.Assert(ipScope[0].Netmask, Equals, "255.255.255.0")
+	check.Assert(ipScope[0].DNS1, Equals, "192.168.202.253")
+	check.Assert(ipScope[0].DNS2, Equals, "192.168.202.254")
+	check.Assert(ipScope[0].DNSSuffix, Equals, "some.net")
 
-	check.Assert(len(ipScope.IPRanges.IPRange), Equals, 1)
-	ipRange := ipScope.IPRanges.IPRange[0]
+	check.Assert(len(ipScope[0].IPRanges.IPRange), Equals, 1)
+	ipRange := ipScope[0].IPRanges.IPRange[0]
 	check.Assert(ipRange.StartAddress, Equals, "192.168.201.3")
 	check.Assert(ipRange.EndAddress, Equals, "192.168.201.250")
 
 	check.Assert(newExternalNetwork.ExternalNetwork.Configuration.FenceMode, Equals, "isolated")
+
+	// Needs to be deleted as port group used by other tests
+	// Workaround to refresh until task is fully completed - as task wait isn't enough
+	// Task still exists and creates NETWORK_DELETE error, so we wait until disappears
+	for i := 0; i < 30; i++ {
+		err = newExternalNetwork.Refresh()
+		check.Assert(err, IsNil)
+		if len(newExternalNetwork.ExternalNetwork.Tasks.Task) == 0 {
+			break
+		} else {
+			time.Sleep(1 * time.Second)
+		}
+	}
+
+	err = newExternalNetwork.DeleteWait()
+	if err != nil {
+		AddToCleanupList(externalNetwork.Name, "externalNetwork", "", "Test_CreateExternalNetwork")
+	}
+	check.Assert(err, IsNil)
 }
--- a/govcd/externalnetwork.go
+++ b/govcd/externalnetwork.go
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2019 VMware, Inc.  All rights reserved.  Licensed under the Apache v2 License.
+ */
+
+package govcd
+
+import (
+	"errors"
+	"fmt"
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+	"github.com/vmware/go-vcloud-director/v2/util"
+	"net/http"
+)
+
+type ExternalNetwork struct {
+	ExternalNetwork *types.ExternalNetwork
+	client          *Client
+}
+
+func NewExternalNetwork(cli *Client) *ExternalNetwork {
+	return &ExternalNetwork{
+		ExternalNetwork: new(types.ExternalNetwork),
+		client:          cli,
+	}
+}
+
+func getExternalNetworkHref(client *Client) (string, error) {
+	extensions, err := getExtension(client)
+	if err != nil {
+		return "", err
+	}
+
+	for _, extensionLink := range extensions.Link {
+		if extensionLink.Type == "application/vnd.vmware.admin.vmwExternalNetworkReferences+xml" {
+			return extensionLink.HREF, nil
+		}
+	}
+
+	return "", errors.New("external network link isn't found")
+}
+
+func (externalNetwork ExternalNetwork) GetByName(networkName string) error {
+	extNetworkHREF, err := getExternalNetworkHref(externalNetwork.client)
+	if err != nil {
+		return err
+	}
+
+	extNetworkRefs := &types.ExternalNetworkReferences{}
+	_, err = externalNetwork.client.ExecuteRequest(extNetworkHREF, http.MethodGet,
+		types.MimeNetworkConnectionSection, "error retrieving external networks: %s", nil, extNetworkRefs)
+	if err != nil {
+		return err
+	}
+
+	for _, netRef := range extNetworkRefs.ExternalNetworkReference {
+		if netRef.Name == networkName {
+			externalNetwork.ExternalNetwork.HREF = netRef.HREF
+			return externalNetwork.Refresh()
+		}
+	}
+
+	return fmt.Errorf("external network %s not found", networkName)
+}
+
+func (externalNetwork ExternalNetwork) Refresh() error {
+
+	_, err := externalNetwork.client.ExecuteRequest(externalNetwork.ExternalNetwork.HREF, http.MethodGet,
+		"", "error refreshing external network: %s", nil, externalNetwork.ExternalNetwork)
+
+	return err
+}
+
+func validateExternalNetwork(externalNetwork *types.ExternalNetwork) error {
+	if externalNetwork.Name == "" {
+		return errors.New("external Network missing required field: Name")
+	}
+	if externalNetwork.Xmlns == "" {
+		return errors.New("external Network missing required field: Xmlns")
+	}
+	return nil
+}
+
+func (externalNetwork *ExternalNetwork) Delete() (Task, error) {
+	util.Logger.Printf("[TRACE] ExternalNetwork.Delete")
+
+	// Return the task
+	return externalNetwork.client.ExecuteTaskRequest(externalNetwork.ExternalNetwork.HREF, http.MethodDelete,
+		"", "error deleting external network: %s", nil)
+}
+
+func (externalNetwork *ExternalNetwork) DeleteWait() error {
+	task, err := externalNetwork.Delete()
+	if err != nil {
+		return err
+	}
+	err = task.WaitTaskCompletion()
+	if err != nil {
+		return fmt.Errorf("couldn't finish removing external network %#v", err)
+	}
+	return nil
+}
--- a/govcd/externalnetwork_test.go
+++ b/govcd/externalnetwork_test.go
@@ -6,31 +6,29 @@ package govcd
 
 import (
 	"fmt"
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
 	. "gopkg.in/check.v1"
 	"net/url"
-
-	"github.com/vmware/go-vcloud-director/v2/types/v56"
 )
 
 // Retrieves an external network and checks that its contents are filled as expected
-func (vcd *TestVCD) Test_GetExternalNetwork(check *C) {
+func (vcd *TestVCD) Test_ExternalNetworkGetByName(check *C) {
 	fmt.Printf("Running: %s\n", check.TestName())
-	if vcd.skipAdminTests {
-		check.Skip(fmt.Sprintf(TestRequiresSysAdminPrivileges, check.TestName()))
+
+	if vcd.config.VCD.ExternalNetwork == "" {
+		check.Skip("Test_GetByName: External network isn't configured. Test can't proceed")
 	}
-	networkName := vcd.config.VCD.ExternalNetwork
-	if networkName == "" {
-		check.Skip("No external network provided")
-	}
-	externalNetwork, err := GetExternalNetworkByName(vcd.client, networkName)
+
+	externalNetwork := NewExternalNetwork(&vcd.client.Client)
+	err := externalNetwork.GetByName(vcd.config.VCD.ExternalNetwork)
 	check.Assert(err, IsNil)
-	LogExternalNetwork(*externalNetwork)
-	check.Assert(externalNetwork.HREF, Not(Equals), "")
-	check.Assert(externalNetwork.Name, Equals, networkName)
-	check.Assert(externalNetwork.Type, Equals, types.MimeExtensionNetwork)
+
+	check.Assert(externalNetwork.ExternalNetwork.Name, Equals, vcd.config.VCD.ExternalNetwork)
 }
 
-func (vcd *TestVCD) Test_CreateExternalNetwork(check *C) {
+// Tests System function Delete by creating external network and
+// deleting it after.
+func (vcd *TestVCD) Test_ExternalNetworkDelete(check *C) {
 	if vcd.skipAdminTests {
 		check.Skip("Configuration org != 'System'")
 	}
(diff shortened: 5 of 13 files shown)


govcd/sample_govcd_test_config.json
```diff
@@ -33,6 +33,10 @@
 		"externalNetmask": "255.255.224.0",
 		"internalIp": "192.168.1.10",
 		"internalNetmask": "255.255.255.0",
+		"externalNetwork": "myexternalnet",
+		"externalNetworkPortGroup": "VMNetwork",
+    "externalNetworkPortGroupType": "NETWORK",
```
lvirbalas: This looks like an additional parameter compared to the GUI. Why do we need it?

Please also add a comment for the users like you did in the YAML sample below.

types/v56/constants.go
```diff
@@ -90,6 +90,10 @@ const (
 	MimeMetaData = "application/vnd.vmware.vcloud.metadata+xml"
 	// Mime for metadata value
 	MimeMetaDataValue = "application/vnd.vmware.vcloud.metadata.value+xml"
+	// Mime for a admin network
+	MimeExtensionNetwork = "application/vnd.vmware.admin.extension.network+xml"
+	// Mime for for a external network
```
lvirbalas: Typo/grammar, suggested fix:
`// Mime for an external network`

types/v56/types.go
```diff
@@ -192,7 +192,7 @@ type SubAllocation struct {
 // Description: Represents a list of IP scopes.
 // Since: 5.1
 type IPScopes struct {
-	IPScope IPScope `xml:"IpScope"` // IP scope.
+	IPScope []*IPScope `xml:"IpScope"` // IP scope.
```
lvirbalas: Please elaborate why this change was needed. It impacted existing code quite a bit with additional casts like these:
`IPScope: []*types.IPScope{&types.IPScope{`

govcd/extension_test.go
```diff
@@ -24,7 +26,112 @@ func (vcd *TestVCD) Test_GetExternalNetwork(check *C) {
...
+
+	check.Assert(newExternalNetwork.ExternalNetwork.Configuration.FenceMode, Equals, "isolated")
+
+	// Needs to be deleted as port group used by other tests
+	// Workaround to refresh until task is fully completed - as task wait isn't enough
+	// Task still exists and creates NETWORK_DELETE error, so we wait until disappears
+	for i := 0; i < 30; i++ {
+		err = newExternalNetwork.Refresh()
+		check.Assert(err, IsNil)
+		if len(newExternalNetwork.ExternalNetwork.Tasks.Task) == 0 {
```
lvirbalas: What kind of task are we waiting to finish here?

govcd/sample_govcd_test_config.json
vbauzysvmware: Cause we can use Networks or distributed switches when create external network. It's in GUI but not explicitly showed. You just choose any of them. And you can have network and dswitch with same name.

types/v56/constants.go
vbauzysvmware: fixed

types/v56/types.go
vbauzysvmware: Existing implementations allows only one ip scope - it wasn't correct. vCD allows to have many ip scopes.

govcd/extension_test.go
vbauzysvmware: Creation task. I have spent many hours tracking the issue - seems in home lab, create task returns it's finished, but in reality it isn't for second or so, so had to add additional wait for checking if task finished.

types/v56/types.go
```diff
@@ -474,8 +475,8 @@ type Task struct {
 // Since: 0.9
 type CapacityWithUsage struct {
 	Units     string `xml:"Units"`
-	Allocated int64  `xml:"Allocated,omitempty"`
-	Limit     int64  `xml:"Limit,omitempty"`
+	Allocated int64  `xml:"Allocated"`
```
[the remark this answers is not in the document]
vbauzysvmware: Megalord changed, I think it is required field. Can be verified..

govcd/sample_govcd_test_config.yaml
```diff
@@ -77,6 +77,15 @@ vcd:
     # An external Network name
     externalNetwork: myexternalnet
     #
+    # A port group name for creating an external network
```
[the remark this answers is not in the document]
vbauzysvmware: I already updated that

govcd/extension_test.go
```diff
@@ -24,7 +26,112 @@ func (vcd *TestVCD) Test_GetExternalNetwork(check *C) {
 	check.Assert(err, IsNil)
 	LogExternalNetwork(*externalNetwork)
 	check.Assert(externalNetwork.HREF, Not(Equals), "")
-	expectedType := "application/vnd.vmware.admin.extension.network+xml"
 	check.Assert(externalNetwork.Name, Equals, networkName)
-	check.Assert(externalNetwork.Type, Equals, expectedType)
+	check.Assert(externalNetwork.Type, Equals, types.MimeExtensionNetwork)
+}
+
+func (vcd *TestVCD) Test_CreateExternalNetwork(check *C) {
```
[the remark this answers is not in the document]
vbauzysvmware: Can you explain what this race condition is? - cause like from external library comes.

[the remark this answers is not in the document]
Didainius: It actually gives this error everytime (wonder if it's on my env only).
It wasn't  full output as the full is pretty big.
```
WARNING: DATA RACE                                                                                                                                                                [169/3097]
Read at 0x00c0000bed10 by goroutine 31:
  github.com/kr/pretty.(*printer).printValue()
      /usr/local/Cellar/go/1.12.4/libexec/src/reflect/value.go:952 +0xd5c
  github.com/kr/pretty.(*printer).printValue()
      /Users/dserplis/go/pkg/mod/github.com/kr/pretty@v0.1.0/formatter.go:182 +0x228c
  github.com/kr/pretty.(*printer).printValue()
      /Users/dserplis/go/pkg/mod/github.com/kr/pretty@v0.1.0/formatter.go:249 +0x1963
  github.com/kr/pretty.(*printer).printValue()
      /Users/dserplis/go/pkg/mod/github.com/kr/pretty@v0.1.0/formatter.go:182 +0x228c
  github.com/kr/pretty.(*printer).printValue()
      /Users/dserplis/go/pkg/mod/github.com/kr/pretty@v0.1.0/formatter.go:249 +0x1963
  github.com/kr/pretty.(*printer).printValue()
      /Users/dserplis/go/pkg/mod/github.com/kr/pretty@v0.1.0/formatter.go:228 +0x34f
  github.com/kr/pretty.(*printer).printValue()
      /Users/dserplis/go/pkg/mod/github.com/kr/pretty@v0.1.0/formatter.go:137 +0x14a4
  github.com/kr/pretty.(*printer).printValue()
      /Users/dserplis/go/pkg/mod/github.com/kr/pretty@v0.1.0/formatter.go:182 +0x228c
  github.com/kr/pretty.(*printer).printValue()
      /Users/dserplis/go/pkg/mod/github.com/kr/pretty@v0.1.0/formatter.go:249 +0x1963
  github.com/kr/pretty.(*printer).printValue()
      /Users/dserplis/go/pkg/mod/github.com/kr/pretty@v0.1.0/formatter.go:182 +0x228c
  github.com/kr/pretty.(*printer).printValue()
      /Users/dserplis/go/pkg/mod/github.com/kr/pretty@v0.1.0/formatter.go:182 +0x228c
  github.com/kr/pretty.(*printer).printValue()
      /Users/dserplis/go/pkg/mod/github.com/kr/pretty@v0.1.0/formatter.go:249 +0x1963
  github.com/kr/pretty.formatter.Format()
      /Users/dserplis/go/pkg/mod/github.com/kr/pretty@v0.1.0/formatter.go:57 +0x223
  github.com/kr/pretty.(*formatter).Format()
      <autogenerated>:1 +0xab
  fmt.(*pp).handleMethods()
      /usr/local/Cellar/go/1.12.4/libexec/src/fmt/print.go:583 +0x186
  fmt.(*pp).printArg()
      /usr/local/Cellar/go/1.12.4/libexec/src/fmt/print.go:699 +0x24f
  fmt.(*pp).doPrintf()
      /usr/local/Cellar/go/1.12.4/libexec/src/fmt/print.go:1016 +0x31c
  fmt.Sprintf()
      /usr/local/Cellar/go/1.12.4/libexec/src/fmt/print.go:214 +0x73                                                                                                              [132/3097]
  github.com/kr/pretty.(*sbuf).Printf()
      /Users/dserplis/go/pkg/mod/github.com/kr/pretty@v0.1.0/diff.go:12 +0x79
  github.com/kr/pretty.diffPrinter.diff()
      /Users/dserplis/go/pkg/mod/github.com/kr/pretty@v0.1.0/diff.go:76 +0x2dba
  github.com/kr/pretty.diffPrinter.diff()
      /Users/dserplis/go/pkg/mod/github.com/kr/pretty@v0.1.0/diff.go:170 +0x373c
  github.com/kr/pretty.Pdiff()
      /Users/dserplis/go/pkg/mod/github.com/kr/pretty@v0.1.0/diff.go:44 +0xaf
  gopkg.in/check%2ev1.formatUnequal()
      /Users/dserplis/go/pkg/mod/github.com/kr/pretty@v0.1.0/diff.go:19 +0x1aa
  gopkg.in/check%2ev1.(*equalsChecker).Check()
      /Users/dserplis/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20180628173108-788fd7840127/checkers.go:238 +0x159
  gopkg.in/check%2ev1.(*notChecker).Check()
      /Users/dserplis/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20180628173108-788fd7840127/checkers.go:94 +0xa2
  gopkg.in/check%2ev1.(*C).internalCheck()
      /Users/dserplis/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20180628173108-788fd7840127/helpers.go:214 +0x29e
  gopkg.in/check%2ev1.(*C).Assert()
      /Users/dserplis/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20180628173108-788fd7840127/helpers.go:173 +0xa6
  github.com/vmware/go-vcloud-director/v2/govcd.(*TestVCD).Test_CreateExternalNetwork()
      /Users/dserplis/Documents/vmware/go-vcloud-director/govcd/extension_test.go:96 +0x10a2
  runtime.call32()
      /usr/local/Cellar/go/1.12.4/libexec/src/runtime/asm_amd64.s:519 +0x3a
  reflect.Value.Call()
      /usr/local/Cellar/go/1.12.4/libexec/src/reflect/value.go:308 +0xc0
  gopkg.in/check%2ev1.(*suiteRunner).forkTest.func1()
      /Users/dserplis/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20180628173108-788fd7840127/check.go:781 +0x978
  gopkg.in/check%2ev1.(*suiteRunner).forkCall.func1()
      /Users/dserplis/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20180628173108-788fd7840127/check.go:675 +0xb7

Previous write at 0x00c0000bed10 by goroutine 29:
  bufio.(*Reader).Peek()
      /usr/local/Cellar/go/1.12.4/libexec/src/bufio/bufio.go:135 +0x79
  net/http.(*persistConn).readLoop()
      /usr/local/Cellar/go/1.12.4/libexec/src/net/http/transport.go:1677 +0x27b

Goroutine 31 (running) created at:
  gopkg.in/check%2ev1.(*suiteRunner).forkCall()
      /Users/dserplis/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20180628173108-788fd7840127/check.go:672 +0x44e
  gopkg.in/check%2ev1.(*suiteRunner).forkTest()
      /Users/dserplis/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20180628173108-788fd7840127/check.go:763 +0x1b9
  gopkg.in/check%2ev1.(*suiteRunner).run()
      /Users/dserplis/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20180628173108-788fd7840127/check.go:818 +0x226
  gopkg.in/check%2ev1.Run()
      /Users/dserplis/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20180628173108-788fd7840127/run.go:92 +0x5a
  gopkg.in/check%2ev1.RunAll()
      /Users/dserplis/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20180628173108-788fd7840127/run.go:84 +0x129
  gopkg.in/check%2ev1.TestingT()
      /Users/dserplis/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20180628173108-788fd7840127/run.go:72 +0x6a4
  github.com/vmware/go-vcloud-director/v2/govcd.Test()
      /Users/dserplis/Documents/vmware/go-vcloud-director/govcd/api_vcd_test.go:233 +0x38
  testing.tRunner()
      /usr/local/Cellar/go/1.12.4/libexec/src/testing/testing.go:865 +0x163

Goroutine 29 (running) created at:
  net/http.(*Transport).dialConn()
      /usr/local/Cellar/go/1.12.4/libexec/src/net/http/transport.go:1357 +0xc32
  net/http.(*Transport).getConn.func4()
      /usr/local/Cellar/go/1.12.4/libexec/src/net/http/transport.go:1015 +0xd0
==================
```
Line 95 triggers it, but don't understand why as the task should have been processed by that time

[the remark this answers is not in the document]
vbauzysvmware: It gives for me too, if I recall when I checked. But don't see option to solve it cause `check.Assert` isn't our thing ^^

govcd/extension_test.go
```diff
@@ -24,7 +26,112 @@ func (vcd *TestVCD) Test_GetExternalNetwork(check *C) {
...
+					},
+					MoRef:         portGroups[0].MoRef,
+					VimObjectType: vcd.config.VCD.ExternalNetworkPortGroupType,
+				},
+			},
+		},
+	}
+	task, err := CreateExternalNetwork(vcd.client, externalNetwork)
+	check.Assert(err, IsNil)
+	check.Assert(task, Not(Equals), Task{})
```
Didainius: To avoid the triggering data race discussed above (because http.Client does some cleanups inside) I'd suggest to change this line to such:
```suggestion
	check.Assert(task.Task, Not(Equals), types.Task{})
```

vbauzysvmware: Tx. Added

govcd/extension_test.go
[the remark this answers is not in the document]
Didainius: Just for future reference. We think http.Client (which is embedded into task struct) does it's own cleanups in a goroutine (Goroutine 29 in the example above) and triggers data race because assertion in line 95 is accessing it.

govcd/sample_govcd_test_config.yaml
```diff
@@ -77,6 +77,15 @@ vcd:
     # An external Network name
     externalNetwork: myexternalnet
```
dataclouder: We need to change the description of this field, saying that it is an already existing external network.

govcd/sample_govcd_test_config.yaml
```diff
@@ -77,6 +77,15 @@ vcd:
...
     #
+    # A port group name for creating an external network
+    externalNetworkPortGroup: ForTestingPG
+    #
+    # A port group type for creating an external network. Can be DV_PORTGROUP or NETWORK
```
dataclouder: We should specify that these fields are used for a new external network to be used in tests.

govcd/externalnetwork_test.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	externalNetwork := NewExternalNetwork(&vcd.client.Client)
+	err := externalNetwork.GetByName(vcd.config.VCD.ExternalNetwork)
+	check.Assert(err, IsNil)
+
+	check.Assert(externalNetwork.ExternalNetwork.Name, Equals, vcd.config.VCD.ExternalNetwork)
+}
+
+// Tests System function Delete by creating external network and
```
dataclouder: A few problems with this test:

1. We need a check that the needed elements were set in the configuration, and abort if they weren't.
2. If the Port group type is wrong, the test fails with a message saying that the port group was not found, instead of saying that the type doesn't match. Is there a way of giving a better message?

NOTE ON TESTING: if we have an already configured Cassini vCD, we can't test the NETWORK type, as it's already taken. To test the creation of a network (using `VM Network`, for example), we need to define at least an organization, and run the test with environment variable `GOVCD_SKIP_VAPP_CREATION=1`.

govcd/sample_govcd_test_config.yaml
vbauzysvmware: Fixed

govcd/sample_govcd_test_config.yaml
vbauzysvmware: Fixed

govcd/externalnetwork_test.go
vbauzysvmware: 1. I added checks in tests to abort when configuration is missing.
2. This is query vDC by name or type. Can be anything between so no good option to tell what is wrong automatically.

govcd/externalnetwork_test.go
```diff
@@ -0,0 +1,138 @@
+/*
```
dataclouder: Consider adding the necessary tags here
```
// +build extnetwork network functional ALL

```
This way, after the PR is merged, it will be ready to run.
The drawback is that this test won't run with `make test` until the PR is merged. 

Having the `network` tag means that the test will run with the rest of the networking. The `extnetwork` tag will also allow running the test in isolation, if we want to try the creation of an external network on a non-configured vCD.

govcd/extension.go
```diff
@@ -5,15 +5,16 @@
 package govcd
 
 import (
-	"errors"
+	"fmt"
 	"github.com/vmware/go-vcloud-director/v2/types/v56"
 	"net/http"
 )
 
+// DEPRECATED please use GetExternalNetwork function instead
```
dataclouder: Why is this function deprecated?

govcd/extension_test.go
```diff
@@ -24,7 +26,129 @@ func (vcd *TestVCD) Test_GetExternalNetwork(check *C) {
 	check.Assert(err, IsNil)
 	LogExternalNetwork(*externalNetwork)
 	check.Assert(externalNetwork.HREF, Not(Equals), "")
-	expectedType := "application/vnd.vmware.admin.extension.network+xml"
 	check.Assert(externalNetwork.Name, Equals, networkName)
-	check.Assert(externalNetwork.Type, Equals, expectedType)
+	check.Assert(externalNetwork.Type, Equals, types.MimeExtensionNetwork)
+}
+
+func (vcd *TestVCD) Test_CreateExternalNetwork(check *C) {
```
dataclouder: For test organization (using tags) this function should be in `externalnetwork_test.go`

govcd/externalnetwork.go
```diff
@@ -0,0 +1,114 @@
...
+		return "", err
+	}
+
+	for _, extensionLink := range extensions.Link {
+		if extensionLink.Type == "application/vnd.vmware.admin.vmwExternalNetworkReferences+xml" {
+			return extensionLink.HREF, nil
+		}
+	}
+
+	return "", errors.New("external network link isn't found")
```
dataclouder: "isn't found" -> "wasn't found"

govcd/externalnetwork.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func (externalNetwork ExternalNetwork) GetByName(networkName string) error {
```
dataclouder: I am puzzled by this method. We have an entity that looks for an instance of the same type, and then returns an error.
Semantically, I can't figure out how this function fits with the rest of the library. I would expect to see a function in `system.go`, using a `VCDClient` as an accessor, and returning an external network.

What happens exactly here, instead? When the function is successful, what do I get? The original external network (the one used for finding by name) becomes the found one?

govcd/system.go
```diff
@@ -108,3 +109,39 @@ func getOrgHREF(vcdClient *VCDClient, orgName string) (string, error) {
 	}
 	return "", fmt.Errorf("couldn't find org with name: %s. Please check Org name as it is case sensitive", orgName)
 }
+
+// Find a list of Virtual Centers matching the filter parameter.
```
dataclouder: Please add a list of parameters that can be used and their format

govcd/system.go
```diff
@@ -108,3 +109,39 @@ func getOrgHREF(vcdClient *VCDClient, orgName string) (string, error) {
...
+		"filter": filter,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return results.Results.VirtualCenterRecord, nil
+}
+
+// Find a Network port groups by name
```
dataclouder: "port groups" -> "port group"

govcd/system.go
```diff
@@ -108,3 +109,39 @@ func getOrgHREF(vcdClient *VCDClient, orgName string) (string, error) {
...
+func QueryNetworkPortGroups(vdcCli *VCDClient, name string) ([]*types.PortGroupRecordType, error) {
+	return QueryPortGroups(vdcCli, fmt.Sprintf("(name==%s;portgroupType==%s)", url.QueryEscape(name), "NETWORK"))
+}
+
+// Find a Distributed port groups by name
```
dataclouder: "port groups" -> "port group"

govcd/externalnetwork.go
vbauzysvmware: Fixed

govcd/externalnetwork_test.go
vbauzysvmware: I merged with latest PR and added tags.

types/v56/types.go
```diff
@@ -2171,6 +2194,48 @@ type ExternalNetworkReference struct {
 	Name string `xml:"name,attr,omitempty"`
 }
 
+// Type: VimObjectRefType
+// Namespace: http://www.vmware.com/vcloud/extension/v1.5
+// https://vdc-repo.vmware.com/vmwb-repository/dcr-public/7a028e78-bd37-4a6a-8298-9c26c7eeb9aa/09142237-dd46-4dee-8326-e07212fb63a8/doc/doc/types/VimObjectRefsType.html
+// Description: Represents the moref and the type of a vSphere object.
```
dataclouder: "moref" ?

govcd/extension.go
vbauzysvmware: Cause it returns `types.ExternalNetworkReference` vCD API type, not `ExternalNetwork` govcd API type

govcd/extension_test.go
vbauzysvmware: Well from that perspective you are right, but from functionality point it is in right place as we match unitest file name against functions which resides there.

types/v56/types.go
vbauzysvmware: Yep - `moref` - this is vCD API - that's how it's called

govcd/extension_test.go
dataclouder: In that case `extension_test.go` needs the tag `extnetwork`.

govcd/externalnetwork.go
vbauzysvmware: This was Megalord idea, I understand it after using it, but I agree this isn't very clean approach. I will look for idea to improve it.

govcd/system.go
vbauzysvmware: Elaborated with additional information

types/v56/types.go
dataclouder: Can we say "Managed Object Reference (MoRef)" for clarity?

govcd/system.go
vbauzysvmware: fixed

govcd/system.go
vbauzysvmware: fixed

types/v56/types.go
vbauzysvmware: Done

govcd/externalnetwork.go
vbauzysvmware: This is how other may think about current structure of API, we have `externalNetwork` which consist of `client` and `externalNetwork` which is types struct. What Megalord did was - create `externalNetwork` with client  and fetch `types.externalNetwork` then with 'getByName'

dataclouder: I like a `Get` function to return something other than an error. 
A function that returns an error is one that runs a procedure, and informs on the outcome.
I am still unsure I understand what this function does. What happens to the external network being found? and what changes in the calling object?

If it searches for something, it should return what it finds, and even then, I don't like to use an object to find an entity of the same type, and potentially itself. 
If the function changes something in the calling object, the name should say what this operation was.

vbauzysvmware: Yep, I agree. I changed to be more understandable

govcd/api_vcd_test.go
```diff
@@ -90,13 +92,16 @@ type TestConfig struct {
...
-		Disk            struct {
+		ExternalIp                   string `yaml:"externalIp,omitempty"`
+		ExternalNetmask              string `yaml:"externalNetmask,omitempty"`
+		InternalIp                   string `yaml:"internalIp,omitempty"`
+		InternalNetmask              string `yaml:"internalNetmask,omitempty"`
+		EdgeGateway                  string `yaml:"edgeGateway,omitempty"`
+		ExternalNetwork              string `yaml:"externalNetwork,omitempty"`
+		ExternalNetworkPortGroup     string `yaml:"externalNetworkPortGroup,omitempty"`
+		ExternalNetworkPortGroupType string `yaml:"externalNetworkPortGroupType,omitempty"`
+		VimServer                    string `yaml:"vimServer,omitempty"`
```
dataclouder: Can we call this field `Vcenter`, as it is in terraform-provider-vcd?

types/v56/types.go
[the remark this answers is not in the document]
vbauzysvmware: Updated according API

govcd/api_vcd_test.go
vbauzysvmware: Yes, I can. If everyone agrees on that. My idea was to match more vCD api in govcd, in terraform vcd GUI.

lvirbalas: It's better if we match vCD API naming in go-vcloud-director - govcd should provide a relatively thin and easily translatable layer over the vCD API.

govcd/system.go
```diff
@@ -108,3 +109,42 @@ func getOrgHREF(vcdClient *VCDClient, orgName string) (string, error) {
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return results.Results.VirtualCenterRecord, nil
+}
+
+// Find a Network port group by name
+func QueryNetworkPortGroup(vdcCli *VCDClient, name string) ([]*types.PortGroupRecordType, error) {
```
dataclouder: `vdcCli` should be `vcdCli`

govcd/sample_govcd_test_config.json
lvirbalas: Thanks for adding comments!

types/v56/constants.go
lvirbalas: Almost, "a" -> "an"

types/v56/types.go
lvirbalas: Still looks a disturbing (a lot of existing code) change, but maybe it's only me. @Didainius and @dataclouder what do you think?

govcd/extension.go
```diff
@@ -33,28 +34,70 @@ func GetExternalNetworkByName(vcdClient *VCDClient, networkName string) (*types.
 	return &types.ExternalNetworkReference{}, nil
 }
 
-func getExternalNetworkHref(vcdClient *VCDClient) (string, error) {
-	extensions, err := getExtension(vcdClient)
+// If user specifies a valid external network name, then this returns a
```
Didainius: Good practice to start function comment with it's name to be go doc friendly. Otherwise it breaks grammar as in the example below.

image

types/v56/constants.go
vbauzysvmware: done

types/v56/types.go
vbauzysvmware: Well this only fix a bug in govcd. All implementation still if were using one scope [0] still using.

govcd/externalnetwork_test.go
```diff
@@ -0,0 +1,143 @@
...
+// Tests System function Delete by creating external network and
+// deleting it after.
+func (vcd *TestVCD) Test_ExternalNetworkDelete(check *C) {
+	fmt.Printf("Running: %s\n", check.TestName())
+	if vcd.skipAdminTests {
+		check.Skip(fmt.Sprintf(TestRequiresSysAdminPrivileges, check.TestName()))
+	}
+
+	if vcd.config.VCD.ExternalNetwork == "" {
+		check.Skip("Test_GetByName: External network isn't configured. Test can't proceed")
```
Didainius: Unless this is some special construct I don't know:
```suggestion
		check.Skip("Test_ExternalNetworkDelete: External network isn't configured. Test can't proceed")
```

govcd/extension_test.go
lvirbalas: @vbauzysvmware what do you think with respect to @Didainius comment above? I'm concerned these lines in current test may be masking a potentially bigger issue.

govcd/externalnetwork_test.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	if vcd.config.VCD.VimServer == "" {
+		check.Skip("Test_GetByName: Vim server isn't configured. Test can't proceed")
```
Didainius: Same as above.

govcd/externalnetwork_test.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	if vcd.config.VCD.ExternalNetworkPortGroup == "" {
+		check.Skip("Test_GetByName: Port group isn't configured. Test can't proceed")
```
Didainius: Same as above.

govcd/externalnetwork_test.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	if vcd.config.VCD.ExternalNetworkPortGroupType == "" {
+		check.Skip("Test_GetByName: Port group type isn't configured. Test can't proceed")
```
Didainius: Same as above.

govcd/extension_test.go
vbauzysvmware: Race condition was due test and http handling go procedures. It is no implicate this case

govcd/system.go
vbauzysvmware: fixed

govcd/extension_test.go
```diff
@@ -26,11 +28,132 @@ func (vcd *TestVCD) Test_GetExternalNetwork(check *C) {
...
+	fmt.Printf("Running: %s\n", check.TestName())
+	if vcd.skipAdminTests {
+		check.Skip(fmt.Sprintf(TestRequiresSysAdminPrivileges, check.TestName()))
+	}
+
+	if vcd.config.VCD.ExternalNetwork == "" {
+		check.Skip("Test_GetByName: External network isn't configured. Test can't proceed")
```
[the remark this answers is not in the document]
vbauzysvmware: fixed

govcd/extension_test.go
```diff
@@ -26,11 +28,132 @@ func (vcd *TestVCD) Test_GetExternalNetwork(check *C) {
...
+	if vcd.config.VCD.VimServer == "" {
+		check.Skip("Test_GetByName: Vim server isn't configured. Test can't proceed")
+	}
+
+	if vcd.config.VCD.ExternalNetworkPortGroup == "" {
+		check.Skip("Test_GetByName: Port group isn't configured. Test can't proceed")
+	}
+
+	if vcd.config.VCD.ExternalNetworkPortGroupType == "" {
+		check.Skip("Test_GetByName: Port group type isn't configured. Test can't proceed")
```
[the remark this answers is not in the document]
vbauzysvmware: fixed

govcd/externalnetwork_test.go
vbauzysvmware: fixed

govcd/externalnetwork_test.go
vbauzysvmware: fixed

govcd/externalnetwork_test.go
vbauzysvmware: fixed

govcd/externalnetwork_test.go
vbauzysvmware: fixed

govcd/extension_test.go
```diff
@@ -26,11 +28,132 @@ func (vcd *TestVCD) Test_GetExternalNetwork(check *C) {
...
+	err = newExternalNetwork.DeleteWait()
+	if err != nil {
+		AddToCleanupList(externalNetwork.Name, "externalNetwork", "", "Test_CreateExternalNetwork")
+	}
+	check.Assert(err, IsNil)
 }
 
 func init() {
-	testingTags["extension"] = "extension_test.go"
+	testingTags["extnetwork"] = "extension_test.go"
```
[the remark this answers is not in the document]
vbauzysvmware: we have extension and external network. Or I don't get idea

govcd/extension.go
vbauzysvmware: fixed

govcd/extension_test.go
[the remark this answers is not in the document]
vbauzysvmware: Idea is that extnetwork tag runs two tests extension and externalnetwork, cause functionality is splitted

govcd/extension_test.go
```diff
@@ -26,11 +28,132 @@ func (vcd *TestVCD) Test_GetExternalNetwork(check *C) {
...
+	if vcd.config.VCD.ExternalNetwork == "" {
+		check.Skip("Test_GetByName: External network isn't configured. Test can't proceed")
+	}
+
+	if vcd.config.VCD.VimServer == "" {
+		check.Skip("Test_GetByName: Vim server isn't configured. Test can't proceed")
+	}
+
+	if vcd.config.VCD.ExternalNetworkPortGroup == "" {
+		check.Skip("Test_GetByName: Port group isn't configured. Test can't proceed")
```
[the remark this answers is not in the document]
vbauzysvmware: fixed

TESTING.md
```diff
@@ -72,7 +72,7 @@ $ go test -v .
 
            * catalog:    Runs catalog related tests (also catalog_item, media)
            * disk:       Runs disk related tests
-           * extension:  Runs extension related tests
+           * extnetwork: Runs external network related tests
```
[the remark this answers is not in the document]
vbauzysvmware: extension is left due API functionality and in future will have more functionality cause anything advanced are in /extension part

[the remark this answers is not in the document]
Didainius: Still don't understand why `CreateExternalNetwork` is in `extension.go` then. To me it sounds as it is part of `extnetwork.go`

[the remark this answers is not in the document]
vbauzysvmware: In general createExternalNetwork may be in system according our API tree, but cause it uses /extesion API it was moved to extensions.go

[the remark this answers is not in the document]
vbauzysvmware: If you look createCatalog is not part of catalog.go is part org org.go cause his dependent on Org.

govcd/extension_test.go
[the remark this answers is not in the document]
dataclouder: I second the request. I am partly responsible for naming this file, and it was a short-sighted decision.

`extension_test.go` is only running tests for external network and everything from this file should be moved to `externalnetwork_test.go`.
When we add new functionalities through extensions, it would be better to name the files according to the features we implement rather than the accessor.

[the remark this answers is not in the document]
dataclouder: Moreover, we have two initialization statements for `testingTags["extnetwork"]`. I missed this in my previous review.
When the same feature spans multiple files, we should initialize the tag canary variable only in one of the files.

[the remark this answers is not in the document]
vbauzysvmware: ok. I will move the test, but what about implmenatation `exension.go` - I recommend to move it to system.go and yet, due deprecated method we still have to leave the file for time being.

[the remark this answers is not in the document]
vbauzysvmware: Test moved

TESTING.md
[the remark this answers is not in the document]
Didainius: Ok. The `extension.go` seemed very confusing to me as it has nothing to do with extension.

types/v56/types.go
```diff
@@ -475,10 +476,10 @@ type Task struct {
 type CapacityWithUsage struct {
 	Units     string `xml:"Units"`
 	Allocated int64  `xml:"Allocated,omitempty"`
-	Limit     int64  `xml:"Limit,omitempty"`
+	Limit     int64  `xml:"Limit"`
 	Reserved  int64  `xml:"Reserved,omitempty"`
 	Used      int64  `xml:"Used,omitempty"`
-	Overhead  int64  `xml:"Overhead,omitempty"`
+	Overhead  int64  `xml:"Overhead,omitempty"` // not available anymore from API v30.0
```
Didainius: More of a question - we're using v27.0, how did you spot v30.0? Was it only in docs or did you hit it somehow?

types/v56/types.go
lvirbalas: Got it, we were not following XML serialisation of vCD API here and now we do. Further issues that became visible after this change will be tracked here:
https://github.com/terraform-providers/terraform-provider-vcd/issues/235

types/v56/types.go
vbauzysvmware: Accidentally spotted checking against API:) from time to time I shuffle between versions

PR #188, opened 2019-05-06
govcd multiple nic support

Didainius:
Ref #162.
Opening new PR to continue @dn1s work for multinic support. This is owned by me for quicker turnarounds while reviewing.

Tests are green on vCD 9.5

**Note** `govcd_test_config.yaml` structure changed and now requires two networks defined. Examples are updated.


govcd/vm.go
```diff
@@ -16,7 +17,7 @@ import (
 
 type VM struct {
 	VM     *types.VM
-	client *Client
+	client *Client /**/
```
dn1s: Empty comment

govcd/vapp.go
```diff
@@ -124,10 +124,8 @@ func (vapp *VApp) AddVM(orgVdcNetworks []*types.OrgVDCNetwork, vappNetworkName s
 			},
 			InstantiationParams: &types.InstantiationParams{
 				NetworkConnectionSection: &types.NetworkConnectionSection{
-					Type:                          vappTemplate.VAppTemplate.Children.VM[0].NetworkConnectionSection.Type,
-					HREF:                          vappTemplate.VAppTemplate.Children.VM[0].NetworkConnectionSection.HREF,
 					Info:                          "Network config for sourced item",
-					PrimaryNetworkConnectionIndex: vappTemplate.VAppTemplate.Children.VM[0].NetworkConnectionSection.PrimaryNetworkConnectionIndex,
+					PrimaryNetworkConnectionIndex: 0,
```
lvirbalas: Curious, why are we able to specify zero now, but previously could not?

govcd/vapp_test.go
```diff
@@ -28,6 +28,61 @@ func (vcd *TestVCD) TestGetParentVDC(check *C) {
...
+//	}
+//
+//	err = vapp.BlockWhileStatus("UNRESOLVED", vapp.client.MaxRetryTimeout)
+//	if err != nil {
+//		return VApp{}, fmt.Errorf("error waitinf for created test vApp to have working state: %s", err)
+//	}
+//
+//	return vapp, err
+//}
+
```
lvirbalas: Why are we keeping this huge commented block?

govcd/vm.go
```diff
@@ -164,53 +205,80 @@ func (vm *VM) ChangeCPUCountWithCore(virtualCpuCount int, coresPerSocket *int) (
...
-				} else if ip == "none" || network["ip"].(string) == "none" {
+				case ipIsSet && ipFieldString == "none": // Legacy ip="none" mode
 					ipAllocationMode = types.IPAllocationModeNone
-				} else if ip != "" {
-					ipAllocationMode = types.IPAllocationModeManual
-					// TODO: Check a valid IP has been given
-					ipAddress = ip
-				} else if network["ip"].(string) != "" {
+
+				// Legacy ip="valid_ip" mode (currently it is hit by ip_allocation_mode=MANUAL as well)
```
lvirbalas: I suggest chaning word "legacy" to "deprecated" in all comments to be more precise.

govcd/vm.go
```diff
@@ -164,53 +205,80 @@ func (vm *VM) ChangeCPUCountWithCore(virtualCpuCount int, coresPerSocket *int) (
...
+
+func (vm *VM) ChangeNetworkConfig(networks []map[string]interface{}) (Task, error) {
+	err := vm.Refresh()
+	if err != nil {
+		return Task{}, fmt.Errorf("error refreshing VM before running customization: %v", err)
+	}
+
+	// The API returns unordered list of NICs. This means that networkSection.NetworkConnection[0] will not
+	// necessarily be NIC 0 if there is more than one NIC available.
+	networkSection, err := vm.GetNetworkConnectionSection()
```
lvirbalas: This is a useful comment, but why do we have it here is not very clear. After reading such comment I am instinctively hoping to see some reordering magic lines :) If they are somewhere here, then it's probably worth elaborating in this comment as to how it is done.

govcd/vm.go
```diff
@@ -164,53 +205,80 @@ func (vm *VM) ChangeCPUCountWithCore(virtualCpuCount int, coresPerSocket *int) (
...
-				} else if ip == "" {
+					ipAddress = ipFieldString
+				case ipIsSet && ipField != "": // Legacy ip="something_invalid" we default to DHCP. This is odd but backwards compatible.
 					ipAllocationMode = types.IPAllocationModeDHCP
+					// TODO 3.0 remove until here when legacy `ip` and `network_name` attributes are removed
+
+				case ipIsSet && net.ParseIP(ipFieldString) != nil && (network["ip_allocation_mode"].(string) == types.IPAllocationModeManual):
+					ipAllocationMode = types.IPAllocationModeManual
+					ipAddress = ipFieldString
+				default: // New functionality. IP was not set and we're defaulting to provided ip_allocation_mode (only manual requires the IP)
```
lvirbalas: What about "New **networks** functionality" to make sure it's clear we're referring to multi networks here.

govcd/vapp.go
Didainius: If I remember @dn1s changed this so that we do not hit empty index when a vApp template does not have network cards assigned.

govcd/vm.go
Didainius: Changed

govcd/vm.go
Didainius: Absolutely right. Over the time this sorting got to other place.

govcd/vm.go
Didainius: Addressed

govcd/vapp_test.go
Didainius: My bad. Left it after merging with latest master.

govcd/api_vcd_test.go
```diff
@@ -293,7 +294,7 @@ func (vcd *TestVCD) SetUpSuite(check *C) {
 		skipVappCreation = true
 	}
 	// creates a new VApp for vapp tests
-	if !skipVappCreation && config.VCD.Network != "" && config.VCD.StorageProfile.SP1 != "" &&
+	if !skipVappCreation && config.VCD.Networks[0] != "" && config.VCD.StorageProfile.SP1 != "" &&
```
dataclouder: Before using `config.VCD.Networks[0]` we need to check that `config.VCD.Networks` has been initialised

dataclouder: Also, this is a good place to check whether the network names are different (i.e. user did not provide the same name twice)

govcd/sample_govcd_test_config.json
```diff
@@ -23,7 +23,7 @@
 			"description": "my cat for loading",
 			"catalogItemDescription": "my item to create vapps"
 		},
-		"network": "mynet",
+		"networks": ["mynet", "mynet2"],
```
dataclouder: What's the required number of networks? I see that we have some tests that check for length of this field as being `>= 2`.
* Is two networks sufficient? 
* Do they need to be of different type?

govcd/vapp.go
lvirbalas: And is it safe to assume that zero is the primary network connection? I'm asking because I recall elsewhere in this PR there was a comment that the first connection is not guaranteed to be the primary.

govcd/vapp.go
```diff
@@ -719,7 +717,48 @@ func (vapp *VApp) GetNetworkConfig() (*types.NetworkConfigSection, error) {
...
+				FenceMode: "bridged",
+			},
+		},
+	)
+
+	apiEndpoint, _ := url.ParseRequestURI(vapp.VApp.HREF)
+	apiEndpoint.Path += "/networkConfigSection/"
+
+	return vapp.client.ExecuteTaskRequest(apiEndpoint.String(), http.MethodPut,
+		types.MimeNetworkConfigSection, "error adding vApp Network: %s", networkConfigSection)
```
lvirbalas: I'm a bit confused - comment on this function doesn't say anything about vApp network, but the error here does. What kind of network does it add?

govcd/vm_test.go
```diff
@@ -865,6 +869,340 @@ func (vcd *TestVCD) Test_VMPowerOnPowerOff(check *C) {
...
+		vm := NewVM(&vcd.client.Client)
+		vm.VM = &vmType
+
+		var net OrgVDCNetwork
+		var nets []map[string]interface{}
+		var err error
+
+		// Find networks
+		for index := range vcd.config.VCD.Networks {
+			n := make(map[string]interface{})
```
dataclouder: Can we give `n` a more meaningful name?

govcd/vm_test.go
```diff
@@ -865,6 +869,340 @@ func (vcd *TestVCD) Test_VMPowerOnPowerOff(check *C) {
...
+		check.Assert(networkConnectionSection.PrimaryNetworkConnectionIndex, Equals, 0)
+	}
+}
+
+// Test_updateNicParameters_multinic is meant to check functionality of a complicated
+// code structure used in vm.ChangeNetworkConfig which is abstracted into
+// vm.updateNicParameters() method so that it does not contain any API calls, but
+// only adjust the object which is meant to be sent to API. Initially we hit a bug
+// which occurred only when API returned NICs in random order.
+func (vcd *TestVCD) Test_updateNicParameters_multiNIC(check *C) {
```
dataclouder: Isn't this function a good candidate for a separate test framework that does not require gocheck? 
It doesn't use `vcd` and could be isolated.

Didainius: Yes it is. And could benefit from subtests. The only thing is to agree how we do assertions. Do we use `if xxxx { t.Fatal` or some assertion lib/framework.

Didainius: Not necessarily `t.Fatal`, could also be `t.Fail`. But the general point. I think it's more of an agreement question rather than a problem.

dataclouder: If we plan to produce many such tests, we could create a lightweight assertion library. I have made [one for my own usage in a different project](https://github.com/datacharmer/dbdeployer/blob/master/compare/compare.go) just because I wanted to avoid unnecessary dependencies.

govcd/vapp.go
dn1s: @lvirbalas you are right first connection may not be the primary connection. This is just a placeholder for first vm spin up. Network config will be changed later to the correct value of "PrimaryNetworkConnectionIndex". This is a fix for the problem if there is no network attached to a template.

Didainius: Yep. Actually such flow (first create with 'any' parameters and then update to correct ones) seems to be flawed and I suspect that https://github.com/terraform-providers/terraform-provider-vcd/issues/49 and https://github.com/terraform-providers/terraform-provider-vcd/issues/216 are about similar thing. I wanted to leave this PR without these fixes just to finally merge it, but  I'll try to see if we could actually create  network cards with proper configs from beginning instead of using two step approach (create anything, update to correct settings).

govcd/vm_test.go
Didainius: So probably the solution now is to keep it light - I will use plain `if` and we evaluate this in future. I don't know if we have many such tests. In my opinion this is good, but can't predict future.

dataclouder: OK. I leave it to your judgment. If you can isolate this test without much work, go for it.

govcd/vapp.go
lvirbalas: Thank you and hope it won't cause too much changes. I'll keep this thread alive to share your findings in.

govcd/vm.go
```diff
@@ -66,6 +67,46 @@ func (vm *VM) Refresh() error {
 	return err
 }
 
+// AppendNetworkConnection appends a network connection from OrgVDCNetwork to VM
+func (vm *VM) AppendNetworkConnection(orgvdcnetwork *types.OrgVDCNetwork) (Task, error) {
```
vbauzysvmware: What is difference between vapp method `AppendNetworkConfig` ?

govcd/vm.go
```diff
@@ -66,6 +67,46 @@ func (vm *VM) Refresh() error {
...
+	networkConnectionSection.Type = types.MimeNetworkConnectionSection
+	networkConnectionSection.Xmlns = types.XMLNamespaceVCloud
+
+	// Append a new NetworkConnectionSection.NetworkConnection to existing ones
+	networkConnectionSection.NetworkConnection = append(networkConnectionSection.NetworkConnection,
+		&types.NetworkConnection{
+			Network:                 orgvdcnetwork.Name,
+			NetworkConnectionIndex:  len(networkConnectionSection.NetworkConnection),
+			IsConnected:             true,
+			IPAddressAllocationMode: "POOL",
```
vbauzysvmware: Are we sure here always should be 'POOL'

govcd/vm.go
```diff
@@ -66,6 +67,46 @@ func (vm *VM) Refresh() error {
...
+		},
+	)
+
+	apiEndpoint, _ := url.ParseRequestURI(vm.VM.HREF)
+	apiEndpoint.Path += "/networkConnectionSection/"
+
+	return vm.client.ExecuteTaskRequest(apiEndpoint.String(), http.MethodPut,
+		types.MimeNetworkConnectionSection, "error adding VM network connection: %s", networkConnectionSection)
```
vbauzysvmware: Maybe we should elaborate what type of network connection

govcd/vm.go
```diff
@@ -66,6 +67,46 @@ func (vm *VM) Refresh() error {
...
+}
+
+// GetNetworkConnectionSection returns current networks attached to VM
+//
```
vbauzysvmware: Is empty line is intentional?

govcd/vm.go
```diff
@@ -164,53 +205,81 @@ func (vm *VM) ChangeCPUCountWithCore(virtualCpuCount int, coresPerSocket *int) (
...
+				// If we have one NIC only then it is primary by default, otherwise we check for "is_primary" key
+				if (len(networks) == 1) || (network["is_primary"] != nil && network["is_primary"].(bool)) {
+					networkSection.PrimaryNetworkConnectionIndex = tfNicSlot
 				}
 			}
 		}
 	}
+}
+
+func (vm *VM) ChangeNetworkConfig(networks []map[string]interface{}) (Task, error) {
```
vbauzysvmware: Missing function docs

govcd/vm_test.go
```diff
@@ -865,6 +869,340 @@ func (vcd *TestVCD) Test_VMPowerOnPowerOff(check *C) {
...
+			check.Skip("skipping test because no VM is found")
+		}
+
+		fmt.Printf("Running: %s part 1 change ip\n", check.TestName())
+
+		vm := NewVM(&vcd.client.Client)
+		vm.VM = &vmType
+
+		var nets []map[string]interface{}
+		n := make(map[string]interface{})
```
vbauzysvmware: 'n' - please use full variable names

Makefile
```diff
@@ -5,9 +5,16 @@ default: fmtcheck vet build
 
 # test runs the test suite and vets the code
 test: fmtcheck
-	@echo "==> Running Tests"
+	@echo "==> Running Unit Tests"
+	cd govcd && go test -v -tags "unit" .
+	@echo "==> Running Functional Tests"
```
dataclouder: I don't remember what the default behavior for Makefile is. 
What happens if the unit tests fail? Does it continue or does it fail?
Should we consider a `scripts/runtest.sh` file to manage complexity like we have in terraform-provider-vcd?

Didainius: I could probably add that if needed.

It exits. I tested (inserted `t.Error()` in the test) it:
```
$  make test
==> Checking that code complies with gofmt requirements...
==> Running Unit Tests
cd govcd && go test -v -tags "unit" .
=== RUN   TestTags
--- PASS: TestTags (0.00s)
=== RUN   Test_updateNicParameters_multiNIC
=== RUN   Test_updateNicParameters_multiNIC/Ordered_NICs_list
=== RUN   Test_updateNicParameters_multiNIC/Unordered_NIC_list
--- FAIL: Test_updateNicParameters_multiNIC (0.00s)
    vm_unit_test.go:99:
    --- PASS: Test_updateNicParameters_multiNIC/Ordered_NICs_list (0.00s)
    --- PASS: Test_updateNicParameters_multiNIC/Unordered_NIC_list (0.00s)
=== RUN   Test_updateNicParameters_singleNIC
=== RUN   Test_updateNicParameters_singleNIC/IPAllocationModeDHCP
=== RUN   Test_updateNicParameters_singleNIC/IPAllocationModePool
=== RUN   Test_updateNicParameters_singleNIC/IPAllocationModeNone
=== RUN   Test_updateNicParameters_singleNIC/IPAllocationModeManual
=== RUN   Test_updateNicParameters_singleNIC/IPAllocationModeDHCPInvalidIP
--- PASS: Test_updateNicParameters_singleNIC (0.00s)
    --- PASS: Test_updateNicParameters_singleNIC/IPAllocationModeDHCP (0.00s)
    --- PASS: Test_updateNicParameters_singleNIC/IPAllocationModePool (0.00s)
    --- PASS: Test_updateNicParameters_singleNIC/IPAllocationModeNone (0.00s)
    --- PASS: Test_updateNicParameters_singleNIC/IPAllocationModeManual (0.00s)
    --- PASS: Test_updateNicParameters_singleNIC/IPAllocationModeDHCPInvalidIP (0.00s)
FAIL
FAIL    github.com/vmware/go-vcloud-director/v2/govcd   0.014s
make: *** [test] Error 1
```

govcd/vdc.go
```diff
@@ -310,6 +310,13 @@ func (vdc *Vdc) ComposeVApp(orgvdcnetworks []*types.OrgVDCNetwork, vapptemplate
 	if vapptemplate.VAppTemplate.Children == nil || orgvdcnetworks == nil {
 		return Task{}, fmt.Errorf("can't compose a new vApp, objects passed are not valid")
 	}
+
+	primaryNetworkConnectionIndex := 0
```
lvirbalas: Please add a comment above this block explaining what it's doing and what primaryNetworkConnectionIndex really means.

govcd/vapp.go
```diff
@@ -91,83 +91,98 @@ func (vapp *VApp) Refresh() error {
...
+		return Task{}, fmt.Errorf("vApp Template must have at least one child VM")
+	}
+
+	// Status 8 means The object is resolved and powered off.
+	// https://vdc-repo.vmware.com/vmwb-repository/dcr-public/94b8bd8d-74ff-4fe3-b7a4-41ae31516ed7/1b42f3b5-8b31-4279-8b3f-547f6c7c5aa8/doc/GUID-843BE3AD-5EF6-4442-B864-BCAE44A51867.html
+	if vappTemplate.VAppTemplate.Status != 8 {
+		return Task{}, fmt.Errorf("vApp Template shape is not ok (status: %d)", vappTemplate.VAppTemplate.Status)
+	}
+
+	// TODO - validate network *types.NetworkConnectionSection for mandatory parameters
```
lvirbalas: Are we sure on how to validate these parameters? If yes, meaning we won't put blockers on use cases we may not foresee, let's add validation and error messages now.

govcd/vdc.go
lvirbalas: Awesome!

govcd/vm.go
```diff
@@ -164,53 +168,91 @@ func (vm *VM) ChangeCPUCountWithCore(virtualCpuCount int, coresPerSocket *int) (
...
+				networkSection.NetworkConnection[loopIndex].NeedsCustomization = true
+				networkSection.NetworkConnection[loopIndex].IsConnected = true
+				networkSection.NetworkConnection[loopIndex].IPAddress = ipAddress
+				networkSection.NetworkConnection[loopIndex].IPAddressAllocationMode = ipAllocationMode
+
+				// for IPAllocationModeNone we hardcode special network name used by vcd 'none'
+				if ipAllocationMode == types.IPAllocationModeNone {
+					networkSection.NetworkConnection[loopIndex].Network = types.NoneNetwork
+				} else {
+					// Network names can be stored in two forms 'name' or 'network_name'. We pick
```
lvirbalas: Sorry, but where are these two variants `name` and `network_name` coming from?

govcd/vapp.go
lvirbalas: Thanks, just one question to be safe: with the new validations it's still possible to add VM without a network connection still, correct?

govcd/vm.go
Didainius: They come in from client consuming `func (vm *VM) ChangeNetworkConfig(networks []map[string]interface{}) (Task, error)`. But you are right in this case - we're not even using this function now. Premature optimization probably....

govcd/vapp.go
Didainius: It accepted empty struct pointer `&types.NetworkConnectionSection{}`, but did not work if `nil` was passed for network. Added test for `nil` option as well. Last run and will push.

govcd/vm.go
Didainius: Removed

govcd/vapp.go
Didainius: Pushed

PR #189, opened 2019-05-09
Fix bug in external network

vbauzysvmware:
Issue regarding not finished task was solved in handling task. This was one rare cases when task returned as part of array. No issue on vCD.


govcd/system.go
```diff
@@ -207,8 +207,15 @@ func CreateExternalNetwork(vcdClient *VCDClient, externalNetwork *types.External
 	externalNetwork.Configuration.FenceMode = "isolated"
 
 	// Return the task
-	return vcdClient.Client.ExecuteTaskRequest(externalNetHREF.String(), http.MethodPost,
+	task, err := vcdClient.Client.ExecuteTaskRequest(externalNetHREF.String(), http.MethodPost,
 		types.MimeExternalNetwork, "error instantiating a new ExternalNetwork: %s", externalNetwork)
+
+	// Real task in task array
+	if err == nil {
+		task.Task = task.Task.Tasks.Task[0]
```
dataclouder: Please check that `task.Task.Tasks.Task[0]` exists before using it

vbauzysvmware: Added check

Issue #130, opened 2018-12-13
implement edge gateway operations


dataclouder:
To make a VDC immediately useful, we need the ability of creating/updating/removing an [Edge Gateway](https://code.vmware.com/apis/287/vcloud#/doc/doc/types/GatewayType.html).

Related to #14, #18, #129

lvirbalas:
Thank you for this issue, @dataclouder 
Please post a suggested Terraform resource definition too.

PR #195, opened 2019-05-27
Edge gateway creation

dataclouder:
Implements Issue #130:

Add functions `CreateEdgeGateway()` and `(egw *EdgeGateway) Delete()`.

1. To create an edge gateway, we need to pass at least 7 parameters, possibly more. My current choice is to pass a simplified structure containing only the necessary parameters, and deal with other customisations in a different function. Another possibility would be passing the API customisation structure to the creation. We do this (for legacy reasons) for networks but I am not sure whether we want to burden client applications with such complexity. I could expose two functions: one that use the current parameters and one that uses the full customisation. TBD. 
2. A strange thing happens on creation, because the task returned by vCD creates the edge gateway entity but does not deploy the VM that is used for that. Another task starts immediately after. For this reason, the first task is refreshed right in the creation function, and the second task gets returned.
3. When creating edge gateways, the GUI mentions Compact, Large, Xlarge, and Quad Large. In the API, I only find something that mentions compact, and it's the "GatewayBackingConfig", which only accepts "compact" and "full". Probably the two things are not related. 
4. Testing edge gateways is a unique challenge, because in Cassini servers we can only set an usable edge gateway when we set the first one. This is because the first external network has already taken the only port group of type "NETWORK" that was available. For this reason, I did some refactoring to the external network testing, to be able to create external networks outside their current tests. I was able to remove duplicated code while keeping the functionality intact.
5. Another change that was necessary is a little refactoring of the function that use the catalog ID. The current one used a slice to get the ID, but this only works when the prefix has a fixed length. When the entity name changes, the prefix is different, and the slice method cannot be used. So I implemented it using a regular expression, and I added a unit test to make sure it works in most cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
-## 2.2.0 (Unreleased)
+## 2.3.0 (Unreleased)
+
+* Added edge gateway create/delete functions [#130](https://github.com/vmware/go-vcloud-director/issues/130).
+
+## 2.2.0 (May 2, 2019)
 
 FEATURES:
 
@@ -8,7 +12,6 @@ FEATURES:
 using VCDClient.APIVCDMaxVersionIs(string) and VCDClient.APIClientVersionIs(string).
 * Added ability to override currently used vCD API version WithAPIVersion(string) [#174](https://github.com/vmware/go-vcloud-director/pull/174).
 * Added ability to enable nested hypervisor option for VM with VM.ToggleNestedHypervisor(bool) [#219](https://github.com/terraform-providers/terraform-provider-vcd/issues/219).
-* Added edge gateway create/delete functions [#130](https://github.com/vmware/go-vcloud-director/issues/130).
 
 
 BREAKING CHANGES:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ FEATURES:
 using VCDClient.APIVCDMaxVersionIs(string) and VCDClient.APIClientVersionIs(string).
 * Added ability to override currently used vCD API version WithAPIVersion(string) [#174](https://github.com/vmware/go-vcloud-director/pull/174).
 * Added ability to enable nested hypervisor option for VM with VM.ToggleNestedHypervisor(bool) [#219](https://github.com/terraform-providers/terraform-provider-vcd/issues/219).
-* Added edge gateway create/delete functions
+* Added edge gateway create/delete functions [#130](https://github.com/vmware/go-vcloud-director/issues/130).
 
 
 BREAKING CHANGES:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 * Added edge gateway create/delete functions [#130](https://github.com/vmware/go-vcloud-director/issues/130).
 
-## 2.2.0 (May 2, 2019)
+## 2.2.0 (May 15, 2019)
 
 FEATURES:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ FEATURES:
 using VCDClient.APIVCDMaxVersionIs(string) and VCDClient.APIClientVersionIs(string).
 * Added ability to override currently used vCD API version WithAPIVersion(string) [#174](https://github.com/vmware/go-vcloud-director/pull/174).
 * Added ability to enable nested hypervisor option for VM with VM.ToggleNestedHypervisor(bool) [#219](https://github.com/terraform-providers/terraform-provider-vcd/issues/219).
+* Added edge gateway create/delete functions
 
 
 BREAKING CHANGES:
--- a/govcd/edgegateway.go
+++ b/govcd/edgegateway.go
@@ -648,7 +648,7 @@ func (egw *EdgeGateway) DeleteAsync(force bool, recursive bool) (Task, error) {
 	util.Logger.Printf("[TRACE] EdgeGateway.Delete - deleting edge gateway with force: %t, recursive: %t", force, recursive)
 
 	if egw.EdgeGateway.HREF == "" {
-		return Task{}, fmt.Errorf("cannot delete, Object is empty")
+		return Task{}, fmt.Errorf("cannot delete, HREF is missing")
 	}
 
 	egwUrl, err := url.ParseRequestURI(egw.EdgeGateway.HREF)
--- a/govcd/edgegateway.go
+++ b/govcd/edgegateway.go
@@ -644,7 +644,7 @@ func (eGW *EdgeGateway) RemoveIpsecVPN() (Task, error) {
 
 // Deletes the edge gateway, returning a task and an error if the vCD call fails.
 // https://code.vmware.com/apis/442/vcloud-director/doc/doc/operations/DELETE-EdgeGateway.html
-func (egw *EdgeGateway) DeleteTask(force bool, recursive bool) (Task, error) {
+func (egw *EdgeGateway) DeleteAsync(force bool, recursive bool) (Task, error) {
 	util.Logger.Printf("[TRACE] EdgeGateway.Delete - deleting edge gateway with force: %t, recursive: %t", force, recursive)
 
 	if egw.EdgeGateway.HREF == "" {
@@ -675,35 +675,10 @@ func (egw *EdgeGateway) DeleteTask(force bool, recursive bool) (Task, error) {
 // https://code.vmware.com/apis/442/vcloud-director/doc/doc/operations/DELETE-EdgeGateway.html
 func (egw *EdgeGateway) Delete(force bool, recursive bool) error {
 
-	task, err := egw.DeleteTask(force, recursive)
+	task, err := egw.DeleteAsync(force, recursive)
 	if err != nil {
 		return err
 	}
-	/*
-		util.Logger.Printf("[TRACE] EdgeGateway.Delete - deleting edge gateway with force: %t, recursive: %t", force, recursive)
-
-		if egw.EdgeGateway.HREF == "" {
-			return  fmt.Errorf("cannot delete, Object is empty")
-		}
-
-		egwUrl, err := url.ParseRequestURI(egw.EdgeGateway.HREF)
-		if err != nil {
-			return  fmt.Errorf("error parsing edge gateway url: %s", err)
-		}
-
-		req := egw.client.NewRequest(map[string]string{
-			"force":     strconv.FormatBool(force),
-			"recursive": strconv.FormatBool(recursive),
-		}, http.MethodDelete, *egwUrl, nil)
-		resp, err := checkResp(egw.client.Http.Do(req))
-		if err != nil {
-			return  fmt.Errorf("error deleting edge gateway: %s", err)
-		}
-		task := NewTask(egw.client)
-		if err = decodeBody(resp, task.Task); err != nil {
-			return  fmt.Errorf("error decoding task response: %s", err)
-		}
-	*/
 	if task.Task.Status == "error" {
 		return fmt.Errorf(combinedTaskErrorMessage(task.Task, fmt.Errorf("edge gateway not properly destroyed")))
 	}
@@ -734,4 +709,21 @@ func (egw *EdgeGateway) GetNetworks() ([]SimpleNetworkIdentifier, error) {
 	}
 
 	return networks, nil
+}
+
+// HasDefaultGateway returns true if the edge gateway uses one of the external
+// networks as default gateway
+func (egw *EdgeGateway) HasDefaultGateway() bool {
+	if egw.EdgeGateway.Configuration != nil &&
+		egw.EdgeGateway.Configuration.GatewayInterfaces != nil {
+		for _, gw := range egw.EdgeGateway.Configuration.GatewayInterfaces.GatewayInterface {
+			if gw.UseForDefaultRoute &&
+				gw.SubnetParticipation != nil &&
+				gw.SubnetParticipation.Gateway != "" &&
+				gw.SubnetParticipation.Netmask != "" {
+				return true
+			}
+		}
+	}
+	return false
 }
--- a/govcd/edgegateway.go
+++ b/govcd/edgegateway.go
@@ -642,7 +642,7 @@ func (eGW *EdgeGateway) RemoveIpsecVPN() (Task, error) {
 	return eGW.AddIpsecVPN(ipsecVPNConfig)
 }
 
-// Deletes the edge gateway, returning a task and an error if the vCD call fails.
+// Deletes the edge gateway, returning a task and an error with the operation result.
 // https://code.vmware.com/apis/442/vcloud-director/doc/doc/operations/DELETE-EdgeGateway.html
 func (egw *EdgeGateway) DeleteAsync(force bool, recursive bool) (Task, error) {
 	util.Logger.Printf("[TRACE] EdgeGateway.Delete - deleting edge gateway with force: %t, recursive: %t", force, recursive)
@@ -671,7 +671,7 @@ func (egw *EdgeGateway) DeleteAsync(force bool, recursive bool) (Task, error) {
 	return *task, err
 }
 
-// Deletes the edge gateway, returning an error if the vCD call fails.
+// Deletes the edge gateway, returning an error with the operation result.
 // https://code.vmware.com/apis/442/vcloud-director/doc/doc/operations/DELETE-EdgeGateway.html
 func (egw *EdgeGateway) Delete(force bool, recursive bool) error {
 
--- a/govcd/edgegateway.go
+++ b/govcd/edgegateway.go
@@ -24,6 +24,7 @@ type EdgeGateway struct {
 	client      *Client
 }
 
+// Simplified structure used to list networks connected to an edge gateway
 type SimpleNetworkIdentifier struct {
 	Name          string
 	InterfaceType string
@@ -641,9 +642,9 @@ func (eGW *EdgeGateway) RemoveIpsecVPN() (Task, error) {
 	return eGW.AddIpsecVPN(ipsecVPNConfig)
 }
 
-// Deletes the edge gateway, returning an error if the vCD call fails.
+// Deletes the edge gateway, returning a task and an error if the vCD call fails.
 // https://code.vmware.com/apis/442/vcloud-director/doc/doc/operations/DELETE-EdgeGateway.html
-func (egw *EdgeGateway) Delete(force bool, recursive bool) (Task, error) {
+func (egw *EdgeGateway) DeleteTask(force bool, recursive bool) (Task, error) {
 	util.Logger.Printf("[TRACE] EdgeGateway.Delete - deleting edge gateway with force: %t, recursive: %t", force, recursive)
 
 	if egw.EdgeGateway.HREF == "" {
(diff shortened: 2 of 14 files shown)


govcd/system_test.go
```diff
@@ -102,6 +104,80 @@ func (vcd *TestVCD) Test_CreateOrg(check *C) {
 	doesOrgExist(check, vcd)
 }
 
+func (vcd *TestVCD) Test_CreateDeleteEdgeGateway(check *C) {
+	skippingReason, externalNetwork, task, err := vcd.testCreateExternalNetwork(check.TestName(), TestCreateExternalNetwork, "")
+	// skippingReason, _ , task, err := vcd.testCreateExternalNetwork(check.TestName(), TestCreateExternalNetwork, "")
```
Didainius: I suspect it's a leftover.

govcd/system.go
```diff
@@ -8,11 +8,26 @@ import (
...
+	OrgName                   string
+	VdcName                   string
+	Name                      string
+	Description               string
+	BackingConfiguration      string
+	AdvancedNetworkingEnabled bool
+}
+
+// List of allowed values for GatewayBackingConfig.
+var EGWAllowedBackingConfig = []string{"compact", "full"}
```
Didainius: API doc mentions `One of: compact, full, full4`. I see it also mentioning special endpoint https://vdc-repo.vmware.com/vmwb-repository/dcr-public/76f491b4-679c-4e1e-8428-f813d668297a/a2555a1b-22f1-4cca-b481-2a98ab874022/doc/doc/types/EdgeGatewayFormFactorType.html

Maybe it also accepts it here. Users may find it useful for more thorough size configuration.

govcd/system.go
```diff
@@ -43,6 +58,134 @@ func CreateOrg(vcdClient *VCDClient, name string, fullName string, description s
...
+	// [][]string{[]string{"urn:vcloud:catalog:97384890-180c-4563-b9b7-0dc50a2430b0", "97384890-180c-4563-b9b7-0dc50a2430b0"}}
+	if len(matchList) == 0 || len(matchList[0]) < 2 {
+		return "", fmt.Errorf("error extracting ID from '%s'", entityId)
+	}
+	return matchList[0][1], nil
+}
+
+// Creates an edge gateway using its parents entities
+// https://code.vmware.com/apis/442/vcloud-director/doc/doc/operations/POST-CreateEdgeGateway.html
+func CreateEdgeGateway(vcdClient *VCDClient, egwc EdgeGatewayCreation) (Task, error) {
```
Didainius: In addition to this I would probably vote for that "wider" function which accepts `types.EdgeGateway`. While it is more complicated (and is harder, if possible at all, to validate), it would allow more flexibility which may be needed in close future:

* One of them would be `HaEnabled` setting which should be used quite often in production environments. 
* The other I'm interested (for load balancers for now) is `< EdgeGatewayServiceConfiguration >` which [would allow to enable ](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/76f491b4-679c-4e1e-8428-f813d668297a/a2555a1b-22f1-4cca-b481-2a98ab874022/doc/doc/types/NetworkServiceType.html) NatService, LoadBalancerService IpsecVpnService/GatewayIpsecVpnService, DhcpService/GatewayDhcpService, FirewallService, or StaticRoutingService. At least for load balancers I will have to use it to enable load balancing at edge gateway level.

API doc ref:
https://code.vmware.com/apis/72/vcloud/doc/doc/types/GatewayType.html

govcd/system_test.go
dataclouder: yes. It was. Deleting ...

govcd/edgegateway.go
```diff
@@ -634,3 +640,53 @@ func (eGW *EdgeGateway) RemoveIpsecVPN() (Task, error) {
 	}
 	return eGW.AddIpsecVPN(ipsecVPNConfig)
 }
+
+// Deletes the edge gateway, returning an error if the vCD call fails.
+// https://code.vmware.com/apis/442/vcloud-director/doc/doc/operations/DELETE-EdgeGateway.html
+func (egw *EdgeGateway) Delete(force bool, recursive bool) (Task, error) {
```
vbauzysvmware: According our previous discussions and decisions I would make Delete sync method which returns only error and DeleteTask or similar name which return async Task to handle by user themself. Edge gtw only handle one action that's why I propose to do that.

govcd/edgegateway.go
```diff
@@ -634,3 +640,53 @@ func (eGW *EdgeGateway) RemoveIpsecVPN() (Task, error) {
...
+	if egw.EdgeGateway.HREF == "" {
+		return Task{}, fmt.Errorf("cannot delete, Object is empty")
+	}
+
+	egwUrl, err := url.ParseRequestURI(egw.EdgeGateway.HREF)
+	if err != nil {
+		return Task{}, fmt.Errorf("error parsing edge gateway url: %s", err)
+	}
+
+	req := egw.client.NewRequest(map[string]string{
```
vbauzysvmware: Why ExecuteTaskRequest helper function wasn't used?

govcd/edgegateway.go
```diff
@@ -634,3 +640,53 @@ func (eGW *EdgeGateway) RemoveIpsecVPN() (Task, error) {
...
+		return Task{}, fmt.Errorf("error decoding task response: %s", err)
+	}
+	if task.Task.Status == "error" {
+		return Task{}, fmt.Errorf("edge gateway not properly destroyed")
+	}
+	return *task, nil
+}
+
+// Returns the list of networks associated with an edge gateway
+func (egw *EdgeGateway) Networks() ([]SimpleNetworkIdentifier, error) {
```
vbauzysvmware: I would like more explicit  naming `getNetorks`

govcd/system.go
dataclouder: I tried, but the vCD doesn't agree:
```
value *errors.errorString = &errors.errorString{s:"backing configuration value 'full4' not accepted. Allowed values: [compact full] "
```

vbauzysvmware: Did you figure out how to match other in UI visible configs?

lvirbalas: What about removing this validation all together? It's better to pass-up the vCD error to the user. We would avoid putting ourselves into a corner if there's a supported value that our vCDs don't support, but others do.

govcd/system.go
dataclouder: The only practical problem with this approach is that to create an edge gateway we still need 5 parameters. 
There are three possibilities:

1. Additional function
`func CreateAndConfigureEdgeGateway(vcdClient *VCDClient, orgName, vdcName, egwName string, egwConfiguration *types.EdgeGateway ) (Task, error)` .  With this, the original creation function becomes a wrapper.

2. Add `egwConfiguration *types.EdgeGateway` to CreateEdgeGateway. If the configuration parameter is nil, it will be ignored.

3. Leave `CreateEdgeGateway` as it is, and add an `UpdateEdgeGateway` function that [updates the existing parameters](https://code.vmware.com/apis/553/vcloud-director/doc/doc/operations/PUT-EdgeGateway.html)

lvirbalas: > One of them would be HaEnabled setting which should be used quite often in production environments.

Actually, HaEnabled should be in the normal function and exposed through Terraform. It is there in the GUI after all.

govcd/system.go
dataclouder: We do this kind of validation for API values in several places. 
I want to have this list of values to use it from Terraform provider, where we are better off shielding users from vCD error messages. So far, this works for 9.1, 9.5, and 9.7.

dataclouder: As mentioned above, these are the only values accepted. 
The values in the GUI refer to something else.

lvirbalas: This is a general comment about this:
> we are better off shielding users from vCD error messages

If the message is meaningful, we **shouldn't** be shielding the user from it. On the opposite, we should be passing the message through. The users are vCD admins they wan't visibility. I'm of course referring to the message only, but not the whole response.

govcd/edgegateway.go
dataclouder: Because I didn't find a variant of `ExecuteTaskRequest` that also accepts header parameters

govcd/system.go
dataclouder: I don't understand this reasoning. I am referring at the many `Validate` clause that we have in our resources. 
For example: 

`ValidateFunc: validation.StringInSlice([]string{"AllocationVApp", "AllocationPool", "ReservationPool"}, false),`
(https://github.com/terraform-providers/terraform-provider-vcd/blob/277d63eadffac079f47cd374b50978e854ac9e0b/vcd/resource_vcd_org_vdc.go#L65)

If we want to throw bare error messages to the users, why bother checking that we are accepting the right values?
Perhaps we are talking about different things here?

lvirbalas: I was talking more generally.

The challenge with this particular field is that there should be more acceptable values ("full4"), though we can't prove that on our vCD instances.

govcd/system_test.go

govcd/system.go
dataclouder: List removed

govcd/system.go
dataclouder: Done. The simplified structure has been updated, and a separate function that accepts a full structure has been added

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ FEATURES:
 using VCDClient.APIVCDMaxVersionIs(string) and VCDClient.APIClientVersionIs(string).
 * Added ability to override currently used vCD API version WithAPIVersion(string) [#174](https://github.com/vmware/go-vcloud-director/pull/174).
 * Added ability to enable nested hypervisor option for VM with VM.ToggleNestedHypervisor(bool) [#219](https://github.com/terraform-providers/terraform-provider-vcd/issues/219).
+* Added edge gateway create/delete functions
```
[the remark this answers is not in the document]
dataclouder: fixed

govcd/edgegateway.go
```diff
@@ -23,6 +24,11 @@ type EdgeGateway struct {
 	client      *Client
 }
 
+type SimpleNetworkIdentifier struct {
```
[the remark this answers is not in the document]
dataclouder: done

govcd/edgegateway.go
dataclouder: done. We now have two functions: `Delete` returns an error, and `DeleteTask` returns task + error.

govcd/edgegateway.go
dataclouder: done

govcd/system.go
```diff
@@ -43,6 +58,134 @@ func CreateOrg(vcdClient *VCDClient, name string, fullName string, description s
 
 }
 
+// Returns the UUID part of an entity ID
+// From "urn:vcloud:vdc:72fefde7-4fed-45b8-a774-79b72c870325",
+// will return "72fefde7-4fed-45b8-a774-79b72c870325"
+// From "urn:vcloud:catalog:97384890-180c-4563-b9b7-0dc50a2430b0"
+// will return "97384890-180c-4563-b9b7-0dc50a2430b0"
+func getBareEntityId(entityId string) (string, error) {
```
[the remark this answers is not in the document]
dataclouder: done

govcd/system.go
```diff
@@ -8,11 +8,26 @@ import (
...
+// Simple structure to pass Edge Gateway creation parameters.
+type EdgeGatewayCreation struct {
+	ExternalNetworks          []string
+	OrgName                   string
+	VdcName                   string
+	Name                      string
+	Description               string
+	BackingConfiguration      string
+	AdvancedNetworkingEnabled bool
+}
```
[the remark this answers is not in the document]
dataclouder: Done. 
Now the simplified structure includes a DefaultGateway item, which is used to set the default gateway. If the item is empty, the default gateway is left unset.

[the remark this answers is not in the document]
lvirbalas: Thanks!

govcd/edgegateway.go
```diff
@@ -634,3 +641,97 @@ func (eGW *EdgeGateway) RemoveIpsecVPN() (Task, error) {
...
+		}, http.MethodDelete, *egwUrl, nil)
+		resp, err := checkResp(egw.client.Http.Do(req))
+		if err != nil {
+			return  fmt.Errorf("error deleting edge gateway: %s", err)
+		}
+		task := NewTask(egw.client)
+		if err = decodeBody(resp, task.Task); err != nil {
+			return  fmt.Errorf("error decoding task response: %s", err)
+		}
+	*/
```
lvirbalas: Too much of commented out code ;)

govcd/edgegateway.go
```diff
@@ -634,3 +641,97 @@ func (eGW *EdgeGateway) RemoveIpsecVPN() (Task, error) {
 	}
 	return eGW.AddIpsecVPN(ipsecVPNConfig)
 }
+
+// Deletes the edge gateway, returning a task and an error if the vCD call fails.
+// https://code.vmware.com/apis/442/vcloud-director/doc/doc/operations/DELETE-EdgeGateway.html
+func (egw *EdgeGateway) DeleteTask(force bool, recursive bool) (Task, error) {
```
lvirbalas: Naming: please check in the codebase whether this is the first precedence of providing sync and async function variants. We need to agree on how to call these functions. By this code you're suggesting `x` for sync and `xTask` for async. Personally for me, when I see `DeleteTask` first it makes me think it "deletes some task" :) What about the following scheme:
- `Delete` - synchronous function(s)
- `DeleteAsync` - function(s) that return a task

@Didainius and @vbauzysvmware please chime in with your preferences/thoughts.

dataclouder: I am OK with `DeleteAsync`.
I had interpreted your `+1` to Vaidotas request as an endorsement of `DeleteTask`

govcd/edgegateway.go
dataclouder: Fell through the cracks. Removing

govcd/edgegateway.go
dataclouder: Done

govcd/edgegateway.go
dataclouder: done

govcd/system.go
vbauzysvmware: For the sake to figure out - I would try to pass parameters using more later API version.

govcd/edgegateway.go
vbauzysvmware: I think our intention by default user to choose not to handle tasks - so main function `Delete`.
And another one which returns tasks. `DeleteAsync` - not for my taste, but at the moment don't have better suggestions as ''DeleteAndReturnTask` or others seems worse.
According current govcd, we have `Delete` which returns tasks and `DeleteWait` which handles tasks in some cases.

dataclouder: If we voted on hiding task, I was surely against. While I appreciate the simplicity of a function that simply does the job, I want to be able to monitor what is happening. Tasks are an essential part of the API, and I want to use them for monitoring and troubleshooting. 
I can change the name to whatever is decided, but I want to have tasks available for potentially long lasting operations.

vbauzysvmware: I agree to leave two functions, no discussion here.

lvirbalas: After more digging, we already have a precedence for naming *synchronous* tasks:

Screen Shot 2019-06-11 at 11 17 47

vbauzysvmware: Linas, I noted that already :)

lvirbalas: We decided to have:
1. Default functions (the simply named ones) as synchronous
2. Corresponding asynchronous functions be named as `xAsync`
3. Default synchronous functions returning objects and these should be references (as opposed to copies)

We're not refactoring the whole library now, but use this as a guideline for all future functions.

dataclouder: Done!

Functions available for creation:
* `CreateEdgeGateway`: uses a simplified structure, returns an edge gateway and an error
* `CreateAndConfigureEdgeGateway`: uses the full edge gateway structure, returns an edge gateway and an error.
* `CreateEdgeGatewayAsync`: uses a simplified structure, returns a task and an error
* `CreateAndConfigureEdgeGatewayAsync`: uses the full edge gateway structure, returns a task and an error

Delete functions follow the same pattern:
* `EdgeGateway.DeleteAsync` returns a task and an error
* `EdgeGateway.Delete` returns an error

govcd/system.go
dataclouder: There is no explicit check anymore. Wrong values will be reported back by vCD

govcd/system.go
```diff
@@ -43,6 +59,213 @@ func CreateOrg(vcdClient *VCDClient, name string, fullName string, description s
...
+	}
+	egw, err := vdc.FindEdgeGateway(egwc.Name)
+	if err != nil {
+		return EdgeGateway{}, err
+	}
+	return egw, nil
+}
+
+// CreateAndConfigureEdgeGateway creates an edge gateway using a full configuration structure
+func CreateAndConfigureEdgeGateway(vcdClient *VCDClient, orgName, vdcName, egwName string, egwConfiguration *types.EdgeGateway) (EdgeGateway, error) {
```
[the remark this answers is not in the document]
dataclouder: This would create a mismatch where we need to use the result of this function in combination with other functions that return a structure containing pointers (the default in all govcd). Either we change all of them (meaning we change the type) or we leave them as they are.
In my case, I would need to deal with structure and pointer within adjacent statements, and I'd like to avoid this situation.

govcd/system.go
```diff
@@ -43,6 +59,213 @@ func CreateOrg(vcdClient *VCDClient, name string, fullName string, description s
...
+	return createEdgeGateway(vcdClient, EdgeGatewayCreation{OrgName: orgName, VdcName: vdcName, Name: egwName}, egwConfiguration)
+}
+
+// CreateEdgeGateway creates an edge gateway using a simplified configuration structure
+func CreateEdgeGateway(vcdClient *VCDClient, egwc EdgeGatewayCreation) (EdgeGateway, error) {
```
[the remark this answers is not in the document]
dataclouder: See above answer

govcd/system.go
```diff
@@ -43,6 +59,213 @@ func CreateOrg(vcdClient *VCDClient, name string, fullName string, description s
...
+				Task:   innerTask,
+				client: &vcdClient.Client,
+			}
+			return deployTask, nil
+		}
+	}
+	return Task{}, fmt.Errorf("no deployment task found for edge gateway %s - The edge gateway might have been created, but not deployed properly", egwName)
+}
+
+func createEdgeGateway(vcdClient *VCDClient, egwc EdgeGatewayCreation, egwConfiguration *types.EdgeGateway) (EdgeGateway, error) {
```
vbauzysvmware: Please add comment :)

govcd/system.go
```diff
@@ -43,6 +59,213 @@ func CreateOrg(vcdClient *VCDClient, name string, fullName string, description s
...
+		return EdgeGateway{}, err
+	}
+	vdc, err := org.GetVdcByName(egwc.VdcName)
+	if err != nil {
+		return EdgeGateway{}, err
+	}
+	egw, err := vdc.FindEdgeGateway(egwc.Name)
+	if err != nil {
+		return EdgeGateway{}, err
+	}
```
vbauzysvmware: In this case very nice to return object, cause finding edge gateway is complicated :)

govcd/system.go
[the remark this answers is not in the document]
Didainius: Ok. This means not this time at least.

dataclouder:
>Looks like the date should be May 15:
fixed

govcd/system.go
dataclouder: done

govcd/edgegateway.go
```diff
@@ -634,3 +641,89 @@ func (eGW *EdgeGateway) RemoveIpsecVPN() (Task, error) {
 	}
 	return eGW.AddIpsecVPN(ipsecVPNConfig)
 }
+
+// Deletes the edge gateway, returning a task and an error if the vCD call fails.
```
lvirbalas: Is the comment really right? Isn't the function returning a task even when delete succeeds?

govcd/system.go
```diff
@@ -149,10 +375,8 @@ func QueryPortGroups(vcdCli *VCDClient, filter string) ([]*types.PortGroupRecord
 	return results.Results.PortGroupRecord, nil
 }
 
-// GetExternalNetwork returns ExternalNetwork object if user specifies a valid external network name.
-// If no valid external network is found, it returns an empty
-// ExternalNetwork and no error. Otherwise it returns an error and an empty
-// ExternalNetwork object
+// GetExternalNetwork returns an ExternalNetwork object if user the network name matches an existing one.
```
lvirbalas: Nit, maybe better: returns *reference* to an ExternalNetwork?

govcd/task.go
```diff
@@ -126,6 +128,22 @@ func (task *Task) WaitInspectTaskCompletion(inspectionFunc InspectionFunc, delay
 			return nil
 		}
 
+		// If the environment variable "GOVCD_TASK_MONITOR" is set, its value
```
lvirbalas: It seems we're missing a section in `TESTING.md` describing what environment variables there are. It's kind of a mystery right now.

dataclouder: done

govcd/system.go
dataclouder: fixed

govcd/edgegateway.go
dataclouder: fixed

govcd/task.go
lvirbalas: Beautiful, thanks!

govcd/edgegateway.go
```diff
@@ -634,3 +641,89 @@ func (eGW *EdgeGateway) RemoveIpsecVPN() (Task, error) {
...
 	return eGW.AddIpsecVPN(ipsecVPNConfig)
 }
+
+// Deletes the edge gateway, returning a task and an error with the operation result.
+// https://code.vmware.com/apis/442/vcloud-director/doc/doc/operations/DELETE-EdgeGateway.html
+func (egw *EdgeGateway) DeleteAsync(force bool, recursive bool) (Task, error) {
+	util.Logger.Printf("[TRACE] EdgeGateway.Delete - deleting edge gateway with force: %t, recursive: %t", force, recursive)
+
+	if egw.EdgeGateway.HREF == "" {
+		return Task{}, fmt.Errorf("cannot delete, Object is empty")
```
vbauzysvmware:
```suggestion
		return Task{}, fmt.Errorf("cannot delete, HREF is empty")
```

govcd/edgegateway_test.go
```diff
@@ -227,8 +231,49 @@ func (vcd *TestVCD) Test_AddIpsecVPN(check *C) {
...
 	newConfState = afterDeletionConf.GatewayIpsecVpnService.IsEnabled
 	newConfTunnel = afterDeletionConf.GatewayIpsecVpnService.Tunnel
-	newConfEndpoint = afterDeletionConf.GatewayIpsecVpnService.Endpoint
+	newConfEndpoint := afterDeletionConf.GatewayIpsecVpnService.Endpoint
 	check.Assert(newConfState, Equals, false)
 	check.Assert(newConfTunnel, IsNil)
 	check.Assert(newConfEndpoint, IsNil)
 }
+
+func (vcd *TestVCD) TestEdgeGateway_Networks(check *C) {
```
vbauzysvmware:
```suggestion
func (vcd *TestVCD) TestEdgeGateway_GetNetworks(check *C) {
```

govcd/system.go
```diff
@@ -43,6 +59,216 @@ func CreateOrg(vcdClient *VCDClient, name string, fullName string, description s
...
+		return Task{}, err
+	}
+	vdc, err := adminOrg.GetVdcByName(vdcName)
+	if err != nil {
+		return Task{}, err
+	}
+
+	egwCreateHREF := vcdClient.Client.VCDHREF
+
+	ID, err := getBareEntityUuid(vdc.Vdc.ID)
```
vbauzysvmware:
```suggestion
	vdcId, err := getBareEntityUuid(vdc.Vdc.ID)
```

dataclouder: fixed

govcd/edgegateway_test.go
dataclouder: fixed

govcd/edgegateway.go
dataclouder: message improved

vbauzysvmware:
LGTM. If I understand correctly current implementation support only two types of edge gateways. Not 4 types which we see in UI. Maybe we should state that in changelog and function docs?

vbauzysvmware:
I suggest to create Issues for matter I mentioned and for import/update too.

PR #197, opened 2019-06-04
Added new function for remove all networks attached to VAPP

vbauzysvmware:
Needs for https://github.com/terraform-providers/terraform-provider-vcd/pull/255

VAPP should be cleaned before deletion

### DIFF
--- a/govcd/vapp_test.go
+++ b/govcd/vapp_test.go
@@ -716,4 +716,3 @@ func (vcd *TestVCD) Test_RemoveAllNetworks(check *C) {
 	}
 	check.Assert(isExist, Equals, false)
 }
-networkConfig.NetworkName == networkName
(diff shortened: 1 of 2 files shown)


govcd/vapp_test.go
```diff
@@ -627,3 +627,93 @@ func verifyNetworkConnectionSection(check *C, actual, desired *types.NetworkConn
...
+	check.Assert(err, IsNil)
+
+	isExist := false
+	for _, networkConfig := range networkConfig.NetworkConfig {
+		if networkConfig.NetworkName == networkName || networkConfig.NetworkName == networkName2 {
+			isExist = true
+		}
+
+	}
+	check.Assert(isExist, Equals, false)
```
dataclouder: Please change `isExist` to either `exists` or `hasNetworks`.

vbauzysvmware: Changed

Issue #199, opened 2019-06-12
Add GetVirtualHardwareSection function to vm.go


frenchtoasters:
There is a `vm.GetNetworkConnectionSection()` function however there is not a `vm.GetVirtualHardwareSection()`. As defined [here](https://www.vmware.com/support/vcd/doc/rest-api-doc-1.5-html/operations/GET-VirtualHardwareSection.html) the endpoint ` /vApp/{id}/virtualHardwareSection` is the endpoint for the information. The structs for this section already exist [here](https://github.com/vmware/go-vcloud-director/blob/master/types/v56/types.go#L1310), I would propose that a function `GetVirtualHardwareSection` be created:

```

func (vm *VM) GetVirtualHardwareSection() (*types.VirtualHardwareSection, error) {

	virtualHardwareSection := &types.VirtualHardwareSection{}

	if vm.VM.HREF == "" {
		return virtualHardwareSection, fmt.Errorf("cannot refresh, Object is empty")
	}

	_, err := vm.client.ExecuteRequest(vm.VM.HREF+"/virtualHardwareSection/", http.MethodGet,
		types.MimeVirtualHardwareSection, "error retrieving virtual hardware: %s", nil, virtualHardwareSection)

	// The request was successful
	return virtualHardwareSection, err
}

```

Along with that we should probably create a constant for the mime type as well, something like this:

```
// Mime for virtual hardware section
MimeVirtualHardwareSection = "application/vnd.vmware.vcloud.virtualHardwareSection+xml"
```

PR #198, opened 2019-06-11
Change Vdc.Status to return Int

frenchtoasters:
Returning the Vdc.Status as a string breaks the usage of the `types.VDCStatuses[]` map as the value returned is of `type string` and not `type int` as the map is defined here https://github.com/vmware/go-vcloud-director/blob/master/types/v56/types.go#37. Currently something like this is required to use the `types.VDCStatues[]` map:

```
i, err := strconv.Atoi(vdc.Vdc.Status)
if err != nil {
	return err
}
log.Info(types.VDCStatuses[i])    // Returns the mapped value       
```

This change would give `Vdc.Status` the same behavior as `Vapp.Status` when using either of their corresponding maps and allow for this to work:

```
log.Info(types.VDCStatuses[vdc.Vdc.Status])
```

## IMPORTANT

To help us process your pull request efficiently, please follow the 
guidelines shown below. 

## A Pull Request should be associated with an Issue

We wish to have discussions in Issues. A single issue may be targeted by multiple PRs.
If you're offering a new feature or fixing anything, we'd like to know beforehand in Issues,
and potentially we'll be able to point development in a particular direction.

We accept PRs without associated issues provided the change is sufficiently evident 
from the commit message. If you have typos or simple bug fixes go for it.

## Description

Related issue: (`<URL or #NUMBER of your Issue>`)

- (Required) Short description of changes in the PR subject

- (Required) Detailed description of changes include tests and
  documentation. If the pull request contains multiple commits with 
  detailed messages, refer to those instead

- (Optional) Names of reviewers using @ sign + name

PR #200, opened 2019-06-12
Add GetVirtualHardwareSection fuction 

frenchtoasters:
## IMPORTANT

guidelines shown below. 

We wish to have discussions in Issues. A single issue may be targeted by multiple PRs.
If you're offering a new feature or fixing anything, we'd like to know beforehand in Issues,
and potentially we'll be able to point development in a particular direction.

We accept PRs without associated issues provided the change is sufficiently evident 
from the commit message. If you have typos or simple bug fixes go for it.

## Description

Related issue: https://github.com/vmware/go-vcloud-director/issues/199

#### Add GetVirtualHardwareSection fuction 
This function is near identical to `GetNetworkConnectionSection()` in `/govcd/vm.go` and does essentially the same thing. It gathers the `/vApp/{id}/virtualHardwareSection` endpoint of a `VM` from the restapi and then returns that the `*types.VirtualHardwareSection` pointer which already has the corresponding structs written, [here](https://github.com/vmware/go-vcloud-director/blob/master/types/v56/types.go#L1310). It also references a new mime type `types.MimeVirtualHardwareSection` which was derived from [here](https://www.vmware.com/support/vcd/doc/rest-api-doc-1.5-html/operations/GET-VirtualHardwareSection.html).

### DIFF
--- a/govcd/vm.go
+++ b/govcd/vm.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"time"
 
 	"github.com/vmware/go-vcloud-director/v2/types/v56"
 	"github.com/vmware/go-vcloud-director/v2/util"
@@ -46,6 +47,15 @@ func (vm *VM) GetStatus() (string, error) {
 		return "", fmt.Errorf("error refreshing VM: %v", err)
 	}
 	return types.VAppStatuses[vm.VM.Status], nil
+}
+
+// IsDeployed checks if the VM is deployed or not
+func (vm *VM) IsDeployed() (bool, error) {
+	err := vm.Refresh()
+	if err != nil {
+		return false, fmt.Errorf("error refreshing VM: %v", err)
+	}
+	return vm.VM.Deployed, nil
 }
 
 func (vm *VM) Refresh() error {
@@ -91,7 +101,7 @@ func (vm *VM) GetNetworkConnectionSection() (*types.NetworkConnectionSection, er
 	networkConnectionSection := &types.NetworkConnectionSection{}
 
 	if vm.VM.HREF == "" {
-		return networkConnectionSection, fmt.Errorf("cannot refresh, Object is empty")
+		return networkConnectionSection, fmt.Errorf("cannot retrieve network when VM HREF is unset")
 	}
 
 	_, err := vm.client.ExecuteRequest(vm.VM.HREF+"/networkConnectionSection/", http.MethodGet,
@@ -99,6 +109,35 @@ func (vm *VM) GetNetworkConnectionSection() (*types.NetworkConnectionSection, er
 
 	// The request was successful
 	return networkConnectionSection, err
+}
+
+// UpdateNetworkConnectionSection applies network configuration of types.NetworkConnectionSection for the VM
+// Runs synchronously, VM is ready for another operation after this function returns.
+func (vm *VM) UpdateNetworkConnectionSection(networks *types.NetworkConnectionSection) error {
+	if vm.VM.HREF == "" {
+		return fmt.Errorf("cannot update network connection when VM HREF is unset")
+	}
+
+	// Retrieve current network configuration so that we are not altering any other internal fields
+	updateNetwork, err := vm.GetNetworkConnectionSection()
+	if err != nil {
+		return fmt.Errorf("cannot read network section for update: %s", err)
+	}
+	updateNetwork.PrimaryNetworkConnectionIndex = networks.PrimaryNetworkConnectionIndex
+	updateNetwork.NetworkConnection = networks.NetworkConnection
+	updateNetwork.Ovf = types.XMLNamespaceOVF
+
+	task, err := vm.client.ExecuteTaskRequest(vm.VM.HREF+"/networkConnectionSection/", http.MethodPut,
+		types.MimeNetworkConnectionSection, "error updating network connection: %s", updateNetwork)
+	if err != nil {
+		return err
+	}
+	err = task.WaitTaskCompletion()
+	if err != nil {
+		return fmt.Errorf("error waiting for task completion after network update for vm %s: %s", vm.VM.Name, err)
+	}
+
+	return nil
 }
 
 func (cli *Client) FindVMByHREF(vmHREF string) (VM, error) {
@@ -121,6 +160,46 @@ func (vm *VM) PowerOn() (Task, error) {
 	return vm.client.ExecuteTaskRequest(apiEndpoint.String(), http.MethodPost,
 		"", "error powering on VM: %s", nil)
 
+}
+
+// PowerOnAndForceCustomization is a synchronous function which is equivalent to the functionality
+// one has in UI. It triggers customization which may be useful in some cases (like altering NICs)
+//
+// The VM _must_ be un-deployed for this action to actually work.
+func (vm *VM) PowerOnAndForceCustomization() error {
+	// PowerOnAndForceCustomization only works if the VM was previously un-deployed
+	vmIsDeployed, err := vm.IsDeployed()
+	if err != nil {
+		return fmt.Errorf("unable to check if VM %s is un-deployed forcing customization: %s",
+			vm.VM.Name, err)
+	}
+
+	if vmIsDeployed {
+		return fmt.Errorf("VM %s must be undeployed before forcing customization", vm.VM.Name)
+	}
+
+	apiEndpoint, _ := url.ParseRequestURI(vm.VM.HREF)
+	apiEndpoint.Path += "/action/deploy"
+
+	powerOnAndCustomize := &types.DeployVAppParams{
+		Xmlns:              types.XMLNamespaceVCloud,
+		PowerOn:            true,
+		ForceCustomization: true,
+	}
+
+	task, err := vm.client.ExecuteTaskRequest(apiEndpoint.String(), http.MethodPost,
+		"", "error powering on VM with customization: %s", powerOnAndCustomize)
+
+	if err != nil {
+		return err
+	}
+
+	err = task.WaitTaskCompletion()
+	if err != nil {
+		return fmt.Errorf("error waiting for task completion after power on with customization %s: %s", vm.VM.Name, err)
+	}
+
+	return nil
 }
 
 func (vm *VM) PowerOff() (Task, error) {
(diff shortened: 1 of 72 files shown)


Didainius:
Hi, @frenchtoasters.  Thanks for the PR.
Would you be able to pull in a test for the new function?

govcd/vm.go
```diff
@@ -67,6 +67,21 @@ func (vm *VM) Refresh() error {
 	return err
 }
 
+func (vm *VM) GetVirtualHardwareSection() (*types.VirtualHardwareSection, error) {
+
+	virtualHardwareSection := &types.VirtualHardwareSection{}
+
+	if vm.VM.HREF == "" {
+		return virtualHardwareSection, fmt.Errorf("cannot refresh, Object is empty")
```
vbauzysvmware: Please update error message to not misleading one.

govcd/vm.go
```diff
@@ -67,6 +67,21 @@ func (vm *VM) Refresh() error {
 	return err
 }
 
+func (vm *VM) GetVirtualHardwareSection() (*types.VirtualHardwareSection, error) {
```
vbauzysvmware: Please add comment for this function

types/v56/types.go
```diff
@@ -367,7 +367,7 @@ type Vdc struct {
 	ID           string `xml:"id,attr,omitempty"`
 	OperationKey string `xml:"operationKey,attr,omitempty"`
 	Name         string `xml:"name,attr"`
-	Status       string `xml:"status,attr,omitempty"`
+	Status       int    `xml:"status,attr,omitempty"`
```
vbauzysvmware: I think this one change is not for this PR :)

frenchtoasters: Ya this change is based on my other PR https://github.com/vmware/go-vcloud-director/pull/198#pullrequestreview-248605514 I can revert it to the current master cause im not sure when that one will get merged in.

frenchtoasters:
> Hi, @frenchtoasters. Thanks for the PR.
> Would you be able to pull in a test for the new function?

Ya sure I didnt see one for the `GetNetworkConnectionSection` but I can mock one up tonight for this.

types/v56/types.go
frenchtoasters: I have revert this back for this pull request.

govcd/vm_test.go
```diff
@@ -880,3 +880,30 @@ func (vcd *TestVCD) Test_VMPowerOnPowerOff(check *C) {
 	check.Assert(err, IsNil)
 	check.Assert(vmStatus, Equals, "POWERED_OFF")
 }
+
+// Test gathering VM virtual hardware items
+func (vcd *TestVCD) Test_GetVirtualHardwareSection(check *C) {
```
vbauzysvmware: Thank you for adding test. Could you improve test by creating VM with some things in VirtualHardwareSection and verify they are returned. As current tests only verify function call.

govcd/vm.go
vbauzysvmware:
```suggestion
		return virtualHardwareSection, fmt.Errorf("VM HREF is empty)
```

govcd/vm_test.go
frenchtoasters: Changed up the test a bit here and added in some `VirtualHardwareItem` data and validated that it returns the equal data.

vbauzysvmware: Thank you

lvirbalas:
@frenchtoasters as a preparation for merge, please provide a commit message following the guidelines as referenced in `CONTRIBUTING.md`:
https://chris.beams.io/posts/git-commit/

govcd/vm_test.go
```diff
@@ -880,3 +880,40 @@ func (vcd *TestVCD) Test_VMPowerOnPowerOff(check *C) {
...
+					InstanceID:          1,
+					AutomaticAllocation: true,
+					Address:             "1.1.1.1",
+					CoresPerSocket:      2,
+				},
+			},
+		},
+	}
+
+	vm, err := vcd.client.Client.FindVMByHREF(newVM.HREF)
```
dataclouder: This call cannot work.
You are calling a function that finds a VM by HREF, but the HREF is only available for entities that have been created or fetched from vCD.

The test should use an already existing VM (preferred), or create a new one, and then get its contents.

lvirbalas:
@frenchtoasters Are you planning to complete this one?

frenchtoasters:
> @frenchtoasters Are you planning to complete this one?

Yes I do just have not been able to budget out the time to correct the unit test. Though I believe the fix is fairly simple, in that I should just be using the `findFirstVapp` and `findFirstVm` functions instead of creating my own vm. 

Though I am wondering @dataclouder would it be acceptable to set the `VirtualHardwareSection` for the VM that was found myself and then validate those values are returned correctly from the `GetVirtualHardwareSection` or should I use a `SetVirtualHardwareSection` function to create them as well?

dataclouder:
> Though I am wondering @dataclouder would it be acceptable to set the `VirtualHardwareSection` for the VM that was found myself and then validate those values are returned correctly from the `GetVirtualHardwareSection` or should I use a `SetVirtualHardwareSection` function to create them as well?

Both approaches work, though I am not sure the first one could get predictable results. 
I would accept both, provided that:
a) the tests are done with real VMs, not mocks.
b) the results are independently verifiable (meaning that creating vApps/VMs with our own OVAs, we get the same results as you do)
c) the testing code has sufficient comments to explain what is the test intention and what's the expected result.

Looking forward to evaluating your improvements!

govcd/vm_test.go
```diff
@@ -880,3 +880,46 @@ func (vcd *TestVCD) Test_VMPowerOnPowerOff(check *C) {
...
+				InstanceID:          1,
+				AutomaticAllocation: true,
+				Address:             "1.1.1.1",
+				CoresPerSocket:      2,
+			},
+		},
+	}
+
+	// Preform check of virtual hardware section
+	section, err := vm.GetVirtualHardwareSection()
```
Didainius: Instead of checking  if the VM has `VirtualHardwareSection` as it was preset you could use regex'es to validate if the fields are actually as they should be. For example, if the `Address` fields contains MAC address, so you could check it like that:
```
check.Assert(field_name, Matches, "^[a-fA-F0-9:]{17}|[a-fA-F0-9]{12}$")
```

govcd/vm_test.go
```diff
@@ -880,3 +880,46 @@ func (vcd *TestVCD) Test_VMPowerOnPowerOff(check *C) {
...
+	check.Assert(err, IsNil)
+	for _, item := range section.Item {
```
Didainius: The `section.Item` can contain many types of information (see [example output](https://gist.github.com/Didainius/6a01c9154cfc582c36370efde18d0baf)).
You will spot many types of `types.VirtualHardwareItem` inside `types.VirtualHardwareSection` so before checking for specific fields you must be sure that this is the correct item you're looking in.
For example - if you're looking for a MAC address - you must be sure that you're checking it in the `types.VirtualHardwareItem` of network card as checking it in `disk` will not have a MAC set.

_Hint_ the following fields should help to identify type of `types.VirtualHardwareItem`: 
```
<rasd:ResourceSubType>VMXNET3</rasd:ResourceSubType>
<rasd:ResourceType>10</rasd:ResourceType>
```

frenchtoasters: No worries ive actually been out of the country for a while and just getting back up to speed on things. Ahh that example output is exactly what I was looking for before creating this test cause I was not sure what the defaults for these would be. Im wondering though for this check should I get super granular and check all the items or would just checking the top level items, i.e. `<ovf:System>`, `<ovf:Item>`, etc... then maybe also check for the existence of child items under those but not trying to check the specifics of the children?

Didainius: Probably no need to be too granular, but you could check for some value inside top level item (to be sure it is not skipped during un-marshalling)

frenchtoasters: Looking at the types it seems that the only field that would always be there would be `VirtualHardwareSection.Info`, as all the other values in both `VirtualHardwareSection` and `VirtualHardwareItem` are set to `omitempty`. Should one assume that those values would always be there after un-marshelling for this test? Or just simply add the check for `check.Assert(section.Info, NotNil)` knowing that they may or may not be present.

Didainius: You could check that `Info` always exists and also loop through the components and find a specific device based on some field value (ensuring that it is set). Like the network card would have types and subtypes as I mentioned:
```
<rasd:ResourceSubType>VMXNET3</rasd:ResourceSubType>
<rasd:ResourceType>10</rasd:ResourceType>
```

govcd/vm_test.go
frenchtoasters: Ahh that is a good idea, I have added that to the latest commit.

govcd/vm_test.go
frenchtoasters: Ok so assuming those my latest commit should cover this, however the check within the loop is only validating the `address` field is valid _if_ its a `VMXNET3` network adapter which im assuming will always map to a `ResourceType` of `10`. Should the check also account for other possible types of network adapters or is `ResourceType==10` the only value that _any_ network adapter would have?

Didainius: That is question for research. As far as I have seen the `ResourceType == 10` hints of a network adapter, while `ResourceSubType` hints it's type (VMXNET3, E1000). Should be safe to leave `ResourceType == 10` and then look for a MAC address.

govcd/vm_test.go
```diff
@@ -880,3 +880,43 @@ func (vcd *TestVCD) Test_VMPowerOnPowerOff(check *C) {
...
+		check.Skip("skipping test because no VM is found")
+	}
+	vm, err := vcd.client.Client.FindVMByHREF(vmType.HREF)
+	check.Assert(err, IsNil)
+
+	// Preform check of virtual hardware section
+	section, err := vm.GetVirtualHardwareSection()
+	check.Assert(err, IsNil)
+
+	// Check that section.Info is not Nil, as its the only field that may not be omitted when marshelled
```
Didainius: Typo
```suggestion
	// Check that section.Info is not Nil, as its the only field that may not be omitted when marshalled
```

frenchtoasters: Nice catch!

Didainius:
The tests seems to be passing for me. Could you merge latest `master` branch as it has conflicts now.

Didainius:
Be sure to `make fmt` and see that TravisCI is green.

govcd/vm.go
```diff
@@ -77,6 +77,22 @@ func (vm *VM) Refresh() error {
 	return err
 }
 
+// GetVirtualHardwareSection returns the virtual hardware items attached to a VM
+func (vm *VM) GetVirtualHardwareSection() (*types.VirtualHardwareSection, error) {
+
+	virtualHardwareSection := &types.VirtualHardwareSection{}
+
+	if vm.VM.HREF == "" {
+		return virtualHardwareSection, fmt.Errorf("cannot refresh, invalid reference url")
```
Didainius: Not a major thing, but we aim to return `nil` instead of empty struct because it is easier in case we check something.
Excerpt from https://github.com/vmware/go-vcloud-director/blob/master/CODING_GUIDELINES.md:
> These methods return a pointer to the entity's structure and a nil error when the search was successful, **a nil pointer and an error in every other case**.


```suggestion
		return nil, fmt.Errorf("cannot refresh, invalid reference url")
```

lvirbalas: @frenchtoasters Please make this change - from my perspective it's the last remaining nit. Thanks!!

Didainius:
Thanks! 
The test passes, looks fine. I'll let other team members look around and approve before we merge it.

Didainius:
It is now merged. Thank you @frenchtoasters !

Issue #201, opened 2019-06-12
Enhance edge gateway creation with more GatewayBackingConfig values


dataclouder:
Edge gateway creation (#130) uses only two values for `GatewayBackingConfig`: "full" and "compact".
The GUI mentions other values "Large", "Xlarge", "Quad Large" which are not in the API.

When we try to enter values other than "full" or "compact" in the creation, we get an error from vCD saying that such values are not supported.

We need to figure out in which version these values are supported and what is the minimum version for our API to handle such requests.

Our current assessment is that the rejection happens because our current API minimal  version (27) is insufficient for handling the request.

We will revisit this issue when we bump up go-vcloud-director minimal version.

vbauzysvmware:
taken from html5 behaviour:

{"name":"tetsv","description":"","configuration":{"distributedRoutingEnabled":false,"fipsModeEnabled":false,"haEnabled":false,"backwardCompatibilityMode":false,"gatewayBackingConfig":"full4","gatewayInterfaces":{"gatewayInterface":[{"name":"extnet-govcd","network":{"href":"https://bos2-vcd-terraform-vcd97-1.eng.vmware.com/cloudapi/1.0.0/externalNetworks/d42a6000-0218-46f1-b449-30f0a145fc85","id":"urn:vcloud:network:d42a6000-0218-46f1-b449-30f0a145fc85"},"interfaceType":"uplink","subnetParticipation":[{"gateway":"10.150.15.253","subnetPrefixLength":20,"useForDefaultRoute":false,"netmask":"255.255.240.0","ipAddress":"","ipRanges":{"ipRange":[]}}],"applyRateLimit":false,"inRateLimit":null,"outRateLimit":null,"useForDefaultRoute":false}]},"advancedNetworkingEnabled":true,"edgeGatewayServiceConfiguration":null}}

[...]

PR #203, opened 2019-06-14
Added vdc meta data functions

vbauzysvmware:
Ref: https://github.com/terraform-providers/terraform-provider-vcd/issues/238

Functions for handling meta data for vDC.
Also from previous research added developed function for future - `QueryOrgVdcNetworkByName`


vbauzysvmware:
According my testing: https://github.com/vmware/go-vcloud-director/issues/201 
"full4" and "x-large" values accepted by API

govcd/system.go
```diff
@@ -500,3 +500,16 @@ func QueryProviderVdcByName(vcdCli *VCDClient, name string) ([]*types.QueryResul
 
 	return results.Results.VMWProviderVdcRecord, nil
 }
+
+// QueryOrgVdcNetworkByName finds a org VDC network by name which has edge gateway as reference
+func QueryOrgVdcNetworkByName(vcdCli *VCDClient, name string) ([]*types.QueryResultOrgVdcNetworkRecordType, error) {
```
lvirbalas: As the usage of this function is not visible from this PR, I'm just curious, why do we need this new function when we have remotely similar `func (vdc *Vdc) FindVDCNetwork(network string) (OrgVDCNetwork, error)` in `vdc.go`? Another way of asking, why do we need to query as opposed to finding in VDC?

vbauzysvmware: This function was result of my other POC where I need figure out network and edge gtw relationship. Find and Query return different set of data about network. I decided to add this function for future use cause already created it, didn't want to delete.

govcd/vdc.go
```diff
@@ -560,3 +560,22 @@ func (vdc *Vdc) FindMediaImage(mediaName string) (MediaItem, error) {
...
+}
+
+// DeleteMetadata() function deletes metadata by key provided as input
+func (vdc *Vdc) DeleteMetadata(key string) (Task, error) {
+	return deleteMetadata(vdc.client, key, getAdminVdcURL(vdc.Vdc.HREF))
+}
+
+// AddMetadata() function adds metadata key, value pair provided as input to vDC.
+func (vdc *Vdc) AddMetadata(key string, value string) (Task, error) {
+	return addMetadata(vdc.client, key, value, getAdminVdcURL(vdc.Vdc.HREF))
```
lvirbalas: Is this GitHub portal or bright sun, but I can't find actual implementation of these `addMetadata` and other functions in this PR. Help :)

vbauzysvmware: This using function already existing in code base

govcd/metadata.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// DeleteMetadata() function calls private function deleteMetadata() with vapp.client and vapp.VApp.HREF
+// which deletes metadata depending on key provided as input from vApp.
+func (vapp *VApp) DeleteMetadata(key string) (Task, error) {
+	return deleteMetadata(vapp.client, key, vapp.VApp.HREF)
+}
+
+// Deletes metadata (type MetadataStringValue) from the vApp
+// TODO: Support all MetadataTypedValue types with this function
```
dataclouder: What are the other types?
What happens if we call this function with unsupported ones?

govcd/metadata_test.go
```diff
@@ -0,0 +1,181 @@
...
+	. "gopkg.in/check.v1"
+)
+
+func init() {
+	testingTags["metadata"] = "metadata_test.go"
+}
+
+func (vcd *TestVCD) Test_AddMetadataForVdc(check *C) {
+	if vcd.config.VCD.Vdc == "" {
+		check.Skip("skipping test because no vDC name is empty")
```
dataclouder:
```suggestion
		check.Skip("skipping test because VDC name is empty")
```

govcd/metadata_test.go
```diff
@@ -0,0 +1,181 @@
...
+	task, err := vcd.vdc.AddMetadata("key", "value")
+	check.Assert(err, IsNil)
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
+	check.Assert(task.Task.Status, Equals, "success")
+
+	// Check if metadata was added correctly
+	metadata, err := vcd.vdc.GetMetadata()
+	check.Assert(err, IsNil)
+	check.Assert(metadata.MetadataEntry[0].Key, Equals, "key")
```
dataclouder: Shall we check whether `metadata.MetadataEntry` is a valid array before addressing an index?

govcd/metadata_test.go
```diff
@@ -0,0 +1,181 @@
...
+	task, err = vcd.vdc.DeleteMetadata("key2")
+	check.Assert(err, IsNil)
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
+	check.Assert(task.Task.Status, Equals, "success")
+	metadata, err := vcd.vdc.GetMetadata()
+	check.Assert(err, IsNil)
+	for _, k := range metadata.MetadataEntry {
+		if k.Key == "key2" {
+			check.Errorf("metadata.MetadataEntry should not contain key: %s", k)
```
dataclouder: Instead of `k`, it should be `k.Key`

govcd/metadata_test.go
```diff
@@ -0,0 +1,181 @@
...
+	task, err := vcd.vapp.AddMetadata("key", "value")
+	check.Assert(err, IsNil)
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
+	check.Assert(task.Task.Status, Equals, "success")
+
+	// Check if metadata was added correctly
+	metadata, err := vcd.vapp.GetMetadata()
+	check.Assert(err, IsNil)
+	check.Assert(metadata.MetadataEntry[0].Key, Equals, "key")
```
dataclouder: Shall we check whether `metadata.MetadataEntry` is a valid array before addressing an index?

govcd/metadata_test.go
```diff
@@ -0,0 +1,181 @@
...
+	task, err = vcd.vapp.DeleteMetadata("key2")
+	check.Assert(err, IsNil)
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
+	check.Assert(task.Task.Status, Equals, "success")
+	metadata, err := vcd.vapp.GetMetadata()
+	check.Assert(err, IsNil)
+	for _, k := range metadata.MetadataEntry {
+		if k.Key == "key2" {
+			check.Errorf("metadata.MetadataEntry should not contain key: %s", k)
```
dataclouder: `k` -> `k.Key`

govcd/metadata_test.go
```diff
@@ -0,0 +1,181 @@
...
+	task, err := vm.AddMetadata("key", "value")
+	check.Assert(err, IsNil)
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
+	check.Assert(task.Task.Status, Equals, "success")
+
+	// Check if metadata was added correctly
+	metadata, err := vm.GetMetadata()
+	check.Assert(err, IsNil)
+	check.Assert(metadata.MetadataEntry[0].Key, Equals, "key")
```
dataclouder: Shall we check whether `metadata.MetadataEntry` is a valid array before addressing an index?

govcd/metadata_test.go
```diff
@@ -0,0 +1,181 @@
...
+	task, err = vm.DeleteMetadata("key2")
+	check.Assert(err, IsNil)
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
+	check.Assert(task.Task.Status, Equals, "success")
+	metadata, err := vm.GetMetadata()
+	check.Assert(err, IsNil)
+	for _, k := range metadata.MetadataEntry {
+		if k.Key == "key2" {
+			check.Errorf("metadata.MetadataEntry should not contain key: %s", k)
```
dataclouder: `k` -> `k.Key`

govcd/system_test.go
```diff
@@ -197,3 +197,16 @@ var INVALID_NAME = `*******************************************INVALID
 func init() {
 	testingTags["system"] = "system_test.go"
 }
+
+func (vcd *TestVCD) Test_QueryOrgVdcNetworkByName(check *C) {
```
Didainius: Should this function check these specifically new parameters which aren't available in different places? (as per discussion on function definition)

govcd/metadata.go
Didainius: I think @vbauzysvmware inherited string only support. The other types are:
image

But we'd need to alter terraform resources as well to accommodate them. For now we can't fit other types because the variable is string in general.

govcd/metadata_test.go
```diff
@@ -0,0 +1,181 @@
...
+	"fmt"
+
+	. "gopkg.in/check.v1"
+)
+
+func init() {
+	testingTags["metadata"] = "metadata_test.go"
+}
+
+func (vcd *TestVCD) Test_AddMetadataForVdc(check *C) {
```
Didainius: Could this test do a cleanup (probably using the helper `AddToCleanupList`)? Now if you run explicitly this one it will leave the `key` hang there forever if the vdc is not recycled on every test suite run.

image

govcd/metadata_test.go
```diff
@@ -0,0 +1,181 @@
...
+	metadata, err := vcd.vdc.GetMetadata()
+	check.Assert(err, IsNil)
+	for _, k := range metadata.MetadataEntry {
+		if k.Key == "key2" {
+			check.Errorf("metadata.MetadataEntry should not contain key: %s", k)
+		}
+	}
+}
+
+func (vcd *TestVCD) Test_AddMetadataOnVapp(check *C) {
```
Didainius: Could it cleanup (same as one Test_AddMetadataOnVDC)

govcd/metadata_test.go
```diff
@@ -0,0 +1,181 @@
...
+	metadata, err := vcd.vapp.GetMetadata()
+	check.Assert(err, IsNil)
+	for _, k := range metadata.MetadataEntry {
+		if k.Key == "key2" {
+			check.Errorf("metadata.MetadataEntry should not contain key: %s", k)
+		}
+	}
+}
+
+func (vcd *TestVCD) Test_AddMetadataOnVm(check *C) {
```
Didainius: Could it cleanup (same as one Test_AddMetadataOnVDC)

govcd/metadata.go
vbauzysvmware: This function existed, I just refactored I moved them to other place. Currently only string supported, so no issue here.

govcd/metadata_test.go
vbauzysvmware: Updated

govcd/metadata_test.go
vbauzysvmware: Added

govcd/metadata_test.go
vbauzysvmware: fixed

govcd/metadata_test.go
vbauzysvmware: fixed

govcd/metadata_test.go
vbauzysvmware: fixed

govcd/metadata_test.go
vbauzysvmware: fixed

govcd/metadata_test.go
vbauzysvmware: Updated

govcd/system_test.go
vbauzysvmware: Improved

govcd/metadata_test.go
vbauzysvmware: Added

govcd/metadata_test.go
vbauzysvmware: it needs separate approach, cause deletion is for vApp

govcd/metadata_test.go
vbauzysvmware: it needs separate approach, cause deletion is for vApp

Didainius: Ok. VMs are being deleted during suite run as opposed to vDC (which is quite permanent).

govcd/system_test.go
```diff
@@ -219,3 +219,17 @@ var INVALID_NAME = `*******************************************INVALID
...
```
Didainius: Running: TestVCD.Test_QueryOrgVdcNetworkByName
system_test.go:232:
    check.Assert(len(orgVdcNetwork), Not(Equals), 0)
... obtained int = 0
... expected int = 0

FAIL: system_test.go:223: TestVCD.Test_QueryOrgVdcNetworkByName

vbauzysvmware: Fixed

PR #205, opened 2019-06-19
Add load balancer server pool CRUD

Didainius:
Part of implementing edge gateway load balancers https://github.com/terraform-providers/terraform-provider-vcd/issues/223

This PR introduces load balancer server pool support in govcd.


govcd/api_vcd_test.go
```diff
@@ -331,7 +332,7 @@ func (vcd *TestVCD) infoCleanup(format string, args ...interface{}) {
 // Gets the two components of a "parent" string, as passed to AddToCleanupList
 func splitParent(parent string, separator string) (first, second, third string) {
 	strList := strings.Split(parent, separator)
-	if len(strList) < 2 && len(strList) > 3 {
+	if len(strList) < 2 || len(strList) > 3 {
```
lvirbalas: Small matter, but would be nice to have a comment above the function with example values for parent and separator and expected output. When you read this function in isolation, it makes you wonder what is it used for.

govcd/api_vcd_test.go
```diff
@@ -1,4 +1,4 @@
-// +build api functional catalog vapp gateway network org query extnetwork task vm vdc system disk ALL
+// +build api functional catalog vapp gateway network org query extnetwork task vm vdc system disk unit ALL
```
lvirbalas: Just a question, which unit tests were added and would you try running only unit tests after this addition? I wonder how the tests cope with such request when there are so many build tags in one line.

govcd/api_vcd_test.go
```diff
@@ -58,6 +58,7 @@ const (
 	TestCreateExternalNetwork     = "TestCreateExternalNetwork"
 	TestDeleteExternalNetwork     = "TestDeleteExternalNetwork"
 	Test_LBServiceMonitor         = "Test_LBServiceMonitor"
+	Test_LBServerPool             = "Test_LBServerPool"
```
lvirbalas: In the sake of consistency, please remove the underscore from both variables and values.

govcd/lbserverpool.go
```diff
@@ -0,0 +1,177 @@
...
+import (
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+)
+
+// CreateLBServerPool
+func (eGW *EdgeGateway) CreateLBServerPool(lbPoolConfig *types.LBPool) (*types.LBPool, error) {
```
lvirbalas: Being a public function would like to ask for a more narrative function comment (what it does, etc.) ;)

govcd/lbserverpool.go
```diff
@@ -0,0 +1,177 @@
...
+	splitLocation := strings.Split(location, "/")
+	lbPoolID := splitLocation[len(splitLocation)-1]
+	readPool, err := eGW.ReadLBServerPool(&types.LBPool{ID: lbPoolID})
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve lb server pool with ID (%s) after creation: %s", readPool.ID, err)
+	}
+	return readPool, nil
+}
+
+func (eGW *EdgeGateway) ReadLBServerPool(lbPoolConfig *types.LBPool) (*types.LBPool, error) {
```
lvirbalas: A function comment would be nice, esp. for other users than Terraform.

govcd/lbserverpool.go
```diff
@@ -0,0 +1,177 @@
...
+			}
+			return pool, nil
+		}
+	}
+
+	return nil, fmt.Errorf("could not find load balancer server pool (name: %s, ID: %s)",
+		lbPoolConfig.Name, lbPoolConfig.ID)
+}
+
+func (eGW *EdgeGateway) UpdateLBServerPool(lbPoolConfig *types.LBPool) (*types.LBPool, error) {
```
lvirbalas: Please add a function comment, also explain caveats (if there are any) with updating.

govcd/lbserverpool.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	readPool, err := eGW.ReadLBServerPool(&types.LBPool{ID: lbPoolConfig.ID})
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve server pool with ID (%s) after update: %s", readPool.ID, err)
+	}
+	return readPool, nil
+}
+
+func (eGW *EdgeGateway) DeleteLBServerPool(lbPoolConfig *types.LBPool) error {
```
lvirbalas: Please add a function comment.

govcd/lbserverpool_test.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+// Test_LBServerPool tests CRUD methods for load balancer server pool.
+// The following things are tested:
+// Creation of load balancer server pool
+// Read load balancer server pool by both ID and Name (server pool name must be unique in single edge gateway)
+// Update - change a single field and compare that configuration and result objects are deeply equal
+// Update - try and fail to update without mandatory field
+// Delete
+func (vcd *TestVCD) Test_LBServerPool(check *C) {
```
lvirbalas: Maybe worth mentioning prerequisites of this test (EG).

govcd/api_vcd_test.go
dataclouder: The "many build tags" are needed in `api_vcd_test.go` because this code needs to be activated with every test. If we don't do that, every tag call would become `-tags api catalog`, `-tags api vapp`, and so on.
But the `unit` tag should not be in this file, or it will trigger a vCD connection for tests that don't need it.

govcd/api_vcd_test.go
```diff
@@ -331,7 +332,7 @@ func (vcd *TestVCD) infoCleanup(format string, args ...interface{}) {
 // Gets the two components of a "parent" string, as passed to AddToCleanupList
```
dataclouder:
```suggestion
 // Gets the two (or three) components of a "parent" string, as passed to AddToCleanupList
```

govcd/api_vcd_unit_test.go
```diff
@@ -0,0 +1,77 @@
+// +build api unit ALL
```
dataclouder: As discussed in chat, this code would be better placed inside `api_vcd_test.go`

govcd/api_vcd_test.go
Didainius: Done.

govcd/lbserverpool.go
```diff
@@ -0,0 +1,177 @@
+package govcd
```
dataclouder: Missing copyright

govcd/api_vcd_test.go
Didainius: Yeah. Moved the testing function to the same file as it was a "test for test" anyway.

govcd/api_vcd_test.go
Didainius: Done.

govcd/lbserverpool.go
Didainius: Sorry. This was my error. Done.

govcd/lbserverpool.go
Didainius: Done.

govcd/lbserverpool.go
Didainius: Done.

govcd/lbserverpool.go
Didainius: Done.

govcd/lbserverpool_test.go
Didainius: Added a note about it.

govcd/lbserverpool.go
dataclouder: The comment to this function should include some info on which elements of  `lbPoolConfig` need to be filled mandatorily

govcd/lbserverpool_test.go
```diff
@@ -0,0 +1,105 @@
...
+				Weight:    2,
+				Condition: "enabled",
+			},
+		},
+	}
+
+	createdLbPool, err := edge.CreateLBServerPool(lbPoolConfig)
+	check.Assert(err, IsNil)
+	check.Assert(createdLbPool.ID, Not(IsNil))
+	check.Assert(createdLbPool.Members[0].Condition, Equals, "enabled")
```
dataclouder: Should we check whether `createdLbPool.Members` is a valid array before addressing an index?

govcd/api_vcd_test.go
Didainius: Not exactly like that, but I altered it with another request from Linas.

govcd/api_vcd_unit_test.go
Didainius: Moved.

govcd/lbserverpool.go
Didainius: Done.

govcd/lbserverpool.go
Didainius: Added.

govcd/lbserverpool_test.go
Didainius: Added.

govcd/api_vcd_test.go
lvirbalas: Thanks for updates and a great test!!

govcd/lbserverpool.go
```diff
@@ -0,0 +1,194 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+)
+
+// CreateLBServerPool creates a load balancer server pool based on mandatory fields. It is a synchronous
+// operation. It returns created object with all fields (including ID) populated or an error.
+// Name and Algorithm fields must be populated.
+func (eGW *EdgeGateway) CreateLBServerPool(lbPoolConfig *types.LBPool) (*types.LBPool, error) {
```
vbauzysvmware: Currently major our functions are held by hierarchy. So usually create belongs to parent and other functions to itself. Are we still follow that or we think it's OK loose a bit. @dataclouder @lvirbalas

govcd/lbserverpool_test.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+// Test_LBServerPool tests CRUD methods for load balancer server pool.
+// The following things are tested if prerequisite Edge Gateway exists:
+// Creation of load balancer server pool
+// Read load balancer server pool by both ID and Name (server pool name must be unique in single edge gateway)
+// Update - change a single field and compare that configuration and result objects are deeply equal
+// Update - try and fail to update without mandatory field
+// Delete
+func (vcd *TestVCD) Test_LBServerPool(check *C) {
```
vbauzysvmware: I am not fond of such unit test where all functionality tested. It more like workflow test and not function unit test. @dataclouder

govcd/lbserverpool.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve lb server pool with ID (%s) after creation: %s", lbPoolID, err)
+	}
+	return readPool, nil
+}
+
+// ReadLBServerPool is able to find the types.LBPool type by Name and/or ID.
+// If both - Name and ID are specified it performs a lookup by ID and returns an error if the specified name and found
+// name do not match.
+func (eGW *EdgeGateway) ReadLBServerPool(lbPoolConfig *types.LBPool) (*types.LBPool, error) {
```
vbauzysvmware: Why we don't use `id, name string` ? this seems overcomplicated

govcd/lbserverpool.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve server pool with ID (%s) after update: %s", lbPoolConfig.ID, err)
+	}
+	return readPool, nil
+}
+
+// DeleteLBServerPool is able to delete the types.LBPool type by Name and/or ID.
+// If both - Name and ID are specified it performs a lookup by ID and returns an error if the specified name and found
+// name do not match.
+func (eGW *EdgeGateway) DeleteLBServerPool(lbPoolConfig *types.LBPool) error {
```
vbauzysvmware: Why need this struct if only `id` is used?

govcd/lbserverpool.go
Didainius: A discussion matter then - waiting for input.
To me - this avoids construction of new object and makes less code because all the operations are done on edge gateway (all load balancer components are children for edge gateway). Once you got a edge gateway, you don't need to construct any new objects.

govcd/lbserverpool_test.go
Didainius: I could split them, however this would make total run time longer due to for situations such as Test_ReadLBServerPool, Test_UpdateLBServerPool, Test_DeleteLBServerPool I would still need to create for  each scenario.

govcd/lbserverpool.go
Didainius: Same as above. I've created these as flexible functions even for future. Having them it is very cheap and easy to create wrappers such as `DeleteLBServerPoolByID`, `DeleteLBServerPoolByName`, etc.

govcd/lbserverpool.go
lvirbalas: > usually create belongs to parent and other functions to itself. Are we still follow that or we think it's OK loose a bit. 

@vbauzysvmware I'm usually for consistency, but how consistent are we in this regard?
@dataclouder your preference?

govcd/lbserverpool.go
Didainius: I've created these as flexible functions even for future. Having them it is very cheap and easy to create wrappers with less inputs.

govcd/lbserverpool.go
vbauzysvmware: For mostly of the times we are. Though my natural habit is more like holding all functions in itself.

govcd/lbserverpool.go
vbauzysvmware: well, currently it's ugly to use: `readLBPool, err := edgeGateway.ReadLBServerPool(&types.LBPool{ID: d.Id()})`

Didainius: I can add wrappers if we agree to do so.

govcd/lbserverpool.go
lvirbalas: After further investigation, I believe @Didainius may leave it where it is, because:

a. There are many create functions that are already in their .go files:
CreateDisk
CreateOrgVDCNetwork
...

b. Previous PR already passed this, so it would require refactoring:
CreateLBServiceMonitor is in lbservicemonitor.go

c. Modern IDE like GoLand shows all functions on a struct even if they are in different files.

vbauzysvmware: So end line. Use as you like it. I will do create also not in hierarchy way.

govcd/lbserverpool.go
Didainius: Added for both - Server Pool and Service Monitor.

govcd/lbserverpool.go
Didainius: Added for both - Server Pool and Service Monitor.

Issue #204, opened 2019-06-15
Can't create disks > 2GiB


elFarto:
https://github.com/vmware/go-vcloud-director/blob/07823bedaca554587356239bfab4f4e4bb2ed072/types/v56/types.go#L2306
The Size field of the Disk struct is currently an int, rather than the int64 the VCloud XSD says it should be, this limits the maximum size of disk you can create to 2GiB.

frenchtoasters:
> https://github.com/vmware/go-vcloud-director/blob/07823bedaca554587356239bfab4f4e4bb2ed072/types/v56/types.go#L2306
> 
> 
> The Size field of the Disk struct is currently an int, rather than the int64 the VCloud XSD says it should be, this limits the maximum size of disk you can create to 2GiB.

Here is a link to the documentation https://pubs.vmware.com/vca/topic/com.vmware.vcloud.api.reference.doc_56/doc/types/DiskType.html

dataclouder:
Did you actually try creating a disk of more than 2GB?
If yes, what was the error?

According to the documentation the `Size` field should indicate "Size of the disk in MB", thus allowing a theoretical size of about 2 million GB.

elFarto:
Ah..I see my issue, I had assumed that field was in bytes as I had some example code (that for the life of me I can't remember where it came from) that had the size in bytes. I didn't try to allocate a disk bigger than that since I thought it was in bytes and 2GiB+ wouldn't fit in there. So, while that field should be a int64, I think a 2.2PiB maximum should be enough for a while.

elFarto:
I got side tracked with another issue and only recently got back to testing this. It turns out the Size parameter is really in bytes.

The documentation linked above is for vCloud Air, not vCloud Director, here's the actual documentation: https://code.vmware.com/apis/442/vcloud#/doc/doc/types/DiskType.html

I have actually tested it this time, requesting that a disk 10,000,000 bytes big, and that's exactly what showed up in the vcloud interface.

dataclouder:
Thanks for testing it. I was fooled by the wrong documentation link provided above. My bad.

I will look into it.

dataclouder:
Fixed

PR #207, opened 2019-06-27
Fix Issue #204 "Can't create disks > 2GiB"

dataclouder:
Change type of size field in definition and usage

As reported in Issue #204, the `Size` field should have been an `int64`, not `int`.

### DIFF
--- a/govcd/api_vcd_test.go
+++ b/govcd/api_vcd_test.go
@@ -106,8 +106,8 @@ type TestConfig struct {
 		ExternalNetworkPortGroupType string `yaml:"externalNetworkPortGroupType,omitempty"`
 		VimServer                    string `yaml:"vimServer,omitempty"`
 		Disk                         struct {
-			Size          int `yaml:"size,omitempty"`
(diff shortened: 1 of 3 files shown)


dataclouder:
> Seems pretty clear here. Now I wonder if we need to fix it on the provider side.
> 
> ~One place I suspect is here (size cast to int): https://github.com/terraform-providers/terraform-provider-vcd/blob/ee539e86d1ebbf0edfc4896e24f05368298e60d5/vcd/resource_vcd_independent_disk.go#L94~
> 
> On the second looks this is already after converting to MB so shouldn't be an issue after all.

We should change to int64 after merging this one, as the code probably won't compile.
And even if it compiles, we would hit the limitation again. What happens is we take the size in MB (say 8000) and multiply for 1 MB size and we get 8,388,608,000, way beyond the `int` limit.

lvirbalas:
@dataclouder we'll be looking forward for a Terraform PR then. Thanks!

PR #209, opened 2019-07-02
VDC create fix

vbauzysvmware:
Additional fix for VDC create. Allows empty and zero values for VDC create.


vbauzysvmware:
Guys if you ok to add a few minutes to runing test suite time overall time consuming, I can add tests which test these few parameters with creating and destroying few VDCs.

lvirbalas:
I'm OK for one additional VDC test, as VDC creation is quite fast.

PR #210, opened 2019-07-03
Add CRUD operations for Org user

dataclouder:
* Update AdminOrg structure to include roles and users
* Define the new type User
* Define auxiliary types RoleReference, RightReference
* Add AdminOrg methods GetRole, GetUserByNameOrId
* Add AdminOrg methods CreateUser, SimpleCreateUser
* Add OrgUser methods GetRoleName, Update, Enable, Disable
* Add OrgUser methods UnconditionalDelete, SafeDelete
* Add OrgUser methods ChangePassword, Unlock, TakeOwnership
* Add tests for org user retrieval, creation, update, deletion


govcd/user.go
```diff
@@ -0,0 +1,463 @@
...
+		User:     new(types.User),
+		client:   cli,
+		adminOrg: org,
+	}
+}
+
+// GetUserByNameOrId retrieves an user within an admin organization
+// by either name or ID
+// Returns a valid user if it exists. If it doesn't, returns nil and ErrorEntityNotFound
+func (adminOrg *AdminOrg) GetUserByNameOrId(identifier string, willRefresh bool) (*OrgUser, error) {
```
lvirbalas: How does ID look like in this case? Is there a corner case where someone may be using a name which conflicts would conflict with some other user's ID?

govcd/user.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+
+	return &types.Reference{}, ErrorEntityNotFound
+}
+
+// CreateUser creates an OrgUser from a full configuration structure
+// The timeOut variable is the maximum time we wait for the user to be ready
+// (This operation does not return a task)
+// Note that the timeout is not absolute, meaning that we only wait the
+// full amount of the timeout if the user has not been created until then.
```
lvirbalas:
```
// Note that the timeout is not absolute, meaning that we only wait the
// full amount of the timeout if the user has not been created until then.
```
This sentence is duplication and not really needed - it's clear from the one below.

govcd/user.go
```diff
@@ -0,0 +1,463 @@
...
+// This function returns as soon as the user has been created, which could be as
+// little as 50ms or as much as TimeOut - 50 ms
```
lvirbalas: "as little as 50ms or as much as TimeOut - 50 ms" - please rephrase, not clear why "50" is used twice.

govcd/user.go
```diff
@@ -0,0 +1,463 @@
...
+	user := NewUser(adminOrg.client, adminOrg)
+
+	_, err = adminOrg.client.ExecuteRequest(userCreateHREF.String(), http.MethodPost,
+		types.MimeAdminUser, "error creating user: %s", userConfiguration, user.User)
+	if err != nil {
+		return nil, err
+	}
+
+	// Uncomment this line to show the user structure after creation
+	// ShowUser(*user.OrgUser)
```
lvirbalas: In general, I'm not keen in leaving such comments for debugging. Either we remove or we add ability to debug at runtime, which would be useful if a user hits something that we need to troubleshoot. What about using some of the debug flags we have already instead?

govcd/user.go
```diff
@@ -0,0 +1,463 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Attempting to retrieve the user
+	// Since there is no task to wait for, we try to retrieve using a max of 200 attempts
+	// at intervals of 50 Ms
+	// Experience shows that it usually takes less than one second to get the result
+	var delayPerAttempt time.Duration = 50
```
lvirbalas: 50ms sounds pretty often, have you checked - could you check if we even get replies that fast?

govcd/user.go
```diff
@@ -0,0 +1,463 @@
...
+func (adminOrg *AdminOrg) GetUserByNameOrId(identifier string, willRefresh bool) (*OrgUser, error) {
+	if willRefresh {
+		err := adminOrg.Refresh()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	for _, orgUser := range adminOrg.AdminOrg.Users.User {
+		if (orgUser.Name == identifier || orgUser.ID == identifier) && orgUser.Type == types.MimeAdminUser {
```
Didainius: Wondering if user type could be identified from something else, not a Mime type.

govcd/user.go
```diff
@@ -0,0 +1,463 @@
...
+func (adminOrg *AdminOrg) GetRole(roleName string) (*types.Reference, error) {
+
+	// There is no need to refresh the AdminOrg, until we implement CRUD for roles
+	for _, role := range adminOrg.AdminOrg.RoleReferences.RoleReference {
+		if role.Name == roleName {
+			return role, nil
+		}
+	}
+
+	return &types.Reference{}, ErrorEntityNotFound
```
Didainius: The `GetUserByNameOrId` returns `return nil, ErrorEntityNotFound` while this returns *empty object and `ErrorEntityNotFound`

govcd/user.go
```diff
@@ -0,0 +1,463 @@
...
+// CreateUser creates an OrgUser from a full configuration structure
+// The timeOut variable is the maximum time we wait for the user to be ready
+// (This operation does not return a task)
+// Note that the timeout is not absolute, meaning that we only wait the
+// full amount of the timeout if the user has not been created until then.
+// This function returns as soon as the user has been created, which could be as
+// little as 50ms or as much as TimeOut - 50 ms
+// Mandatory fields are: Name, Role, Password.
+// https://code.vmware.com/apis/442/vcloud-director#/doc/doc/operations/POST-CreateUser.html
+func (adminOrg *AdminOrg) CreateUser(userConfiguration *types.User, timeOut time.Duration) (*OrgUser, error) {
```
Didainius: I'm thinking if we could use the `maxRetryTimeout` we already have to be consistent.

govcd/user.go
dataclouder: The ID looks like `urn:vcloud:user:765a94b2-414b-4981-aff1-439687c8dd48`
and it can't conflict with a name because it can't be used as a name (colons are not allowed)

govcd/user.go
```diff
@@ -0,0 +1,463 @@
...
+			return nil, err
+		}
+	}
+
+	// Attempting to retrieve the user
+	// Since there is no task to wait for, we try to retrieve using a max of 200 attempts
+	// at intervals of 50 Ms
+	// Experience shows that it usually takes less than one second to get the result
+	var delayPerAttempt time.Duration = 50
+	maxAttempts := 200
```
Didainius: Could we also somehow tie this to `maxRetryTimeout`. This could avoid some "magic number"

govcd/user.go
dataclouder: The first 50ms is the minimum. The second one is used as a value to subtract from the maximum.

If timeout is 10 seconds, a successful operation can take from 50 ms (the minimum) to 9 seconds and 950 ms (= 10 seconds - 50 ms). It would then fail at 10 seconds.

govcd/user.go
```diff
@@ -0,0 +1,463 @@
...
+		return nil, fmt.Errorf("failure to retrieve a new user after %s : %s", elapsed, err)
+	}
+
+	return newUser, nil
+}
+
+// SimpleCreateUser creates an org user from a simplified structure
+func (adminOrg *AdminOrg) SimpleCreateUser(userData OrgUserConfiguration) (*OrgUser, error) {
+
+	if userData.Name == "" {
```
Didainius: Not fully sure, but maybe this could rely on underlying function `validateUserForCreation` which is called during the wrapped call `CreateUser`. That way all `create` validation would be in one place.

govcd/user.go
dataclouder: 50 ms is not a reply time. It's how long we wait before attempting the next retrieval

govcd/user.go
```diff
@@ -0,0 +1,463 @@
...
+
+	return user.Update()
+}
+
+// TakeOwnership takes ownership of the user's objects.
+// Ownership is transferred to the caller.
+// This is a call to make before deleting. Calling user.SafeDelete() will
+// run TakeOwnership before the actual user removal.
+// API Documentation: https://code.vmware.com/apis/442/vcloud-director#/doc/doc/operations/POST-TakeOwnership.html
+func (user *OrgUser) TakeOwnership() error {
```
Didainius: Nice approach. I'm just wondering - is there any risk of "canibalizing" the objects? I mean if I'm running as a `sysadmin` on some org and then while deleting it sysadmin takes ownership. Does it then mean that no one in that specific org is able to perform actions on it except the `sysadmin`?

govcd/user.go
```diff
@@ -0,0 +1,463 @@
...
+
+	return user.client.ExecuteRequestWithoutResponse(userHREF.String(), http.MethodPost,
+		types.MimeAdminUser, "error taking ownership from user : %s", nil)
+}
+
+// validateUserForInput makes sure that the minimum data
+// needed for creating an org user has been included in the configuration
+func validateUserForCreation(user *types.User) error {
+	var missingField = "missing field %s"
+	if user.Xmlns == "" {
```
Didainius: I'm thinking if `CreateUser` could implicitly set this field. It's not going to change and not user dependent anyway.
Probably it would be the same for `type` or are there any chances these could differ?

govcd/user.go
lvirbalas: Ah! Heh, had no clue that this hyphen represents a minus ;) Please change "-" to "minus" and "50 ms" to "50ms" OR just drop the last "- 50 ms" part.

govcd/user.go
lvirbalas: OK, still feels too often/DOSy. What about 200ms as in vapp.go?

govcd/user.go
dataclouder: The approach is crude, but we don't have much choice.
If we need to delete a user, we can :
a) take ownership of the objects and then deal with privileges by reassigning ownership appropriately;
b) stop any running vApp/VM belonging to the user and then delete the user. This will also delete all vApps.

I assume that deleting a user is always a drastic decision, and there are going to be consequences. Scenario "a" seems the one that gives time to handle things nicely, while scenario "b" seems appropriate in case of a security threat.

govcd/user.go
lvirbalas: Great idea, +1!

govcd/user.go
dataclouder: In fact, I could also skip the user type comparison. Such comparison is useful when we loop through `Link` type, where every item could be something different. But in this case we are looping through an explicit list of users. I can drop the comparison/

govcd/user.go
dataclouder: Well spotted. I will change it

govcd/user.go
dataclouder: I forgot about that. Somehow I was convinced that we had such timeout only for terraform-provider-vcd.
Will reuse the existing one

govcd/user.go
dataclouder: `validateUserForCreation` uses a different structure. Here we're dealing with the simplified one.

govcd/user.go
dataclouder: it actually could.

govcd/user.go
dataclouder: removed, because of simplified approach

govcd/user.go
dataclouder: Removed because simplified approach

govcd/user.go
dataclouder: removed

govcd/user.go
dataclouder: Changed, and used Client.MaxRetryTimeout as suggested by Dainius, instead of a function parameter.

govcd/user.go
dataclouder: done

govcd/user.go
```diff
@@ -0,0 +1,456 @@
...
+		User:     new(types.User),
+		client:   cli,
+		AdminOrg: org,
+	}
+}
+
+// GetUserByNameOrId retrieves an user within an admin organization
+// by either name or ID
+// Returns a valid user if it exists. If it doesn't, returns nil and ErrorEntityNotFound
+func (adminOrg *AdminOrg) GetUserByNameOrId(identifier string, willRefresh bool) (*OrgUser, error) {
```
vbauzysvmware: Seems here we explicitly state that `identifier` can hold name of ID. Or maybe have seperata method GetUserByNameOrId(identifier string, identifier name, refresh bool).. O wrappers - `GetUserByName, GetUserById`

govcd/user.go
```diff
@@ -0,0 +1,456 @@
...
+}
+
+// CreateUser creates an OrgUser from a full configuration structure
+// The timeOut variable is the maximum time we wait for the user to be ready
+// (This operation does not return a task)
+// This function returns as soon as the user has been created, which could be as
+// little as 200ms or as much as Client.MaxRetryTimeout
+// Mandatory fields are: Name, Role, Password.
+// https://code.vmware.com/apis/442/vcloud-director#/doc/doc/operations/POST-CreateUser.html
+func (adminOrg *AdminOrg) CreateUser(userConfiguration *types.User) (*OrgUser, error) {
```
vbauzysvmware: I think we need to check that user is sysAdmin

govcd/user.go
```diff
@@ -0,0 +1,456 @@
...
+
+	user := NewUser(adminOrg.client, adminOrg)
+
+	_, err = adminOrg.client.ExecuteRequest(userCreateHREF.String(), http.MethodPost,
+		types.MimeAdminUser, "error creating user: %s", userConfiguration, user.User)
+	if err != nil {
+		return nil, err
+	}
+
+	// If there is a valid task, we try to follow through
```
vbauzysvmware: When we do get Task and when don't. Please add to comment

govcd/user.go
```diff
@@ -0,0 +1,456 @@
...
+		task := NewTask(adminOrg.client)
+		task.Task = user.User.Tasks.Task[0]
+		err = task.WaitTaskCompletion()
+
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Attempting to retrieve the user
```
vbauzysvmware: I recommend to move all this verifying stuff to separate method

govcd/user.go
```diff
@@ -0,0 +1,456 @@
...
+	// and the time it occurred to get to this point, so that they may try with a longer time
+	if err != nil {
+		return nil, fmt.Errorf("failure to retrieve a new user after %s : %s", elapsed, err)
+	}
+
+	return newUser, nil
+}
+
+// SimpleCreateUser creates an org user from a simplified structure
+func (adminOrg *AdminOrg) SimpleCreateUser(userData OrgUserConfiguration) (*OrgUser, error) {
```
vbauzysvmware: I would recommend to inline with CRUD naming. CreateSimpleUser(userData SimpleOrgUserConfiguration)

govcd/user.go
```diff
@@ -0,0 +1,456 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	userHREF, err := url.ParseRequestURI(user.User.Href)
+	if err != nil {
+		return fmt.Errorf("error getting HREF for user %s : %v", user.User.Name, err)
+	}
+	util.Logger.Printf("[TRACE] Url for deleting user : %#v and name: %s", userHREF, user.User.Name)
```
vbauzysvmware: Log please ownership flag

govcd/user.go
```diff
@@ -0,0 +1,456 @@
...
+		types.MimeAdminUser, "error deleting user : %s", nil)
+}
+
+// UnconditionalDelete deletes the user, WITHOUT running a call to take ownership of the user objects
+// This call will fail if the user has owns *running* vApps/VMs
+func (user *OrgUser) UnconditionalDelete() error {
+	return user.delete(false)
+}
+
+// SafeDelete deletes the user after taking ownership of its objects
```
vbauzysvmware: More elaborated comment needed -> that runing code user will take ownership, not deleted one or somehow else

govcd/user.go
vbauzysvmware: What todo if you want delete User with all dependencies?

govcd/user_test.go
```diff
@@ -0,0 +1,185 @@
...
+		check.Assert(err, IsNil)
+		user, err = adminOrg.GetUserByNameOrId(user.User.ID, true)
+		check.Assert(err, NotNil)
+		// Tests both the error directly and the function IsNotFound
+		check.Assert(err, Equals, ErrorEntityNotFound)
+		check.Assert(IsNotFound(err), Equals, true)
+		// Expect a null pointer when user is not founc
+		check.Assert(user, IsNil)
+	}
+}
```
vbauzysvmware: As far as I guess it's hard to write test for others API methods like `TakeOwnership` ? or it's just not yet done?

govcd/api_vcd_test.go
```diff
@@ -545,6 +545,28 @@ func (vcd *TestVCD) removeLeftoverEntities(entity CleanupEntity) {
 			vcd.infoCleanup(notDeletedMsg, entity.EntityType, entity.Name, err)
 		}
 		return
+	case "user":
```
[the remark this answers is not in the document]
Didainius: I was also thinking how to make it smaller, but not sure if it's with this PR or do we make something separately to cleanup (if we ever will). A lot of it could be shrunken down based on the lookups.

One idea I had for load balancer components would be to rewrite it similarly:
```go
case "lbServiceMonitor", "lbServerPool", "other_component_having_same_requirements":  // share the object lookup part
 		if entity.Parent == "" {
			vcd.infoCleanup("removeLeftoverEntries: [ERROR] No parent specified '%s'\n", entity.Name)
			return
		}

    // these objects should be defined at function scope, rather than `switch` scope
		orgName, vdcName, edgeName := splitParent(entity.Parent, "|")

		org, vdc, edge, err = getOrgVdcEdgeByNames(vcd, orgName, vdcName, edgeName)
		if err != nil {
			vcd.infoCleanup("removeLeftoverEntries: [ERROR] %s \n", err)
		}
    fallthrough

 
case "lbServerPool":
		err = edge.DeleteLBServerPoolByName(entity.Name)
		if err != nil {
			vcd.infoCleanup(notFoundMsg, entity.EntityType, entity.Name)
			return
		}

		vcd.infoCleanup(removedMsg, entity.EntityType, entity.Name, entity.CreatedBy)
		return

case "lbServiceMonitor":
		err = edge.DeleteLBServiceMonitorByName(entity.Name)
		if err != nil {
			vcd.infoCleanup(notFoundMsg, entity.EntityType, entity.Name)
			return
		}

		vcd.infoCleanup(removedMsg, entity.EntityType, entity.Name, entity.CreatedBy)
		return
case "other_component_having_same_requirements":
		err = edge.DeleteLBOtherComponentByName(entity.Name)
		if err != nil {
			vcd.infoCleanup(notFoundMsg, entity.EntityType, entity.Name)
			return
		}

		vcd.infoCleanup(removedMsg, entity.EntityType, entity.Name, entity.CreatedBy)
		return

....
```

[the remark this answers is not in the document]
vbauzysvmware: Let's move it separate issue?

govcd/user.go
dataclouder: you can do it only if the dependencies are not running vApps/VMs. If they are stopped, and you don't run take ownership first, the dependencies will be deleted together with the user

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Added load balancer server pool [#205](https://github.com/vmware/go-vcloud-director/pull/205)
 * Added functions for refreshing, getting and update Org VDC [#206](https://github.com/vmware/go-vcloud-director/pull/206)
 * Added VDC meta data create/get/delete functions [#203](https://github.com/vmware/go-vcloud-director/pull/203)
+* Added org user create/delete/update functions [#18](https://github.com/vmware/go-vcloud-director/issues/18)
```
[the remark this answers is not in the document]
dataclouder: We don't handle roles beyond the very minimum that is needed to set up a user. To handle roles properly, we should implement:
1. define the full role structure (so far we only use a reference)
2. define the rights structure (we know that the AdminOrg refers to them but we don't use them
3. retrieve full roles
4. Create/ delete/modify roles using the individual rights.

What we have so far allows us to use roles that are already defined in the Org.

govcd/api_vcd_test.go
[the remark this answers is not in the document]
dataclouder: I think this change requires its own PR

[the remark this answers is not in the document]
Didainius: Agreed on my side as well.

govcd/user.go
```diff
@@ -0,0 +1,456 @@
...
+		client:   cli,
+		AdminOrg: org,
+	}
+}
+
+// GetUserByNameOrId retrieves an user within an admin organization
+// by either name or ID
+// Returns a valid user if it exists. If it doesn't, returns nil and ErrorEntityNotFound
+func (adminOrg *AdminOrg) GetUserByNameOrId(identifier string, willRefresh bool) (*OrgUser, error) {
+	if willRefresh {
```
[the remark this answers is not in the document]
dataclouder: Like other functions where we search the parent to find the entity, we need to refresh the entity to make sure we have a fresh image. However, since the refresh is costly, if we know that we want to get what is already there (such as when we want to get all the current users, we don't need to refresh every time) we set `willRefhesh` to `false`.

govcd/user.go
dataclouder: As I mentioned above, it's unlikely that we can mistake an ID as a name, and for OrgUsers it is impossible, as names can't contain some of the characters we find in IDs.

I don't see the risk of this approach: even if it were possible to use a name that looks like `urn:vcloud:user:440a94b2-414b-4923-aff7-439603c8dd48`, what's the risk, since we check both fields and return a match if they are the same?

govcd/user_test.go
dataclouder: It is time consuming, as we need running vApps to prove that it does what we assume.
I proved it with manual testing so far. 
We may use some of the existing vApp tests to prove this point, but it is not straightforward.
A practical difficulty would be that we need to either create the vApps using a Org User different from the one running the tests, or assign ownership of some objects to a different user (for which we don't have the functionality in the library yet).

govcd/user.go
dataclouder: `CreateSimpleUser` sounds to me like we create a user that is less complex than a full one.
In my naming, `Simple` refers to the operation, and that's why it prepends `Create`.

govcd/user.go
dataclouder: Why? We only need this operation while creating the user. It is not reusable. I haven't seen other entities where this approach is necessary.

govcd/user.go
dataclouder: Isn't this implicit when we use an AdminOrg?

govcd/user.go
```diff
@@ -0,0 +1,456 @@
...
+type OrgUserConfiguration struct {
+	Name            string // Mandatory
+	Password        string // Mandatory
+	RoleName        string // Mandatory
+	ProviderType    string // Optional: defaults to "INTEGRATED"
+	IsEnabled       bool   // Optional: defaults to false
+	IsLocked        bool   // Only used for updates
+	DeployedVmQuota int    // Optional: 0 means "unlimited"
+	StoredVmQuota   int    // Optional: 0 means "unlimited"
+	FullName        string
```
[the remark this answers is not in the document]
dataclouder: done

govcd/user.go
dataclouder: added

govcd/user.go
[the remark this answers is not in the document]
dataclouder: added comment

govcd/user.go
dataclouder: done

govcd/user.go
dataclouder: added

govcd/user.go
dataclouder: Refactored the code using maxRetryTimeout without number of attempts

govcd/user.go
vbauzysvmware: Do we have capability to find dependent vApps/VMs ?

dataclouder: > Do we have capability to find dependent vApps/VMs ?

With our code, not starting from the user. We can follow the `Owner` structure of every vApp/VM and see to whom they belong.

govcd/user.go
lvirbalas: Fantastic, thanks!

govcd/user.go
lvirbalas: I'm also on the boat of those who expect all create operations to begin with `Create...` - it makes it very easy to look at structure of the code.

`CreateSimpleUser` to me sounds fine here, because you are creating a "simple" user, as the param is simplified, but for me `CreateUserSimple` sounds even better, because then it's clear that this is a Create for a User and implies Simple way of doing it. Another synonims for "simple" could be "light" or "plain".

govcd/user.go
lvirbalas: For the first iteration, I'd keep it simple. Just try to delete the user and shine-through any resulting API error (assuming it says that it can't be deleted because of that and that).

CHANGELOG.md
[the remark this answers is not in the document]
lvirbalas: Please note that this is lower priority than data sources, import and reads that are waiting around the corner.

govcd/user.go
dataclouder: That's what it does

govcd/user.go
dataclouder: Renamed as agreed on chat

lvirbalas: Reminder to adjust the commit message, because now it (PR description) is still using the old function names.

govcd/user.go
vbauzysvmware: I just personally don't like implicit behaviour - double meaning parameters. Especially in public API. If everyone else ok with this - I can live with it.

govcd/user.go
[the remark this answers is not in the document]
vbauzysvmware: Sorry, I meant 'willRefresh -> 'refresh' or something else

govcd/user.go
vbauzysvmware: I guess in this case you are right.

govcd/user.go
vbauzysvmware: This complicated method read with "insignificant" thing. For me it easier to read:
```
func {
doAPIStuffCode...
doAPIStuffCode...
doAPIStuffCode..
bunchOfCode handling waiting
doAPIStuffCode..
return
}
Vs
func {
doAPIStuffCode...
doAPIStuffCode...
doAPIStuffCode..
waitUntilRead()
doAPIStuffCode..
return
}

File: govcd/user_test.go (hunk 20 above)
Comment (vbauzysvmware) [reply govcd/user_test.go]: As I expected :) So let's add TODO with explanation why we didn't done yet :) -> for someone who forgets it's complicated

File: govcd/user.go (hunk 19 above)
Comment (vbauzysvmware) [reply govcd/user.go]: nice

File: govcd/user.go (hunk 15 above)
Comment (vbauzysvmware) [reply govcd/user.go]: This comment much better, but still I missing point - from API we get task or not, when?

File: govcd/user.go (hunk 13 above)
Comment (dataclouder) [reply govcd/user.go]: It is not implicit. It does what it says. Finds "By name or by ID".
This is also on line with what I propose in https://github.com/vmware/go-vcloud-director/issues/211

File: govcd/user_test.go (hunk 20 above)
Comment (dataclouder) [reply govcd/user_test.go]: done

File: govcd/user.go (hunk 23 above)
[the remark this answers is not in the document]
Comment (dataclouder) [reply govcd/user.go]: "refresh" can be either a verb or a noun, while `willRefresh` shows that I want it to do something.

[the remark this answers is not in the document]
Comment (vbauzysvmware) [reply govcd/user.go]: @lvirbalas @Didainius what is your opinion?

File: govcd/user.go (hunk 13 above)
Comment (vbauzysvmware) [reply govcd/user.go]: I would like opinion @lvirbalas and @Didainius here

File: govcd/user.go (hunk 23 above)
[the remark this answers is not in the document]
Comment (Didainius) [reply govcd/user.go]: I can add one more  `withRefresh` :) but it won't change too much so it can be even as it is for me.

File: govcd/user.go (hunk 13 above)
Comment (Didainius) [reply govcd/user.go]: IMHO probably separate fields would be clearer here.

File: govcd/user.go (hunk 11 above)
Comment (lvirbalas) [reply govcd/user.go]: After reading this after a break, I realize that the real issue here is not with `TakeOwnership()`, it's with naming of `SafeDelete()` which calls it.

The word "Safe" here is counter-intuitive, because actually it's on the opposite of safe - it will delete the user even if there are objects belonging to it. It's better to rename `SafeDelete()` to something like:
- `DeleteForce()`
- `DeleteTakeOwnership()`

p.s. As in these suggestions, we should also use the CRUD naming - "Delete" being at the beginning of the method.

CC: @Didainius @vbauzysvmware

Comment (Didainius) [reply govcd/user.go]: Agree about beginning with *Delete*

File: govcd/user.go (hunk 23 above)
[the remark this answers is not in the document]
Comment (lvirbalas) [reply govcd/user.go]: I have a pretty strong opinion on this one. It should be just `refresh` and *we should use the same approach in other places if/when we need it*.

File: govcd/user.go (hunk 11 above)
Comment (vbauzysvmware) [reply govcd/user.go]: Me too

File: govcd/user_test.go [hunk 25]
```diff
@@ -0,0 +1,219 @@
...
+		// fmt.Printf("# retrieving role %s\n", roleName)
+		roleReference, err := adminOrg.GetRole(roleName)
+		check.Assert(err, IsNil)
+		check.Assert(roleReference, NotNil)
+		check.Assert(roleReference.Name, Equals, roleName)
+		check.Assert(roleReference.HREF, Not(Equals), "")
+	}
+}
+
+// Checks that we can retrieve an user by name or ID
```
lvirbalas: Grammar: "an user" -> "a user" (repeating in other places as well)

govcd/user.go
```diff
@@ -0,0 +1,505 @@
...
+	return &OrgUser{
+		User:     new(types.User),
+		client:   cli,
+		AdminOrg: org,
+	}
+}
+
+// Private function to perform the user search by either name or ID
+// It will only look for one of the two fields, depending on the value of byName
+func (adminOrg *AdminOrg) fetchUserByNameOrId(identifier string, byName, refresh bool) (*OrgUser, error) {
```
lvirbalas: Please split this function into two separate ones as it would aid code readability a lot. Now places like these require comments:

```
        // First look by ID
	orgUser, err := adminOrg.fetchUserByNameOrId(identifier, false, true)
	// if it fails, look by name
	if IsNotFound(err) {
		orgUser, err = adminOrg.fetchUserByNameOrId(identifier, true, false)
	}
```

But would become self-explanatory:

```
        orgUser, err := adminOrg.fetchUserByName(identifier, true)
        if IsNotFound(err) {
            orgUser, err = adminOrg.fetchUserById(identifier, false)
        }
```

Even these new functions would be easier to read as would avoid error-prone ifs like these:

```
		if !byName && orgUser.ID == identifier {
			userHREF = orgUser.HREF
			break
		}
		if byName && orgUser.Name == identifier {
			userHREF = orgUser.HREF
			break
		}
```

Even though it would result in a bit of code duplication, it's really tiny compared to the readability improvement.

govcd/user.go
```diff
@@ -0,0 +1,505 @@
...
+	return adminOrg.fetchUserByNameOrId(name, true, true)
+}
+
+// FetchUserById retrieves an user within an admin organization by ID
+// Returns a valid user if it exists. If it doesn't, returns nil and ErrorEntityNotFound
+// If argument refresh is true, the AdminOrg will be refreshed before searching.
+// This is usually done after creating, modifying, or deleting users.
+// If it is false, it will search within the data already in memory (useful when
+// looping through the users and we know that no changes have occurred in the meantime)
+func (adminOrg *AdminOrg) FetchUserById(ID string, refresh bool) (*OrgUser, error) {
```
lvirbalas: Nit/naming convention as this lays a precedence for the new era of fetch functions (#211): `id string` instead of `ID string`

dataclouder: done

govcd/user.go
dataclouder: done

govcd/user_test.go
dataclouder: done

govcd/user.go
lvirbalas: Thanks, it's nice how you solved it with the new `FetchUserByHref(href string)` function - it may prove useful for other use cases as well!

PR #214, opened 2019-07-09
Add functions to allow create SNAT and DNAT rules for ext netw on edge gtw

vbauzysvmware:
Ref: https://github.com/terraform-providers/terraform-provider-vcd/issues/225

New function added:
AddSNATRule
AddDNATRule
AddNATRuleAsync

These new functions supports NAT rules creation for External netw or Org VDC netw for Edge gtw configuration. Terraform changes also ready.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * Added VDC meta data create/get/delete functions [#203](https://github.com/vmware/go-vcloud-director/pull/203)
 * Added org user create/delete/update functions [#18](https://github.com/vmware/go-vcloud-director/issues/18)
 * Added load balancer application profile [#208](https://github.com/vmware/go-vcloud-director/pull/208)
-* Added edge gateway SNAT/DNAT rules functions which supports org VDC network and external network [#225](https://github.com/terraform-providers/terraform-provider-vcd/issues/225)
+* Added edge gateway SNAT/DNAT rule functions which support org VDC network and external network [#225](https://github.com/terraform-providers/terraform-provider-vcd/issues/225)
 
 ## 2.2.0 (May 15, 2019)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Added VDC meta data create/get/delete functions [#203](https://github.com/vmware/go-vcloud-director/pull/203)
 * Added org user create/delete/update functions [#18](https://github.com/vmware/go-vcloud-director/issues/18)
 * Added load balancer application profile [#208](https://github.com/vmware/go-vcloud-director/pull/208)
+* Added edge gateway SNAT/DNAT rules functions which supports org VDC network and external network [#225](https://github.com/terraform-providers/terraform-provider-vcd/issues/225)
 
 ## 2.2.0 (May 15, 2019)
 
--- a/govcd/edgegateway.go
+++ b/govcd/edgegateway.go
@@ -186,6 +186,121 @@ func (eGW *EdgeGateway) RemoveNATPortMapping(natType, externalIP, externalPort s
 
 }
 
+// AddDNATRule creates firewall DNAT rule and return refreshed existing NAT rules array or error
+func (eGW *EdgeGateway) AddDNATRule(networkHref, externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType, description string) ([]*types.NatRule, error) {
+
+	task, err := eGW.AddNATRuleAsync(networkHref, "DNAT", externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType, description)
+	if err != nil {
+		return nil, fmt.Errorf("error creating DNAT rule: %#v", err)
+	}
+	err = task.WaitTaskCompletion()
+	if err != nil {
+		return nil, fmt.Errorf("%s", combinedTaskErrorMessage(task.Task, err))
+	}
+
+	err = eGW.Refresh()
+	if err != nil {
+		return nil, fmt.Errorf("error refreshing edge gateway: %#v", err)
+	}
+
+	return eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule, nil
+}
+
+// AddSNATRule creates firewall SNAT rule and return refreshed existing NAT rules array or error
+func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, description string) ([]*types.NatRule, error) {
+
+	task, err := eGW.AddNATRuleAsync(networkHref, "SNAT", externalIP, "any", internalIP, "any", "any", "", description)
+	if err != nil {
+		return nil, fmt.Errorf("error creating SNAT rule: %#v", err)
+	}
+	err = task.WaitTaskCompletion()
+	if err != nil {
+		return nil, fmt.Errorf("%s", combinedTaskErrorMessage(task.Task, err))
+	}
+
+	err = eGW.Refresh()
+	if err != nil {
+		return nil, fmt.Errorf("error refreshing edge gateway: %#v", err)
+	}
+
+	return eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule, nil
+}
+
+// AddNATRuleAsync creates firewall NAT rule and return task or err
+func (eGW *EdgeGateway) AddNATRuleAsync(networkHref, natType, externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType, description string) (Task, error) {
+	if !isValidProtocol(protocol) {
+		return Task{}, fmt.Errorf("provided protocol is not one of TCP, UDP, TCPUDP, ICMP, ANY")
+	}
+
+	if strings.ToUpper(protocol) == "ICMP" && !isValidIcmpSubType(icmpSubType) {
+		return Task{}, fmt.Errorf("provided icmp sub type is not correct")
+	}
+
+	newEdgeConfig := eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration
+
+	// Take care of the NAT service
+	newNatService := &types.NatService{}
+
+	if newEdgeConfig.NatService == nil {
+		newNatService.IsEnabled = true
+	} else {
+		newNatService.IsEnabled = newEdgeConfig.NatService.IsEnabled
+		newNatService.NatType = newEdgeConfig.NatService.NatType
+		newNatService.Policy = newEdgeConfig.NatService.Policy
+		newNatService.ExternalIP = newEdgeConfig.NatService.ExternalIP
+
+		for _, natRule := range newEdgeConfig.NatService.NatRule {
+
+			// Kludgy IF to avoid deleting DNAT rules not created by us.
+			// If matches, let's skip it and continue the loop
+			if natRule.RuleType == natType &&
+				natRule.GatewayNatRule.OriginalIP == externalIP &&
+				natRule.GatewayNatRule.OriginalPort == externalPort &&
+				natRule.GatewayNatRule.TranslatedIP == internalIP &&
+				natRule.GatewayNatRule.TranslatedPort == internalPort &&
+				natRule.GatewayNatRule.Interface.HREF == networkHref {
+				continue
+			}
+
+			newNatService.NatRule = append(newNatService.NatRule, natRule)
+		}
+	}
+
+	//add rule
+	natRule := &types.NatRule{
+		RuleType:    natType,
+		IsEnabled:   true,
+		Description: description,
+		GatewayNatRule: &types.GatewayNatRule{
+			Interface: &types.Reference{
+				HREF: networkHref,
+			},
+			OriginalIP:     externalIP,
+			OriginalPort:   externalPort,
+			TranslatedIP:   internalIP,
+			TranslatedPort: internalPort,
+			Protocol:       protocol,
+			IcmpSubType:    icmpSubType,
+		},
+	}
+	newNatService.NatRule = append(newNatService.NatRule, natRule)
+
+	newEdgeConfig.NatService = newNatService
+
+	newRules := &types.EdgeGatewayServiceConfiguration{
+		Xmlns:      types.XMLNamespaceVCloud,
+		NatService: newNatService,
+	}
+
+	apiEndpoint, _ := url.ParseRequestURI(eGW.EdgeGateway.HREF)
+	apiEndpoint.Path += "/action/configureServices"
+
+	// Return the task
+	return eGW.client.ExecuteTaskRequest(apiEndpoint.String(), http.MethodPost,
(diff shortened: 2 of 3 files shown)


govcd/edgegateway.go
```diff
@@ -186,6 +186,121 @@ func (eGW *EdgeGateway) RemoveNATPortMapping(natType, externalIP, externalPort s
...
+		return nil, fmt.Errorf("error refreshing edge gateway: %#v", err)
+	}
+
+	return eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule, nil
+}
+
+// AddNATRuleAsync creates firewall NAT rule and return task or err
+func (eGW *EdgeGateway) AddNATRuleAsync(networkHref, natType, externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType, description string) (Task, error) {
+	if !isValidProtocol(protocol) {
+		return Task{}, fmt.Errorf("provided protocol is not one of TCP, UDP, TCPUDP, ICMP, ANY")
```
Didainius:
```suggestion
		return Task{}, fmt.Errorf("provided protocol is not one of TCP, UDP, UDP, ICMP, ANY")
```

govcd/edgegateway.go
```diff
@@ -186,6 +186,121 @@ func (eGW *EdgeGateway) RemoveNATPortMapping(natType, externalIP, externalPort s
...
+	if err != nil {
+		return nil, fmt.Errorf("error refreshing edge gateway: %#v", err)
+	}
+
+	return eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule, nil
+}
+
+// AddNATRuleAsync creates firewall NAT rule and return task or err
+func (eGW *EdgeGateway) AddNATRuleAsync(networkHref, natType, externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType, description string) (Task, error) {
+	if !isValidProtocol(protocol) {
```
Didainius: Is the API not returning valid errors? Just checking as I think we left validations for API in some other places.

govcd/edgegateway_test.go
```diff
@@ -277,3 +277,157 @@ func (vcd *TestVCD) TestEdgeGateway_GetNetworks(check *C) {
...
+	check.Assert(externalNetwork.ExternalNetwork.Name, Equals, vcd.config.VCD.ExternalNetwork)
+
+	description1 := "my Dnat Description 1"
+	description2 := "my Dnatt Description 2"
+
+	natRules, err := edge.AddDNATRule(orgVdcNetwork.OrgVDCNetwork.HREF, vcd.config.VCD.ExternalIp, "1177", vcd.config.VCD.InternalIp, "77", "TCP", "", description1)
+	check.Assert(err, IsNil)
+
+	found := false
+	var rule *types.NatRule
```
Didainius: Couldn't this become a (`Fetch` or `Read`) method  as well? It should soon be useful. Although if it's more work then we can leave it for later on.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * Added VDC meta data create/get/delete functions [#203](https://github.com/vmware/go-vcloud-director/pull/203)
 * Added org user create/delete/update functions [#18](https://github.com/vmware/go-vcloud-director/issues/18)
 * Added load balancer application profile [#208](https://github.com/vmware/go-vcloud-director/pull/208)
+* Added edge gateway SNAT/DNAT rules functions which supports org VDC network and external network [#225](https://github.com/terraform-providers/terraform-provider-vcd/issues/225)
```
Didainius: I think it's this:
```suggestion
* Added edge gateway SNAT/DNAT rule functions which support org VDC network and external network [#225](https://github.com/terraform-providers/terraform-provider-vcd/issues/225)
```

govcd/edgegateway.go
```diff
@@ -186,6 +186,121 @@ func (eGW *EdgeGateway) RemoveNATPortMapping(natType, externalIP, externalPort s
...
+
+	err = eGW.Refresh()
+	if err != nil {
+		return nil, fmt.Errorf("error refreshing edge gateway: %#v", err)
+	}
+
+	return eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule, nil
+}
+
+// AddNATRuleAsync creates firewall NAT rule and return task or err
```
Didainius: I think it's simply "NAT" rule, not firewall.
```suggestion
// AddNATRuleAsync creates NAT rule and return task or err
```

govcd/edgegateway.go
```diff
@@ -186,6 +186,121 @@ func (eGW *EdgeGateway) RemoveNATPortMapping(natType, externalIP, externalPort s
...
+
+	err = eGW.Refresh()
+	if err != nil {
+		return nil, fmt.Errorf("error refreshing edge gateway: %#v", err)
+	}
+
+	return eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule, nil
+}
+
+// AddSNATRule creates firewall SNAT rule and return refreshed existing NAT rules array or error
```
Didainius:
```suggestion
// AddSNATRule creates SNAT rule and returns refreshed existing NAT rules array or error
```

govcd/edgegateway.go
dataclouder: in Dainius's suggestion, we are repeating UDP

govcd/edgegateway.go
```diff
@@ -186,6 +186,121 @@ func (eGW *EdgeGateway) RemoveNATPortMapping(natType, externalIP, externalPort s
 
 }
 
+// AddDNATRule creates firewall DNAT rule and return refreshed existing NAT rules array or error
+func (eGW *EdgeGateway) AddDNATRule(networkHref, externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType, description string) ([]*types.NatRule, error) {
```
dataclouder: I have two problems with this function:
(1) Given that the title of this PR says allow creation of rules for external networks, why is this a method of edge gateway and not the network itself?
(2) I know it's a matter for discussion, but I don't see how passing 8 arguments by position is any better than passing a dedicated structure.

vbauzysvmware:
> Please add checks to all the tests that NAT rules were actually removed. Right now they only remove, but do not check that removal happened.

Added checks and also temporary fix for removing NAT rules to work. These methods temporary - will be changed with ID handling

govcd/edgegateway.go
vbauzysvmware: TCPUDP Is supported -> it's not mistake. As far as I saw - it create two rule

govcd/edgegateway.go
vbauzysvmware: not nice: 
```
errors.errorString = &errors.errorString{s:"operation error: task did not complete successfully:  [400:BAD_REQUEST] - [ add0f867-76df-41e8-98f4-44af29cadce2 ] Cannot update edge gateway \"vb_edge2\"\n - Cannot update edge gateway \"vb_edge2\"\n - Invalid input 'vava' provided for field featureConfigs.features[16].rules.natRulesDtos[0].protocol., error code 10050 - task error: [400 - BAD_REQUEST] [ add0f867-76df-41e8-98f4-44af29cadce2 ] Cannot update edge gateway \"vb_edge2\"\n - Cannot update edge gateway \"vb_edge2\"\n - Invalid input 'vava' provided for field featureConfigs.features[16].rules.natRulesDtos[0].protocol., error code 10050"} ("operation error: task did not complete successfully:  [400:BAD_REQUEST] - [ add0f867-76df-41e8-98f4-44af29cadce2 ] Cannot update edge gateway \"vb_edge2\"\n - Cannot update edge gateway \"vb_edge2\"\n - Invalid input 'vava' provided for field featureConfigs.features[16].rules.natRulesDtos[0].protocol., error code 10050 - task error: [400 - BAD_REQUEST] [ add0f867-76df-41e8-98f4-44af29cadce2 ] Cannot update edge gateway \"vb_edge2\"\n - Cannot update edge gateway \"vb_edge2\"\n - Invalid input 'vava' provided for field featureConfigs.features[16].rules.natRulesDtos[0].protocol., error code 10050")
```

govcd/edgegateway.go
```diff
@@ -186,6 +255,121 @@ func (eGW *EdgeGateway) RemoveNATPortMapping(natType, externalIP, externalPort s
...
+	if err != nil {
+		return nil, fmt.Errorf("%s", combinedTaskErrorMessage(task.Task, err))
+	}
+
+	err = eGW.Refresh()
+	if err != nil {
+		return nil, fmt.Errorf("error refreshing edge gateway: %#v", err)
+	}
+
+	return eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule, nil
```
Didainius: Just one more check. Are we returning all of the []*types.NatRule or just the created ones?

govcd/edgegateway_test.go
vbauzysvmware: No - fetch will be using ID.

CHANGELOG.md
vbauzysvmware: Updated

govcd/edgegateway.go
vbauzysvmware: Updated

govcd/edgegateway.go
vbauzysvmware: Updated

govcd/edgegateway.go
vbauzysvmware: All NAT Rules

govcd/edgegateway.go
Didainius: Oh. This is my nightmare tonight :) Thanks

govcd/edgegateway.go
Didainius: Ok. That's a bit different than what I would expect after adding one. How do you know which one is your created one?

govcd/edgegateway.go
vbauzysvmware: (1) Explicitly state in Title and description - this Edge gtw configuration part
(2) Yes, I'm keep inline with existing ones. Cause to many args passed I will update to use Struct

govcd/edgegateway.go
vbauzysvmware: We don't know, until I create workaround to know ID's - API create and don't return ID. So we agreed on workaround where I store govcd ID in description when figure out rule and will know rule of vCD ID  and then I can update description and return ID or exact rule

vbauzysvmware: this is intermediate PR to support only external netw assigned nat rules - ID comes after this. I try to handle changes bit by bit.

Didainius: API Hurray. 
Ok.

govcd/edgegateway.go
```diff
@@ -131,11 +132,81 @@ func (eGW *EdgeGateway) AddDhcpPool(network *types.OrgVDCNetwork, dhcppool []int
 
 }
 
+// Temporary fix for existing Remove functions - until ID support
```
dataclouder: Could you add a TODO listing what should we do when the ID support comes?
And perhaps explain what "ID support" means

govcd/edgegateway.go
```diff
@@ -131,11 +132,81 @@ func (eGW *EdgeGateway) AddDhcpPool(network *types.OrgVDCNetwork, dhcppool []int
...
+func (eGW *EdgeGateway) RemoveNATMappingRule(networkHref, natType, externalIP, internalIP, port string) (Task, error) {
+	return eGW.RemoveNATPortRule(networkHref, natType, externalIP, port, internalIP, port)
+}
+
+// Temporary fix for existing Remove functions - until ID support
```
dataclouder: Could you add a TODO listing what should we do when the ID support comes?

govcd/edgegateway.go
```diff
@@ -131,11 +132,81 @@ func (eGW *EdgeGateway) AddDhcpPool(network *types.OrgVDCNetwork, dhcppool []int
...
+	apiEndpoint, _ := url.ParseRequestURI(eGW.EdgeGateway.HREF)
+	apiEndpoint.Path += "/action/configureServices"
+
+	// Return the task
+	return eGW.client.ExecuteTaskRequest(apiEndpoint.String(), http.MethodPost,
+		"application/vnd.vmware.admin.edgeGatewayServiceConfiguration+xml", "error reconfiguring Edge Gateway: %s", newRules)
+
+}
+
+func extractObjectIDfromPath(locationPath string) (string, error) {
```
dataclouder: What is the purpose of this function? Depending on the kind of path and the kind of ID that we want to extract, this function would look differently. 
Could you add a unit test to ensure that this does what we expect?

govcd/edgegateway.go
vbauzysvmware: (2) change added

govcd/edgegateway.go
vbauzysvmware: Update comments

govcd/edgegateway.go
vbauzysvmware: Update comments

govcd/edgegateway.go
vbauzysvmware: Add comment. P.s. with next PR all these function will be removed

lvirbalas: A general reminder about function naming: we agreed to use `Id` instead of `ID` (and `From` instead of `from`).

govcd/edgegateway.go
```diff
@@ -39,6 +40,19 @@ func NewEdgeGateway(cli *Client) *EdgeGateway {
 	}
 }
 
+// Struct which covers NAT rule fields
+type NatRule struct {
```
lvirbalas: Could we use the vCD API type instead of creating a new one? Suggesting to do it like with LB create functions. This avoids additional level of abstraction and all the cons coming with it.

vbauzysvmware: IMHO - no. Cause: GatewayNatRule has Interface field, which is Reference object. GatewayNatRule has Xmls field. Description, ruletype, enabled is in NatRule which is parent. Parent Natule has 5 fields which we don't need

govcd/edgegateway.go
vbauzysvmware: Tx. Won't change as it will be removed

govcd/edgegateway.go
```diff
@@ -186,6 +272,124 @@ func (eGW *EdgeGateway) RemoveNATPortMapping(natType, externalIP, externalPort s
 
 }
 
+// AddDNATRule creates firewall DNAT rule and return refreshed existing NAT rules array or error
+func (eGW *EdgeGateway) AddDNATRule(ruleDetails NatRule) ([]*types.NatRule, error) {
```
lvirbalas: Please _extend_ the comments of all new `Add...` functions to make it clear that they support assigning specific network in contrast to the old function (please give the name of the old function as well). This way we will always remember why we have different functions and it will be easy to not forget that there is also the old one.

vbauzysvmware: Added

PR #215, opened 2019-07-12
Add load balancer virtual server CRUD

Didainius:
Work continues on introducing edge gateway load balancers https://github.com/terraform-providers/terraform-provider-vcd/issues/223

**NOTE** Until a next PR being submitted one must manually enable load balancing (and advanced load balancing). This can be done in `Edge Gateway -> Load balancer -> Global configuration`. Enable both **Enabled** and **Acceleration Enabled** as it will not work otherwise until we have this configuration in govcd code (my next PR after this is merged as it has a dependency).

### DIFF
--- a/govcd/api_vcd_test.go
+++ b/govcd/api_vcd_test.go
@@ -690,9 +690,12 @@ func (vcd *TestVCD) removeLeftoverEntities(entity CleanupEntity) {
 		}
 
 		err = edge.DeleteLBServiceMonitorByName(entity.Name)
-		if err != nil {
+		if err != nil && strings.Contains(err.Error(), ErrorEntityNotFound.Error()) {
 			vcd.infoCleanup(notFoundMsg, entity.EntityType, entity.Name)
 			return
+		}
+		if err != nil {
+			vcd.infoCleanup(notDeletedMsg, entity.EntityType, entity.Name, err)
 		}
 
 		vcd.infoCleanup(removedMsg, entity.EntityType, entity.Name, entity.CreatedBy)
@@ -712,9 +715,12 @@ func (vcd *TestVCD) removeLeftoverEntities(entity CleanupEntity) {
 		}
 
 		err = edge.DeleteLBServerPoolByName(entity.Name)
-		if err != nil {
+		if err != nil && strings.Contains(err.Error(), ErrorEntityNotFound.Error()) {
 			vcd.infoCleanup(notFoundMsg, entity.EntityType, entity.Name)
 			return
+		}
+		if err != nil {
+			vcd.infoCleanup(notDeletedMsg, entity.EntityType, entity.Name, err)
 		}
 
 		vcd.infoCleanup(removedMsg, entity.EntityType, entity.Name, entity.CreatedBy)
@@ -733,9 +739,12 @@ func (vcd *TestVCD) removeLeftoverEntities(entity CleanupEntity) {
 		}
 
 		err = edge.DeleteLBAppProfileByName(entity.Name)
-		if err != nil {
+		if err != nil && strings.Contains(err.Error(), ErrorEntityNotFound.Error()) {
 			vcd.infoCleanup(notFoundMsg, entity.EntityType, entity.Name)
 			return
+		}
+		if err != nil {
+			vcd.infoCleanup(notDeletedMsg, entity.EntityType, entity.Name, err)
 		}
 
 		vcd.infoCleanup(removedMsg, entity.EntityType, entity.Name, entity.CreatedBy)
@@ -755,9 +764,12 @@ func (vcd *TestVCD) removeLeftoverEntities(entity CleanupEntity) {
 		}
 
 		err = edge.DeleteLBVirtualServerByName(entity.Name)
-		if err != nil {
+		if err != nil && strings.Contains(err.Error(), ErrorEntityNotFound.Error()) {
 			vcd.infoCleanup(notFoundMsg, entity.EntityType, entity.Name)
 			return
+		}
+		if err != nil {
+			vcd.infoCleanup(notDeletedMsg, entity.EntityType, entity.Name, err)
 		}
 
 		vcd.infoCleanup(removedMsg, entity.EntityType, entity.Name, entity.CreatedBy)
@@ -776,9 +788,12 @@ func (vcd *TestVCD) removeLeftoverEntities(entity CleanupEntity) {
 		}
 
 		err = edge.DeleteLBAppRuleByName(entity.Name)
-		if err != nil {
+		if err != nil && strings.Contains(err.Error(), ErrorEntityNotFound.Error()) {
 			vcd.infoCleanup(notFoundMsg, entity.EntityType, entity.Name)
 			return
+		}
+		if err != nil {
+			vcd.infoCleanup(notDeletedMsg, entity.EntityType, entity.Name, err)
 		}
 
 		vcd.infoCleanup(removedMsg, entity.EntityType, entity.Name, entity.CreatedBy)
--- a/govcd/api_vcd_test.go
+++ b/govcd/api_vcd_test.go
@@ -1,4 +1,4 @@
-// +build api functional catalog vapp gateway network org query extnetwork task vm vdc system disk lbAppProfile lbServerPool lbServiceMonitor user ALL
+// +build api functional catalog vapp gateway network org query extnetwork task vm vdc system disk lbAppProfile lbServerPool lbServiceMonitor lbVirtualServer user ALL
 
 /*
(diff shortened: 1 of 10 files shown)


govcd/lbvirtualserver_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+// Test_LBVirtualServer tests CRUD methods for load balancer virtual server.
+// With help of function buildTestLBVirtualServerPrereqs such prerequisite components are created:
+// service monitor, server pool, application profile and application rule.
+// The following things are tested if prerequisites are met:
+// Creation of load balancer virtual server
+// Read load balancer virtual server by both Id and Name (virtual server name must be unique in single edge gateway)
+// Update - change a single field and compare that configuration and result objects are deeply equal
+// Update - try and fail to update without mandatory field
+// Delete
```
lvirbalas: Nit: I'd add numbered list here, so it's easier to notice when scanning the code.

types/v56/types.go
```diff
@@ -1740,44 +1740,24 @@ type LBAppRule struct {
 
 type LBAppRules []LBAppRule
 
-// LoadBalancerVirtualServer represents a load balancer virtual server.
```
vbauzysvmware: Why we don't these types? didn't you add with previous PR's?

Didainius: Nope. These were old leftover from sometime ago. They aren't the same.

govcd/sample_govcd_test_config.json
```diff
@@ -60,6 +60,11 @@
   },
   "media": {
     "mediaPath": "../test-resources/test.iso",
-    "mediaName": "uploadedMediaName"
+    "mediaName": "uploadedMediaName",
+	"//": "Relative path for Photon OS image or existing image name (as it appears in the Catalog UI).",
+	"//": "the Catalog UI). This is required for load balancer integration test. If the image ",
+	"//": "already exists in your environment just set it's name here to reuse it",
```
dataclouder: "it's" -> "its"

govcd/sample_govcd_test_config.yaml
```diff
@@ -143,4 +143,7 @@ media:
   mediaPath: ../test-resources/test.iso
   # Existing media in test system
   mediaName: uploadedMediaName
-
+  # Relative path for Photon OS image or existing image name (as it appears in the Catalog UI).
+  # This is required for load balancer integration test. If the image already exists in your
+  # environment just set it's name here to reuse it (e.g. photon-hw11-3.0-26156e2.ova, photon-os)
```
dataclouder: "it's", "its"

govcd/lbvirtualserver.go
```diff
@@ -0,0 +1,230 @@
...
+		if lbVirtualServerConfig.Id != "" && virtualServer.Id == lbVirtualServerConfig.Id {
+			return virtualServer, nil
+		}
+
+		// If Name was specified for lookup - look for the same Name
+		if lbVirtualServerConfig.Name != "" && virtualServer.Name == lbVirtualServerConfig.Name {
+			// We found it by name. Let's verify if search Id was specified and it matches the lookup object
+			if lbVirtualServerConfig.Id != "" && virtualServer.Id != lbVirtualServerConfig.Id {
+				return nil, fmt.Errorf("load balancer virtual server was found by name (%s), "+
+					"but it's Id (%s) does not match specified Id (%s)",
```
dataclouder: "it's" -> "its"

govcd/lbvirtualserver_test.go
```diff
@@ -0,0 +1,182 @@
...
+	check.Assert(lbVirtualServerById.Name, Equals, lbVirtualServerByName.Name)
+
+	// Test updating fields
+	// Update algorithm
+	lbVirtualServerById.Port = 8889
+	updatedLBPool, err := edge.UpdateLBVirtualServer(lbVirtualServerById)
+	check.Assert(err, IsNil)
+	check.Assert(updatedLBPool.Port, Equals, lbVirtualServerById.Port)
+
+	// Verify that updated pool and it's configuration are identical
```
dataclouder: "it's" -> "its"

govcd/lbvirtualserver_test.go
Didainius: Added

govcd/api_vcd_test.go
```diff
@@ -128,8 +130,9 @@ type TestConfig struct {
 		OVAChunkedPath string `yaml:"ovaChunkedPath,omitempty"`
 	} `yaml:"ova"`
 	Media struct {
-		MediaPath string `yaml:"mediaPath,omitempty"`
-		Media     string `yaml:"mediaName,omitempty"`
+		MediaPath       string `yaml:"mediaPath,omitempty"`
+		Media           string `yaml:"mediaName,omitempty"`
+		PhotonOsOvaPath string `yaml:"photonOsOvaPath,omitempty"`
```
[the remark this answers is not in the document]
Didainius: Fixed

govcd/sample_govcd_test_config.yaml
Didainius: Fixed

Didainius: Fixed

govcd/lbvirtualserver.go
Didainius: Fixed

govcd/lbvirtualserver_test.go
Didainius: Fixed

govcd/lb_test.go
```diff
@@ -0,0 +1,356 @@
...
+	_, ovaFile := path.Split(ovaPath)
+	catalogItem, err := catalog.FindCatalogItem(ovaFile)
+	check.Assert(err, IsNil)
+
+	// If the image is not already uploaded - upload it
+	if catalogItem == (CatalogItem{}) || catalogItem.CatalogItem.Name == "" {
+		fmt.Printf("# OS image '%s' not found in catalog '%s. Uploading '%s'",
+			ovaFile, catalog.Catalog.Name, ovaPath)
+		// Upload OS image
+		uploadImage(ovaPath, ovaFile, catalog, vcd, check)
```
lvirbalas: Great test here. I'm only thinking whether it's this test's responsibility to upload an OVA. Maybe we should just skip it if there's no such OVA already uploaded by the test framework? @dataclouder your thoughts?

govcd/api_vcd_test.go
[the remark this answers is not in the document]
lvirbalas: Is there any way we could avoid adding one more parameter to the test config? Why does LB test explicitly need Photon? Wouldn't other OVAs with VMware Tools (setup correctly) work?

I'm thinking that either way most of us should be using a capable OVA for the tests, so I suggest we skip or fail LB test if the OVA is not capable enough as opposed to adding one more parameter.

@dataclouder @vbauzysvmware ?

govcd/lb_test.go
dataclouder: Skipping test on missing prerequisites is a good practice.
We have many tests that depend on a catalog item. We can do the same here, if the item is missing.
We should keep the separate `photonOsOva` (not `Path`, as we don't want to upload during the test) in the configuration file, as the vApp template that we use for normal tests may be different. 
IMO, an empty `photonOsOva` or a missing catalog item as suggested by this field should just skip the test without errors.

Didainius: Ok. So I change it to `photonOsOva` and do not upload it. If it's not there - I skip.

govcd/api_vcd_test.go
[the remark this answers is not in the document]
Didainius: As far as the OVA capability - inside the test there is a script which relies on guest customization, python3, specific firewall disable script and being able to set IPs properly. Not all OSes do it in exactly the same way. It would be overkill to identify OSes if an OS supports all that.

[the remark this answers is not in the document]
Didainius: Changed to catalogItem as agreed. Now it checks if catalogItem is Photon OS. If it's not - just skips the test.

govcd/lb_test.go
Didainius: As per other comment I relly on catalogItem being Photon OS now. If it's no Photon OS - I skip the test.

govcd/api_vcd_test.go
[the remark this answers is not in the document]
lvirbalas: Neat!

Didainius:
NP.
There is still an issue that the tests not always work fine under 9.7 (9.5 works fine). I'm to investigate if I can make this work for 9.7.

Didainius:
Fix/workaround added for #218

Issue #211, opened 2019-07-04
Standardize GetByName/FindbyName methods


dataclouder:
**NB: This issue is relevant to the implementation of data sources/imports in terraform-provider-vdc.**

# Problem

Currently we have several methods with the format

```go
func (parent *ParentEntity) GetEntityByName(name string) (Entity, error) {}
func (parent *ParentEntity) FindEntityByName(name string) (Entity, error) {}
```

We have three problems with these methods:
1. Non standard names. They should all be called `Find...` or all `Get...`
2. Some return an `Entity` structure, some a pointer to the entity
3. Some return an error when the entity is not found, some return nil and an empty structure when the entity was not found.
4. Not really a problem, but a limitation: we only search by name, while we will soon need to search for ID too.

Problem n.2 and n.3 contribute to complicate usage of such methods. Currently we do something like:

```go
    entity, err := parent.FindEntityByName(name)
    if err != nil {
      return err
    }
    if entity == Entity{} {
        return err
    }
    // OR something like
    if entity.Child.HREF == "" {
        return err
    }
```
The point is that we don't know for sure what the error means, and we don't know what to expect from the return value.

# Proposal
To alleviate the problems above, we need to
1. User standard names
2. Make it simpler to detect when the search failed
3. Assign meaning to the errors

## Solution 1: remove the error  and use pointers

```go
func (parent *ParentEntity) GetEntityByName(name string) *Entity {}
```
If we want the users to inspect the returned object to determine whether the search was successful, we don't need the error. We just return a pointer to the structure and the receiver function can simply say `if entity != nil` and that's it.

## Solution 2: Give meaning to the error and return pointers

```go
var ErrorEntityNotFound = fmt.Errorf("entity was not found")
// ...
func (parent *ParentEntity) GetEntityByName(name string) (*Entity, error) {
    if weDidNotFindTheEntity {
      return nil, ErrorEntityNotFound
    }
}
```
Using this method, the receiving function will do:

```go
    entity, err := parent.FindEntityByName(name)
    if err != nil {  // meaning we did not find, or there was some other error
      return someError
    }
    // OR
    if entity == nil { // also this means that we did not find
        return someError
    }
    // OR
    if err ==  ErrorEntityNotFound { // also this means that we did not find
        return someError
    }
```

# Standardization

Whichever solution we choose, we should do five things:

1. Deprecate the existing methods, but leave them in place, as they need to still work in Terraform
[...]

PR #216, opened 2019-07-15
Add NAT rule handling IDs

vbauzysvmware:
Ref: https://github.com/terraform-providers/terraform-provider-vcd/issues/244

Previous temporary functions were removed and new functions added to support ID handling. Majorly solves Nat rule removal.

### DIFF
--- a/govcd/edgegateway.go
+++ b/govcd/edgegateway.go
@@ -186,6 +186,121 @@ func (eGW *EdgeGateway) RemoveNATPortMapping(natType, externalIP, externalPort s
 
 }
 
+// AddDNATRule creates firewall DNAT rule and return refreshed existing NAT rules array or error
+func (eGW *EdgeGateway) AddDNATRule(networkHref, externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType, description string) ([]*types.NatRule, error) {
+
+	task, err := eGW.AddNATRuleAsync(networkHref, "DNAT", externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType, description)
+	if err != nil {
+		return nil, fmt.Errorf("error creating DNAT rule: %#v", err)
+	}
+	err = task.WaitTaskCompletion()
+	if err != nil {
+		return nil, fmt.Errorf("%s", combinedTaskErrorMessage(task.Task, err))
+	}
+
+	err = eGW.Refresh()
+	if err != nil {
+		return nil, fmt.Errorf("error refreshing edge gateway: %#v", err)
+	}
+
+	return eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule, nil
+}
+
+// AddSNATRule creates firewall SNAT rule and return refreshed existing NAT rules array or error
+func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, description string) ([]*types.NatRule, error) {
+
+	task, err := eGW.AddNATRuleAsync(networkHref, "SNAT", externalIP, "any", internalIP, "any", "any", "", description)
+	if err != nil {
+		return nil, fmt.Errorf("error creating SNAT rule: %#v", err)
+	}
+	err = task.WaitTaskCompletion()
+	if err != nil {
+		return nil, fmt.Errorf("%s", combinedTaskErrorMessage(task.Task, err))
+	}
+
+	err = eGW.Refresh()
+	if err != nil {
+		return nil, fmt.Errorf("error refreshing edge gateway: %#v", err)
+	}
+
+	return eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule, nil
+}
+
+// AddNATRuleAsync creates firewall NAT rule and return task or err
+func (eGW *EdgeGateway) AddNATRuleAsync(networkHref, natType, externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType, description string) (Task, error) {
+	if !isValidProtocol(protocol) {
+		return Task{}, fmt.Errorf("provided protocol is not one of TCP, UDP, TCPUDP, ICMP, ANY")
+	}
+
+	if strings.ToUpper(protocol) == "ICMP" && !isValidIcmpSubType(icmpSubType) {
+		return Task{}, fmt.Errorf("provided icmp sub type is not correct")
+	}
+
+	newEdgeConfig := eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration
+
+	// Take care of the NAT service
+	newNatService := &types.NatService{}
+
+	if newEdgeConfig.NatService == nil {
+		newNatService.IsEnabled = true
+	} else {
+		newNatService.IsEnabled = newEdgeConfig.NatService.IsEnabled
+		newNatService.NatType = newEdgeConfig.NatService.NatType
+		newNatService.Policy = newEdgeConfig.NatService.Policy
+		newNatService.ExternalIP = newEdgeConfig.NatService.ExternalIP
+
+		for _, natRule := range newEdgeConfig.NatService.NatRule {
+
+			// Kludgy IF to avoid deleting DNAT rules not created by us.
+			// If matches, let's skip it and continue the loop
+			if natRule.RuleType == natType &&
+				natRule.GatewayNatRule.OriginalIP == externalIP &&
+				natRule.GatewayNatRule.OriginalPort == externalPort &&
+				natRule.GatewayNatRule.TranslatedIP == internalIP &&
+				natRule.GatewayNatRule.TranslatedPort == internalPort &&
+				natRule.GatewayNatRule.Interface.HREF == networkHref {
+				continue
+			}
+
+			newNatService.NatRule = append(newNatService.NatRule, natRule)
+		}
+	}
+
+	//add rule
+	natRule := &types.NatRule{
+		RuleType:    natType,
+		IsEnabled:   true,
+		Description: description,
+		GatewayNatRule: &types.GatewayNatRule{
+			Interface: &types.Reference{
+				HREF: networkHref,
+			},
+			OriginalIP:     externalIP,
+			OriginalPort:   externalPort,
+			TranslatedIP:   internalIP,
+			TranslatedPort: internalPort,
+			Protocol:       protocol,
+			IcmpSubType:    icmpSubType,
+		},
+	}
+	newNatService.NatRule = append(newNatService.NatRule, natRule)
+
+	newEdgeConfig.NatService = newNatService
+
+	newRules := &types.EdgeGatewayServiceConfiguration{
+		Xmlns:      types.XMLNamespaceVCloud,
+		NatService: newNatService,
+	}
+
+	apiEndpoint, _ := url.ParseRequestURI(eGW.EdgeGateway.HREF)
+	apiEndpoint.Path += "/action/configureServices"
+
+	// Return the task
+	return eGW.client.ExecuteTaskRequest(apiEndpoint.String(), http.MethodPost,
+		"application/vnd.vmware.admin.edgeGatewayServiceConfiguration+xml", "error reconfiguring Edge Gateway: %s", newRules)
+}
+
+// Deprecated: Use eGW.AddSNATRule() or eGW.AddDNATRule()
 func (eGW *EdgeGateway) AddNATRule(network *types.OrgVDCNetwork, natType, externalIP, internalIP string) (Task, error) {
 	return eGW.AddNATPortMappingWithUplink(network, natType, externalIP, "any", internalIP, "any", "any", "")
 }
@@ -250,6 +365,7 @@ func isValidIcmpSubType(protocol string) bool {
 	return false
 }
 
+// Deprecated: Use eGW.AddNATFirewallRule()
 func (eGW *EdgeGateway) AddNATPortMappingWithUplink(network *types.OrgVDCNetwork, natType, externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType string) (Task, error) {
 	// if a network is provided take it, otherwise find first uplink on the edge gateway
 	var uplinkRef string
--- a/govcd/edgegateway.go
+++ b/govcd/edgegateway.go
@@ -200,8 +200,9 @@ func (eGW *EdgeGateway) RemoveNATPortMapping(natType, externalIP, externalPort, 
 
 }
 
-// RemoveNATRule removes NAT rule and handles task. Return errors when issue risen.
-// Old functions RemoveNATPortMapping and RemoveNATMapping removes using rule details and expects interface is external network.
+// RemoveNATRule removes NAT removes NAT rule identified by ID and handles task. Returns error if issues rise.
+// Old functions RemoveNATPortMapping and RemoveNATMapping removed using rule details
+// and expected interface to be of external network type.
 func (eGW *EdgeGateway) RemoveNATRule(id string) error {
 	task, err := eGW.RemoveNATRuleAsync(id)
 	if err != nil {
@@ -216,8 +217,8 @@ func (eGW *EdgeGateway) RemoveNATRule(id string) error {
 }
 
 // RemoveNATRuleAsync removes NAT rule and return  or error.
-// Old functions RemoveNATPortMapping and RemoveNATMapping
-// removes using rule details and expects interface is external network.
+// Old functions RemoveNATPortMapping and RemoveNATMapping removed using rule details
+// and expected interface to be of external network type.
 func (eGW *EdgeGateway) RemoveNATRuleAsync(id string) (Task, error) {
 	if "" == id {
 		return Task{}, fmt.Errorf("provided id is empty")
@@ -267,8 +268,9 @@ func (eGW *EdgeGateway) RemoveNATRuleAsync(id string) (Task, error) {
 // AddDNATRule creates DNAT rule and return created NAT struct or error.
 // Allows to assign specific Org VDC or external network.
 // When edge gateway is advanced vCD API uses element <tag> to map with NSX edge gateway ID. Currently existing issue,
-// that changed rule using UI resets <tag> as so mapping and as result fetched NatRule.ID won't be valid anymore.
-// Old functions AddNATPortMapping and AddNATMapping assigns rule only to first external network
+// that updating rule using User interface resets <tag> and as result mapping is lost.
+// Fetching using NatRule.ID won't be valid anymore.
+// Old functions AddNATPortMapping and AddNATMapping assigned rule only to first external network
 func (eGW *EdgeGateway) AddDNATRule(ruleDetails NatRule) (*types.NatRule, error) {
 	mappingId, err := getPseudoUuid()
 	if err != nil {
@@ -312,12 +314,13 @@ func (eGW *EdgeGateway) AddDNATRule(ruleDetails NatRule) (*types.NatRule, error)
 
 // AddSNATRule creates SNAT rule and returns created NAT rule or error.
 // Allows to assign specific Org VDC or external network.
-// Old function AddNATPortMapping and AddNATMapping assigns rule to only first external network
+// Old functions AddNATPortMapping and AddNATMapping aren't correct as assigned rule only to first external network
 func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, description string) (*types.NatRule, error) {
 
-	// As vCD API don't return rule API we fetch it manually:
+	// As vCD API doesn't return rule ID we fetch it manually:
 	//  * create rule with description which value is our generated Id
 	//  * find rule which has description with our generated Id
+	//  * read the real (vCD's) rule Id
 	//  * update description with real value and return nat rule
 
 	mappingId, err := getPseudoUuid()
@@ -447,7 +450,7 @@ func (eGW *EdgeGateway) FetchNatRule(id string) (*types.NatRule, error) {
 
 // AddNATRuleAsync creates NAT rule and return task or err
 // Allows to assign specific network Org VDC or external. Old function AddNATPortMapping and
-// AddNATMapping assigns rule to first external network
+// AddNATMapping function shouldn't be used cause assigns rule to first external network
 func (eGW *EdgeGateway) AddNATRuleAsync(ruleDetails NatRule) (Task, error) {
 	if !isValidProtocol(ruleDetails.Protocol) {
 		return Task{}, fmt.Errorf("provided protocol is not one of TCP, UDP, TCPUDP, ICMP, ANY")
--- a/govcd/edgegateway.go
+++ b/govcd/edgegateway.go
@@ -275,7 +275,7 @@ func (eGW *EdgeGateway) AddDNATRule(networkHref, externalIP, externalPort, inter
 	return eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule, nil
 }
 
-// AddSNATRule creates firewall SNAT rule and return refreshed existing NAT rules array or error
+// AddSNATRule creates SNAT rule and returns refreshed existing NAT rules array or error
 func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, description string) ([]*types.NatRule, error) {
 
 	task, err := eGW.AddNATRuleAsync(networkHref, "SNAT", externalIP, "any", internalIP, "any", "any", "", description)
@@ -295,7 +295,7 @@ func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, descrip
 	return eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule, nil
 }
 
-// AddNATRuleAsync creates firewall NAT rule and return task or err
+// AddNATRuleAsync creates NAT rule and return task or err
 func (eGW *EdgeGateway) AddNATRuleAsync(networkHref, natType, externalIP, externalPort, internalIP, internalPort, protocol, icmpSubType, description string) (Task, error) {
 	if !isValidProtocol(protocol) {
 		return Task{}, fmt.Errorf("provided protocol is not one of TCP, UDP, TCPUDP, ICMP, ANY")
--- a/govcd/edgegateway.go
+++ b/govcd/edgegateway.go
@@ -272,7 +272,9 @@ func (eGW *EdgeGateway) RemoveNATPortMapping(natType, externalIP, externalPort, 
 
 }
 
-// AddDNATRule creates firewall DNAT rule and return refreshed existing NAT rules array or error
+// AddDNATRule creates DNAT rule and return refreshed existing NAT rules array or error
+// Allows to assign specific network Org VDC or external. Old function AddNATPortMapping and
+// AddNATMapping assigns rule to first external network
 func (eGW *EdgeGateway) AddDNATRule(ruleDetails NatRule) ([]*types.NatRule, error) {
 
 	ruleDetails.natType = "DNAT"
@@ -294,6 +296,8 @@ func (eGW *EdgeGateway) AddDNATRule(ruleDetails NatRule) ([]*types.NatRule, erro
 }
 
 // AddSNATRule creates SNAT rule and returns refreshed existing NAT rules array or error
+// Allows to assign specific network Org VDC or external. Old function AddNATPortMapping and
+// AddNATMapping assigns rule to first external network
 func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, description string) ([]*types.NatRule, error) {
 
 	task, err := eGW.AddNATRuleAsync(NatRule{networkHref: networkHref, natType: "SNAT", externalIP: externalIP,
@@ -316,6 +320,8 @@ func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, descrip
 }
 
 // AddNATRuleAsync creates NAT rule and return task or err
+// Allows to assign specific network Org VDC or external. Old function AddNATPortMapping and
+// AddNATMapping assigns rule to first external network
 func (eGW *EdgeGateway) AddNATRuleAsync(ruleDetails NatRule) (Task, error) {
 	if !isValidProtocol(ruleDetails.protocol) {
 		return Task{}, fmt.Errorf("provided protocol is not one of TCP, UDP, TCPUDP, ICMP, ANY")
--- a/govcd/edgegateway.go
+++ b/govcd/edgegateway.go
@@ -216,7 +216,7 @@ func (eGW *EdgeGateway) RemoveNATRule(id string) error {
 	return nil
 }
 
-// RemoveNATRuleAsync removes NAT rule and return  or error.
+// RemoveNATRuleAsync removes NAT rule or returns an error.
 // Old functions RemoveNATPortMapping and RemoveNATMapping removed using rule details
 // and expected interface to be of external network type.
 func (eGW *EdgeGateway) RemoveNATRuleAsync(id string) (Task, error) {
@@ -265,9 +265,9 @@ func (eGW *EdgeGateway) RemoveNATRuleAsync(id string) (Task, error) {
 		"application/vnd.vmware.admin.edgeGatewayServiceConfiguration+xml", "error reconfiguring Edge Gateway: %s", newRules)
 }
 
-// AddDNATRule creates DNAT rule and return created NAT struct or error.
-// Allows to assign specific Org VDC or external network.
-// When edge gateway is advanced vCD API uses element <tag> to map with NSX edge gateway ID. Currently existing issue,
+// AddDNATRule creates DNAT rule and returns the NAT struct that was created or an error.
+// Allows assigning a specific Org VDC or an external network.
+// When edge gateway is advanced vCD API uses element <tag> to map with NSX edge gateway ID. A known issue is
 // that updating rule using User interface resets <tag> and as result mapping is lost.
 // Fetching using NatRule.ID won't be valid anymore.
 // Old functions AddNATPortMapping and AddNATMapping assigned rule only to first external network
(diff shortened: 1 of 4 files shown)


govcd/edgegateway.go
```diff
@@ -311,87 +326,173 @@ func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, descrip
...
+		return
+	}
+
+	uuid = fmt.Sprintf("%X-%X-%X-%X-%X", b[0:4], b[4:6], b[6:8], b[8:10], b[10:])
+
+	return
+}
+
+// UpdateNatRule updates NAT rule and handles task. Returns updated NAT rule or error.
+func (eGW *EdgeGateway) UpdateNatRule(natRule *types.NatRule) (*types.NatRule, error) {
```
Didainius: Would be nice to have some tests covering `Update` and `Fetch`

govcd/edgegateway.go
```diff
@@ -311,87 +326,173 @@ func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, descrip
...
-	return eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule, nil
+	for _, natRule := range eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule {
+		if natRule.Description == mappingId {
+			createdNatRule = natRule
+			break
+		}
+	}
+
+	if createdNatRule == nil {
+		return nil, fmt.Errorf("error creating SNAT rule, didn't matched created rule")
```
Didainius:
```suggestion
		return nil, fmt.Errorf("error creating SNAT rule, didn't match created rule")
```

govcd/edgegateway.go
```diff
@@ -200,82 +200,75 @@ func (eGW *EdgeGateway) RemoveNATPortRule(networkHref, natType, externalIP, exte
...
-// TODO: remove this function when ruleId is available
-func extractObjectIDfromPath(locationPath string) (string, error) {
-	if locationPath == "" {
-		return "", fmt.Errorf("unable to get ID from empty path")
+// RemoveNATRule removes NAT rule and handles task. Return errors when issue risen.
+// Old functions RemoveNATPortMapping and RemoveNATMapping removes using rule details and expects interface is external network.
+func (eGW *EdgeGateway) RemoveNATRule(id string) error {
+	task, err := eGW.RemoveNATRuleAsync(id)
+	if err != nil {
+		return fmt.Errorf("error creating DNAT rule: %#v", err)
```
Didainius:
```suggestion
		return fmt.Errorf("error removing DNAT rule: %#v", err)
```

govcd/edgegateway.go
```diff
@@ -200,82 +200,75 @@ func (eGW *EdgeGateway) RemoveNATPortRule(networkHref, natType, externalIP, exte
...
+	} else {
+		return Task{}, fmt.Errorf("edge gtw doesn't have NAT rules")
 	}
 
-	newEdgeConfig.NatService = newNatService
+	if ruleIndex == -1 {
+		return Task{}, fmt.Errorf("edge gtw doesn't have rule with such Id")
+	}
+
+	natServiceToUpdate.NatRule = append(natServiceToUpdate.NatRule[:ruleIndex], natServiceToUpdate.NatRule[ruleIndex+1:]...)
```
Didainius: Wouldn't this fail on incorrect index if we had only 1 rule (array element 0)?

govcd/edgegateway.go
```diff
@@ -311,87 +326,173 @@ func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, descrip
...
+	uuid = fmt.Sprintf("%X-%X-%X-%X-%X", b[0:4], b[4:6], b[6:8], b[8:10], b[10:])
+
+	return
+}
+
+// UpdateNatRule updates NAT rule and handles task. Returns updated NAT rule or error.
+func (eGW *EdgeGateway) UpdateNatRule(natRule *types.NatRule) (*types.NatRule, error) {
+	task, err := eGW.UpdateNatRuleAsync(natRule)
+	if err != nil {
+		return nil, fmt.Errorf("error updating SNAT rule: %#v", err)
```
Didainius: As I understand it handles both SNAT and DNAT?
```suggestion
		return nil, fmt.Errorf("error updating NAT rule: %#v", err)
```

vbauzysvmware:
> ```shell
> Test_NATPortMapping
> ```

Test_NATPortMapping and Test_NATMapping do this. These two methods use old delete which isn't working correctly in these test cases.

Possbile solutions: 
* leave as these are deprecated methods and deprecated tetss
* change tests to use new delete method - may be possible. Test still will verify deprecated methods.

govcd/edgegateway.go
```diff
@@ -145,15 +145,22 @@ func (eGW *EdgeGateway) AddDhcpPool(network *types.OrgVDCNetwork, dhcppool []int
...
+func (eGW *EdgeGateway) RemoveNATMapping(natType, externalIP, internalIP, port string) (Task, error) {
+	return eGW.RemoveNATPortMapping(natType, externalIP, port, internalIP, port)
 }
 
-// Temporary fix for existing Remove functions as it doesn't support network interfaces
-// TODO: remove this function when ruleId is available
-func (eGW *EdgeGateway) RemoveNATPortRule(networkHref, natType, externalIP, externalPort, internalIP, internalPort string) (Task, error) {
+// Deprecated in favor of RemoveNATRuleAsync, RemoveNATRule
+func (eGW *EdgeGateway) RemoveNATPortMapping(natType, externalIP, externalPort, internalIP, internalPort string) (Task, error) {
+	// Find uplink interface
```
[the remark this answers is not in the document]
vbauzysvmware: I guess when external network doesn't exist

govcd/edgegateway.go
vbauzysvmware: Fixed

govcd/edgegateway.go
vbauzysvmware: didn't fail. Added check to be 100% valid

govcd/edgegateway.go
vbauzysvmware: Fixed

govcd/edgegateway.go
vbauzysvmware: Fixed

govcd/edgegateway_test.go
```diff
@@ -434,29 +404,22 @@ func (vcd *TestVCD) Test_AddDNATRule(check *C) {
...
+
+	check.Assert(natRule.GatewayNatRule.TranslatedIP, Equals, vcd.config.VCD.InternalIp)
+	check.Assert(natRule.GatewayNatRule.TranslatedPort, Equals, "88")
+	check.Assert(natRule.GatewayNatRule.OriginalIP, Equals, vcd.config.VCD.ExternalIp)
+	check.Assert(natRule.GatewayNatRule.OriginalPort, Equals, "1188")
+	check.Assert(natRule.GatewayNatRule.Protocol, Equals, "tcp")
+	check.Assert(natRule.GatewayNatRule.IcmpSubType, Equals, "")
+	check.Assert(natRule.Description, Equals, description2)
+	check.Assert(natRule.RuleType, Equals, "DNAT")
+	//check.Assert(natRule.GatewayNatRule.Interface.HREF, Equals, externalNetwork.ExternalNetwork.HREF)
```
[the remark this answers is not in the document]
vbauzysvmware: Fixed

govcd/edgegateway.go
```diff
@@ -311,87 +326,173 @@ func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, descrip
...
+	natServiceToUpdate := eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService
+
+	if natServiceToUpdate != nil {
+		for n, existingNatRule := range natServiceToUpdate.NatRule {
+			if existingNatRule.ID == natRule.ID {
+				natServiceToUpdate.NatRule[n] = natRule
+			}
+		}
+	} else {
+		return Task{}, fmt.Errorf("edge gtw doesn't have such nat rule")
```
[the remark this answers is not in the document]
vbauzysvmware: Fixed

govcd/edgegateway.go
```diff
@@ -287,22 +280,44 @@ func (eGW *EdgeGateway) AddDNATRule(ruleDetails NatRule) ([]*types.NatRule, erro
...
-	return eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule, nil
+	for _, natRule := range eGW.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule {
+		if natRule.Description == mappingId {
+			createdNatRule = natRule
+			break
+		}
+	}
+
+	if createdNatRule == nil {
+		return nil, fmt.Errorf("error creating SNAT rule, didn't matched created rule")
```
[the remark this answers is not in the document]
vbauzysvmware: Fixed

govcd/edgegateway.go
```diff
@@ -311,87 +326,173 @@ func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, descrip
...
+	}
+
+	createdNatRule.Description = description
+
+	return eGW.UpdateNatRule(createdNatRule)
+}
+
+// Creates unique ID/UUID
```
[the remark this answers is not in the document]
vbauzysvmware: This private function, anyway added function name

govcd/edgegateway.go
```diff
@@ -311,87 +326,173 @@ func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, descrip
...
+func getPseudoUuid() (uuid string) {
+
+	b := make([]byte, 16)
+	_, err := rand.Read(b)
+	if err != nil {
+		fmt.Println("Error: ", err)
```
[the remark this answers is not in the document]
vbauzysvmware: change to be returned

govcd/edgegateway.go
vbauzysvmware: Added test

lvirbalas:
> After the tests I've got 2 nat rules left:

@Didainius @vbauzysvmware was the resolution for this issue?

vbauzysvmware:
> > After the tests I've got 2 nat rules left:
> 
> @Didainius @vbauzysvmware was there resolution for this issue?

Test_NATPortMapping and Test_NATMapping do this - they are old unit tests. These two test methods use old delete which isn't working correctly in these test cases.

Possbile solutions:

    leave as these are deprecated methods and deprecated tetss
    change tests to use new delete method - may be possible. Test still will verify deprecated methods.

govcd/edgegateway.go
```diff
@@ -43,14 +43,14 @@ func NewEdgeGateway(cli *Client) *EdgeGateway {
 // Struct which covers NAT rule fields
 type NatRule struct {
 	natType      string
```
lvirbalas: You capitalized all fields of this struct, but left this one poor guy lower case :)

govcd/edgegateway.go
```diff
@@ -200,82 +200,82 @@ func (eGW *EdgeGateway) RemoveNATPortRule(networkHref, natType, externalIP, exte
 
 }
 
-// Temporary fix to support parsing network Id from HREF
-// TODO: remove this function when ruleId is available
-func extractObjectIDfromPath(locationPath string) (string, error) {
-	if locationPath == "" {
-		return "", fmt.Errorf("unable to get ID from empty path")
+// RemoveNATRule removes NAT rule and handles task. Return errors when issue risen.
```
lvirbalas: For clarity: "removes NAT rule and handles task" -> "removes NAT rule identified by ID and handles task"

Grammar: "Return errors when issue risen" -> "Returns error if issues rise"

govcd/edgegateway.go
```diff
@@ -200,82 +200,82 @@ func (eGW *EdgeGateway) RemoveNATPortRule(networkHref, natType, externalIP, exte
...
+// Old functions RemoveNATPortMapping and RemoveNATMapping removes using rule details and expects interface is external network.
```
lvirbalas: "removes using rule details and expects interface is external network" -> "removed using rule details and expected interface to be of external network type."

govcd/edgegateway.go
```diff
@@ -200,82 +200,82 @@ func (eGW *EdgeGateway) RemoveNATPortRule(networkHref, natType, externalIP, exte
...
-	// Find uplink interface
-	var uplink types.Reference
-	for _, gi := range eGW.EdgeGateway.Configuration.GatewayInterfaces.GatewayInterface {
-		if gi.InterfaceType != "uplink" {
-			continue
-		}
-		uplink = *gi.Network
+// RemoveNATRuleAsync removes NAT rule and return  or error.
+// Old functions RemoveNATPortMapping and RemoveNATMapping
+// removes using rule details and expects interface is external network.
```
lvirbalas: "removes using rule details and expects interface is external network" -> "removed using rule details and expected interface to be of external network type."

govcd/edgegateway.go
```diff
@@ -200,82 +200,82 @@ func (eGW *EdgeGateway) RemoveNATPortRule(networkHref, natType, externalIP, exte
...
 
-// AddDNATRule creates DNAT rule and return refreshed existing NAT rules array or error
-// Allows to assign specific network Org VDC or external. Old function AddNATPortMapping and
-// AddNATMapping assigns rule to first external network
-func (eGW *EdgeGateway) AddDNATRule(ruleDetails NatRule) ([]*types.NatRule, error) {
+// AddDNATRule creates DNAT rule and return created NAT struct or error.
+// Allows to assign specific Org VDC or external network.
+// When edge gateway is advanced vCD API uses element <tag> to map with NSX edge gateway ID. Currently existing issue,
+// that changed rule using UI resets <tag> as so mapping and as result fetched NatRule.ID won't be valid anymore.
+// Old functions AddNATPortMapping and AddNATMapping assigns rule only to first external network
```
lvirbalas: Grammar: "assigns" -> "assigned"

govcd/edgegateway.go
```diff
@@ -200,82 +200,82 @@ func (eGW *EdgeGateway) RemoveNATPortRule(networkHref, natType, externalIP, exte
...
 }
 
-// AddDNATRule creates DNAT rule and return refreshed existing NAT rules array or error
-// Allows to assign specific network Org VDC or external. Old function AddNATPortMapping and
-// AddNATMapping assigns rule to first external network
-func (eGW *EdgeGateway) AddDNATRule(ruleDetails NatRule) ([]*types.NatRule, error) {
+// AddDNATRule creates DNAT rule and return created NAT struct or error.
+// Allows to assign specific Org VDC or external network.
+// When edge gateway is advanced vCD API uses element <tag> to map with NSX edge gateway ID. Currently existing issue,
+// that changed rule using UI resets <tag> as so mapping and as result fetched NatRule.ID won't be valid anymore.
```
lvirbalas: > // When edge gateway is advanced vCD API uses element <tag> to map with NSX edge gateway ID. Currently existing issue,
> // that changed rule using UI resets <tag> as so mapping and as result fetched NatRule.ID won't be valid anymore.

This comment tries to convey important info, but it's not clear to a new reader. Questions come up:
- What is this <tag> and why it's being mentioned here?
- How UI comes into play here?

After reading this it looks like NAT rule ID is mangled only if touched by ID, but that's not the case, right? Need to rephrase to add clarity that NAT rule ID is removed (?) if created with API (if that's the case).

govcd/edgegateway.go
```diff
@@ -287,22 +287,47 @@ func (eGW *EdgeGateway) AddDNATRule(ruleDetails NatRule) ([]*types.NatRule, erro
...
+	return eGW.UpdateNatRule(createdNatRule)
 }
 
-// AddSNATRule creates SNAT rule and returns refreshed existing NAT rules array or error
-// Allows to assign specific network Org VDC or external. Old function AddNATPortMapping and
-// AddNATMapping assigns rule to first external network
-func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, description string) ([]*types.NatRule, error) {
+// AddSNATRule creates SNAT rule and returns created NAT rule or error.
+// Allows to assign specific Org VDC or external network.
+// Old function AddNATPortMapping and AddNATMapping assigns rule to only first external network
```
lvirbalas: Grammer: "assigns rule to only" -> "assigned rule only to"

govcd/edgegateway.go
```diff
@@ -287,22 +287,47 @@ func (eGW *EdgeGateway) AddDNATRule(ruleDetails NatRule) ([]*types.NatRule, erro
...
+func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, description string) (*types.NatRule, error) {
+
+	// As vCD API don't return rule API we fetch it manually:
```
lvirbalas: "don't return rule API" -> "doesn't return rule ID"?

govcd/edgegateway.go
```diff
@@ -287,22 +287,47 @@ func (eGW *EdgeGateway) AddDNATRule(ruleDetails NatRule) ([]*types.NatRule, erro
...
+	//  * create rule with description which value is our generated Id
+	//  * find rule which has description with our generated Id
```
lvirbalas: Please add additional bullet:
`* read the real (vCD's) rule Id`

govcd/edgegateway.go
```diff
@@ -311,87 +336,172 @@ func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, descrip
...
+	egwConfigureHref, _ := url.ParseRequestURI(eGW.EdgeGateway.HREF)
+	egwConfigureHref.Path += "/action/configureServices"
+
+	// Return the task
+	return eGW.client.ExecuteTaskRequest(egwConfigureHref.String(), http.MethodPost,
+		"application/vnd.vmware.admin.edgeGatewayServiceConfiguration+xml", "error reconfiguring Edge Gateway: %s", newRules)
+}
+
+// FetchNatRule returns NAT rule or error.
+func (eGW *EdgeGateway) FetchNatRule(id string) (*types.NatRule, error) {
```
lvirbalas: I believe we agreed to use `Read...` after all. @dataclouder please confirm if nothing changed.

govcd/edgegateway.go
```diff
@@ -311,87 +336,172 @@ func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, descrip
...
+			}
+		}
+	}
+
+	return nil, ErrorEntityNotFound
 }
 
 // AddNATRuleAsync creates NAT rule and return task or err
 // Allows to assign specific network Org VDC or external. Old function AddNATPortMapping and
 // AddNATMapping assigns rule to first external network
```
lvirbalas: "assigns" -> "assigned"

govcd/edgegateway.go
vbauzysvmware: This is done intentionally. This field is set by specific functions.

govcd/edgegateway.go
vbauzysvmware: I wasn't informed. You did meeting without me. I was against READ

govcd/edgegateway.go
vbauzysvmware: Comment changed

govcd/edgegateway.go
vbauzysvmware: done

govcd/edgegateway.go
vbauzysvmware: fixed

govcd/edgegateway.go
vbauzysvmware: fixed

govcd/edgegateway.go
vbauzysvmware: change comment

govcd/edgegateway.go
vbauzysvmware: not sure about past tense, cause function still are and we comment only to show user why they are deprecated

vbauzysvmware: changed

govcd/edgegateway.go
vbauzysvmware: changed

govcd/edgegateway.go
vbauzysvmware: changed

govcd/edgegateway.go
vbauzysvmware: Comment improved. 
Mentioned cause here is risk that you use ID and it won't work anymore if you update rule using UI

govcd/edgegateway.go
Didainius: But in this case having exactly one rule would miss it. Wouldn't it?

vbauzysvmware: From running point works fine. Also added check for 1 rule case

Didainius: Ok. Got it.

govcd/edgegateway.go
```diff
@@ -145,15 +145,22 @@ func (eGW *EdgeGateway) AddDhcpPool(network *types.OrgVDCNetwork, dhcppool []int
 
 }
 
-// Temporary fix for existing Remove functions as it doesn't support network interfaces
-// TODO: remove this function when ruleId is available
-func (eGW *EdgeGateway) RemoveNATMappingRule(networkHref, natType, externalIP, internalIP, port string) (Task, error) {
-	return eGW.RemoveNATPortRule(networkHref, natType, externalIP, port, internalIP, port)
+// Deprecated in favor of RemoveNATRuleAsync, RemoveNATRule
```
Didainius: I risk to be annoying but I just read the suggested deprecation string and I think it's great opportunity to use it as there may be some tooling that is aware of it. Added an example below from standard lib.
It still contains function documentation and then a `Deprecated: followed by some information about the deprecation`


https://blog.golang.org/godoc-documenting-go-code
> Sometimes a struct field, function, type, or even a whole package becomes redundant or unnecessary, but must be kept for compatibility with existing programs. To signal that an identifier should not be used, add a paragraph to its doc comment that begins with "Deprecated:" followed by some information about the deprecation. There are a few examples in the standard library. 

```go
// Copy returns a new Regexp object copied from re.
// Calling Longest on one copy does not affect another.
//
// Deprecated: In earlier releases, when using a Regexp in multiple goroutines,
// giving each goroutine its own copy helped to avoid lock contention.
// As of Go 1.12, using Copy is no longer necessary to avoid lock contention.
// Copy may still be appropriate if the reason for its use is to make
// two copies with different Longest settings.
func (re *Regexp) Copy() *Regexp {
	re2 := *re
	return &re2
}
```

Here's how it looks in the generated documentation: https://golang.org/pkg/regexp/#Regexp.Copy

govcd/edgegateway.go
```diff
@@ -145,15 +145,22 @@ func (eGW *EdgeGateway) AddDhcpPool(network *types.OrgVDCNetwork, dhcppool []int
...
+func (eGW *EdgeGateway) RemoveNATMapping(natType, externalIP, internalIP, port string) (Task, error) {
+	return eGW.RemoveNATPortMapping(natType, externalIP, port, internalIP, port)
 }
 
-// Temporary fix for existing Remove functions as it doesn't support network interfaces
-// TODO: remove this function when ruleId is available
-func (eGW *EdgeGateway) RemoveNATPortRule(networkHref, natType, externalIP, externalPort, internalIP, internalPort string) (Task, error) {
+// Deprecated in favor of RemoveNATRuleAsync, RemoveNATRule
```
Didainius: Same as above - would be nice to fix deprecation string.

Didainius:
> > > After the tests I've got 2 nat rules left:
> > 
> > 
> > @Didainius @vbauzysvmware was there resolution for this issue?
> 
> Test_NATPortMapping and Test_NATMapping do this - they are old unit tests. These two test methods use old delete which isn't working correctly in these test cases.
> 
> Possbile solutions:
> 
> ```
> leave as these are deprecated methods and deprecated tetss
> change tests to use new delete method - may be possible. Test still will verify deprecated m
> ```

I vote for option 2 and fixing tests. It constantly leaves mess in NAT table and we will not remove it soon.

govcd/edgegateway.go
vbauzysvmware: Fixed

govcd/edgegateway.go
vbauzysvmware: Fixed

govcd/edgegateway.go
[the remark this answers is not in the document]
dataclouder: What happens if the edge gateway has two uplink interfaces?

govcd/edgegateway.go
```diff
@@ -200,82 +200,84 @@ func (eGW *EdgeGateway) RemoveNATPortRule(networkHref, natType, externalIP, exte
...
 
-func (eGW *EdgeGateway) RemoveNATPortMapping(natType, externalIP, externalPort, internalIP, internalPort string) (Task, error) {
-	// Find uplink interface
-	var uplink types.Reference
-	for _, gi := range eGW.EdgeGateway.Configuration.GatewayInterfaces.GatewayInterface {
-		if gi.InterfaceType != "uplink" {
-			continue
-		}
-		uplink = *gi.Network
+// RemoveNATRuleAsync removes NAT rule and return  or error.
```
dataclouder: "and return or error" -> "or returns an error"?

govcd/edgegateway.go
```diff
@@ -200,82 +200,84 @@ func (eGW *EdgeGateway) RemoveNATPortRule(networkHref, natType, externalIP, exte
...
+	return eGW.client.ExecuteTaskRequest(egwConfigureHref.String(), http.MethodPost,
 		"application/vnd.vmware.admin.edgeGatewayServiceConfiguration+xml", "error reconfiguring Edge Gateway: %s", newRules)
-
 }
 
-// AddDNATRule creates DNAT rule and return refreshed existing NAT rules array or error
-// Allows to assign specific network Org VDC or external. Old function AddNATPortMapping and
-// AddNATMapping assigns rule to first external network
-func (eGW *EdgeGateway) AddDNATRule(ruleDetails NatRule) ([]*types.NatRule, error) {
+// AddDNATRule creates DNAT rule and return created NAT struct or error.
```
dataclouder: "and return created NAT struct or error" -> "and returns the NAT struct that was created or an error"

govcd/edgegateway.go
```diff
@@ -200,82 +200,84 @@ func (eGW *EdgeGateway) RemoveNATPortRule(networkHref, natType, externalIP, exte
...
+// Allows to assign specific Org VDC or external network.
```
dataclouder: "Allows to assign specific Org VDC or external network" -> "Allows assigning a specific Org VDC or an external network"

govcd/edgegateway.go
```diff
@@ -200,82 +200,84 @@ func (eGW *EdgeGateway) RemoveNATPortRule(networkHref, natType, externalIP, exte
...
+// When edge gateway is advanced vCD API uses element <tag> to map with NSX edge gateway ID. Currently existing issue,
```
dataclouder: "Currently existing issue," -> "A known issue is "

govcd/edgegateway.go
```diff
@@ -287,22 +289,48 @@ func (eGW *EdgeGateway) AddDNATRule(ruleDetails NatRule) ([]*types.NatRule, erro
...
+
+	return eGW.UpdateNatRule(createdNatRule)
 }
 
-// AddSNATRule creates SNAT rule and returns refreshed existing NAT rules array or error
-// Allows to assign specific network Org VDC or external. Old function AddNATPortMapping and
-// AddNATMapping assigns rule to first external network
-func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, description string) ([]*types.NatRule, error) {
+// AddSNATRule creates SNAT rule and returns created NAT rule or error.
+// Allows to assign specific Org VDC or external network.
```
dataclouder: "to assign" -> "assigning"

govcd/edgegateway.go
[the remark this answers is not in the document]
vbauzysvmware: first one used

govcd/edgegateway.go
```diff
@@ -311,87 +339,172 @@ func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, descrip
...
+	if createdNatRule == nil {
+		return nil, fmt.Errorf("error creating SNAT rule, didn't match created rule")
+	}
+
+	createdNatRule.Description = description
+
+	return eGW.UpdateNatRule(createdNatRule)
+}
+
+// getPseudoUuid creates unique ID/UUID
```
dataclouder: Please add a Unit test for this function. BTW, could we use a library function (such as Google UUID) for this purpose?

govcd/edgegateway.go
dataclouder: We agreed on Read.
See https://github.com/vmware/go-vcloud-director/issues/211

govcd/edgegateway.go
```diff
@@ -311,87 +339,172 @@ func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, descrip
...
+		}
+	}
+
+	return nil, ErrorEntityNotFound
 }
 
 // AddNATRuleAsync creates NAT rule and return task or err
 // Allows to assign specific network Org VDC or external. Old function AddNATPortMapping and
-// AddNATMapping assigns rule to first external network
+// AddNATMapping function shouldn't be used cause assigns rule to first external network
```
dataclouder: "cause" -> "because"

govcd/edgegateway.go
vbauzysvmware: Changed

govcd/edgegateway.go
vbauzysvmware: Fixed

govcd/edgegateway.go
vbauzysvmware: Changed

govcd/edgegateway.go
vbauzysvmware: Changed

govcd/edgegateway.go
vbauzysvmware: Could you help me with test? I didn't want to introduce dependency for simple matter for generating ID which is used for 1s. Even math.rand may be sufficient..

govcd/edgegateway.go
vbauzysvmware: fixed

vbauzysvmware:
> > > > After the tests I've got 2 nat rules left:
> > > 
> > > 
> > > @Didainius @vbauzysvmware was there resolution for this issue?
> > 
> > 
> > Test_NATPortMapping and Test_NATMapping do this - they are old unit tests. These two test methods use old delete which isn't working correctly in these test cases.
> > Possbile solutions:
> > ```
> > leave as these are deprecated methods and deprecated tetss
> > change tests to use new delete method - may be possible. Test still will verify deprecated m
> > ```
> 
> I vote for option 2 and fixing tests. It constantly leaves mess in NAT table and we will not remove it soon.

Changed to use new delete - now clean.

govcd/edgegateway.go
vbauzysvmware: fixed

govcd/edgegateway.go
vbauzysvmware: fixed

govcd/edgegateway.go
dataclouder: Sent to you in Slack

vbauzysvmware: Thank you, added The test

govcd/edgegateway.go
lvirbalas: Re-met, re-evaluated & decided to go with `Get` everywhere.

govcd/edgegateway.go
lvirbalas: Thanks. BTW, I suggested past tense to reflect that those functions are "old" and this one is new

govcd/edgegateway.go
vbauzysvmware: already changed

govcd/edgegateway_test.go
```diff
@@ -465,3 +540,23 @@ func (vcd *TestVCD) Test_AddDNATRule(check *C) {
 
 	check.Assert(len(edge.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule), Equals, beforeChangeNatRulesNumber)
 }
+
+func TestGetPseudoUUID(t *testing.T) {
```
dataclouder: This function should be in `edgegateway_unit_test.go` with tag `unit`, similar to `system_unit_test.go`

vbauzysvmware: fixed

govcd/edgegateway.go
```diff
@@ -200,82 +200,84 @@ func (eGW *EdgeGateway) RemoveNATPortRule(networkHref, natType, externalIP, exte
...
-	for _, gi := range eGW.EdgeGateway.Configuration.GatewayInterfaces.GatewayInterface {
-		if gi.InterfaceType != "uplink" {
-			continue
-		}
-		uplink = *gi.Network
+// RemoveNATRuleAsync removes NAT rule or returns an error.
+// Old functions RemoveNATPortMapping and RemoveNATMapping removed using rule details
+// and expected interface to be of external network type.
+func (eGW *EdgeGateway) RemoveNATRuleAsync(id string) (Task, error) {
+	if "" == id {
```
dataclouder: Please change to `if id == "" {`

govcd/edgegateway.go
```diff
@@ -311,87 +339,172 @@ func (eGW *EdgeGateway) AddSNATRule(networkHref, externalIP, internalIP, descrip
...
+	if err != nil {
+		return nil, fmt.Errorf("%s", combinedTaskErrorMessage(task.Task, err))
+	}
+
+	return eGW.GetNatRule(natRule.ID)
+}
+
+// UpdateNatRuleAsync updates NAT rule and returns task or error.
+func (eGW *EdgeGateway) UpdateNatRuleAsync(natRule *types.NatRule) (Task, error) {
+	if "" != natRule.GatewayNatRule.Protocol && !isValidProtocol(natRule.GatewayNatRule.Protocol) {
```
dataclouder: Please change `if "" != natRule.GatewayNatRule.Protocol` to `if natRule.GatewayNatRule.Protocol != ""`

govcd/edgegateway.go
vbauzysvmware: Fixed

govcd/edgegateway.go
vbauzysvmware: Fixed

govcd/edgegateway.go
lvirbalas: Which specific functions? Why does it have to be "different"?

vbauzysvmware: AddSNat and AddDNAT - you have pass to different type of rule to API

vbauzysvmware:
```

vbauzysvmware: Made public, cause you still can access AddNATRuleAsync as it's public, so field still have to be setted.

lvirbalas: Thanks!

PR #217, opened 2019-07-17
Refactored Get* methods 1

dataclouder:
This is the first batch of changes to support the creation of data sources in terraform-provider-vcd.

Changes:
* Add Get* methods for Org and AdminOrg
* Add function getUuidFromHref
* Add function getOrgHREFById, to retrieve an HREF from an Org/AdminOrg ID
* Add method vcdClient.GetOrgByName
* Add method vcdClient.GetOrgById
* Add method vcdClient.GetOrgByNameOrId
* Add method vcdClient.GetAdminOrgByName
* Add method vcdClient.GetAdminOrgById
* Add method vcdClient.GetAdminOrgByNameOrId
* Add Fix for values in OrgGeneralSettings, which are ignored unless some fields are always filled.
* Add test for getUuidFromHref
* Add tests for all Read* methods
* Add more tests for Org creation
* Deprecate functions GetOrgByName, GetAdminOrgByName
* Replace all calls to the above function with the new vcdClient.Get* methods
*  Fix AdminOrg.Update method and add more tests
* Add CODING_GUIDELINES.md

Note: when a function was returning `Org` and other entities, I have only changed the return type for `Org` (from structure to pointer), leaving the refactoring of the other return types when we tackle the other entities.


govcd/catalog_test.go
```diff
@@ -44,8 +44,8 @@ func (vcd *TestVCD) Test_FindCatalogItem(check *C) {
 // Creates a Catalog, updates the description, and checks the changes against the
 // newly updated catalog. Then deletes the catalog
 func (vcd *TestVCD) Test_UpdateCatalog(check *C) {
-	org, err := GetAdminOrgByName(vcd.client, vcd.config.VCD.Org)
-	check.Assert(org, Not(Equals), AdminOrg{})
+	org, err := vcd.client.GetAdminOrgByName(vcd.config.VCD.Org)
+	check.Assert(org, NotNil)
```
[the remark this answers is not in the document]
dataclouder: In most cases, these were mechanical replacements of existing checks. I will look into it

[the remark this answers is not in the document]
dataclouder: done

govcd/system.go
```diff
@@ -528,3 +587,111 @@ func QueryOrgVdcNetworkByName(vcdCli *VCDClient, name string) ([]*types.QueryRes
...
+	return adminOrg, nil
+}
+
+// GetAdminOrgByNameOrId finds an Admin Organization by name or ID
+// On success, returns a pointer to the Admin Org structure and a nil error
+// On failure, returns a nil pointer and an error
+func (vcdClient *VCDClient) GetAdminOrgByNameOrId(identifier string) (*AdminOrg, error) {
+
+	adminOrg, err := vcdClient.GetAdminOrgById(identifier)
+	if err != nil {
```
vbauzysvmware: here if error isn't `entity not foun` then it's lost

govcd/user.go
```diff
@@ -71,9 +71,9 @@ func NewUser(cli *Client, org *AdminOrg) *OrgUser {
 	}
 }
 
-// FetchUserByHref returns a user by its HREF, without need for
+// GetUserByHref returns a user by its HREF, without need for
 // searching in the adminOrg user list
-func (adminOrg *AdminOrg) FetchUserByHref(href string) (*OrgUser, error) {
```
vbauzysvmware: Isn't  `FetchUserByHref` was already release and we have to deprecate?

govcd/user.go
```diff
@@ -85,13 +85,13 @@ func (adminOrg *AdminOrg) FetchUserByHref(href string) (*OrgUser, error) {
...
 
-// FetchUserByName retrieves a user within an admin organization by name
+// GetUserByName retrieves a user within an admin organization by name
 // Returns a valid user if it exists. If it doesn't, returns nil and ErrorEntityNotFound
 // If argument refresh is true, the AdminOrg will be refreshed before searching.
 // This is usually done after creating, modifying, or deleting users.
 // If it is false, it will search within the data already in memory (useful when
 // looping through the users and we know that no changes have occurred in the meantime)
-func (adminOrg *AdminOrg) FetchUserByName(name string, refresh bool) (*OrgUser, error) {
+func (adminOrg *AdminOrg) GetUserByName(name string, refresh bool) (*OrgUser, error) {
```
vbauzysvmware: Isn't FetchUserByName was already release and we have to deprecate?

govcd/user.go
```diff
@@ -101,19 +101,19 @@ func (adminOrg *AdminOrg) FetchUserByName(name string, refresh bool) (*OrgUser,
...
 }
 
-// FetchUserById retrieves a user within an admin organization by ID
+// GetUserById retrieves a user within an admin organization by ID
 // Returns a valid user if it exists. If it doesn't, returns nil and ErrorEntityNotFound
 // If argument refresh is true, the AdminOrg will be refreshed before searching.
 // This is usually done after creating, modifying, or deleting users.
 // If it is false, it will search within the data already in memory (useful when
 // looping through the users and we know that no changes have occurred in the meantime)
-func (adminOrg *AdminOrg) FetchUserById(id string, refresh bool) (*OrgUser, error) {
```
vbauzysvmware: Same question

govcd/user.go
```diff
@@ -123,31 +123,31 @@ func (adminOrg *AdminOrg) FetchUserById(id string, refresh bool) (*OrgUser, erro
...
-// FetchUserByNameOrId retrieves a user within an admin organization
+// GetUserByNameOrId retrieves a user within an admin organization
 // by either name or ID
 // Returns a valid user if it exists. If it doesn't, returns nil and ErrorEntityNotFound
 // If argument refresh is true, the AdminOrg will be refreshed before searching.
 // This is usually done after creating, modifying, or deleting users.
 // If it is false, it will search within the data already in memory (useful when
 // looping through the users and we know that no changes have occurred in the meantime)
-func (adminOrg *AdminOrg) FetchUserByNameOrId(identifier string, refresh bool) (*OrgUser, error) {
+func (adminOrg *AdminOrg) GetUserByNameOrId(identifier string, refresh bool) (*OrgUser, error) {
```
vbauzysvmware: Same question

govcd/user.go
```diff
@@ -123,31 +123,31 @@ func (adminOrg *AdminOrg) FetchUserById(id string, refresh bool) (*OrgUser, erro
...
-		orgUser, err = adminOrg.FetchUserById(identifier, false)
+		orgUser, err = adminOrg.GetUserById(identifier, false)
 	}
 	return orgUser, err
 }
 
-// GetRole finds a role within the organization
-func (adminOrg *AdminOrg) GetRole(roleName string) (*types.Reference, error) {
+// GetRoleReference finds a role within the organization
+func (adminOrg *AdminOrg) GetRoleReference(roleName string) (*types.Reference, error) {
```
vbauzysvmware: Same question

govcd/user.go
dataclouder: Blast! This change was supposed to be in 2.3 already.
But we have changed our collective mind a few times. 
I will change to deprecation + new methods.

govcd/system.go
dataclouder: We don't know if `identifier` contains a name or an ID.
The most probable reason that this call fails is that we are looking by ID using a name. It is expected that it fails in this case, and we immediately search by name.
Whatever is returned at that point, is a legitimate error.

dataclouder: Changed the implementation of `GetAdminOrgByNameOrId` (and also `GetOrgByNameOrId` and `GetUserByNameOrId`) to make it more readable and robust

govcd/user.go
dataclouder: yes. Fixed

govcd/user.go
dataclouder: fixed

govcd/user.go
dataclouder: fixed

govcd/user.go
dataclouder: fixed

govcd/user.go
dataclouder: fixed

govcd/system.go
```diff
@@ -51,6 +51,14 @@ func CreateOrg(vcdClient *VCDClient, name string, fullName string, description s
 		OrgSettings: settings,
 	}
 
+	// There is a bug in the settings of CanPublishCatalogs.
```
Didainius: In the light of recent finding with load balancers and `bool` fields having `omitempty` (https://github.com/vmware/go-vcloud-director/issues/223)

Can this bug be caused that when `settings.OrgGeneralSettings.UseServerBootSequence = false` it is actually not sent via API?

Probably this is not this PRs scope, but I see danger in these `bool` fields having `omitempty` now. If user had `CanPublishCatalogs=true`, but decided to _update_ it to `false` - we would never set it because the fields would be skipped unmarshaled XML representation.

image

govcd/org_test.go
```diff
@@ -352,8 +354,8 @@ func (vcd *TestVCD) Test_Admin_FindCatalog(check *C) {
 		check.Skip(fmt.Sprintf(TestRequiresSysAdminPrivileges, check.TestName()))
 	}
 	// Fetch admin org version of current test org
-	adminOrg, err := GetAdminOrgByName(vcd.client, vcd.org.Org.Name)
-	check.Assert(adminOrg, Not(Equals), AdminOrg{})
+	adminOrg, err := vcd.client.GetAdminOrgByName(vcd.org.Org.Name)
+	check.Assert(adminOrg, NotNil)
```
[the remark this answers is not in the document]
dataclouder: done

govcd/system_test.go
```diff
@@ -85,23 +168,46 @@ func (vcd *TestVCD) Test_CreateOrg(check *C) {
...
+		check.Assert(err, IsNil)
+		// After a successful creation, the entity is added to the cleanup list.
+		// If something fails after this point, the entity will be removed
+		AddToCleanupList(orgName, "org", "", "TestCreateOrg")
+		err = task.WaitTaskCompletion()
+		check.Assert(err, IsNil)
+		// fetch newly created org
+		adminOrg, err := vcd.client.GetAdminOrgByName(orgName)
+		check.Assert(adminOrg, NotNil)
+		check.Assert(err, IsNil)
```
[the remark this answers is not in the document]

govcd/system_test.go
```diff
@@ -197,8 +303,8 @@ func (vcd *TestVCD) Test_GetNetworkPoolByHREF(check *C) {
 
 	fmt.Printf("Running: %s\n", check.TestName())
 
-	adminOrg, err := GetAdminOrgByName(vcd.client, vcd.config.VCD.Org)
-	check.Assert(adminOrg, Not(Equals), AdminOrg{})
+	adminOrg, err := vcd.client.GetAdminOrgByName(vcd.config.VCD.Org)
+	check.Assert(adminOrg, NotNil)
```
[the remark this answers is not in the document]

govcd/system.go
dataclouder: I added an explicit test for updates using these fields both in govcd and Terraform

Didainius: Thanks for the test. Apparently proxied NSX API and vCD API behave differently. When the field is skipped vCD API inherently sets it to `false`, while the proxied NSX API leaves previous value. This works well in your case.

govcd/system.go
```diff
@@ -337,6 +365,37 @@ func getOrgHREF(vcdClient *VCDClient, orgName string) (string, error) {
 	return "", fmt.Errorf("couldn't find org with name: %s. Please check Org name as it is case sensitive", orgName)
 }
 
+// Returns the HREF of the org from the org ID
+func getOrgHREFById(vcdClient *VCDClient, orgId string) (string, error) {
+	orgListHREF := vcdClient.Client.VCDHREF
```
vbauzysvmware: Maybe it's possible to Query using href/id in filter?

dataclouder: Query struct doesn't have Organization info.

vbauzysvmware: Yes, It would be need to add.

PR #219, opened 2019-07-23
Add global load balancing configuration support for edge gateway

Didainius:
Work continues on https://github.com/terraform-providers/terraform-provider-vcd/issues/223

This PR adds capability to set global load balancing configuration on edge gateway as per the below UI screenshot.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 2.3.0 (Unreleased)
 
 * Added edge gateway create/delete functions [#130](https://github.com/vmware/go-vcloud-director/issues/130).
-* Added global load balancer configuration support [#219](https://github.com/vmware/go-vcloud-director/pull/219)
+* Added edge gateway global load balancer configuration support (e.g. enable/disable) [#219](https://github.com/vmware/go-vcloud-director/pull/219)
 * Added load balancer service monitor [#196](https://github.com/vmware/go-vcloud-director/pull/196)
 * Added load balancer server pool [#205](https://github.com/vmware/go-vcloud-director/pull/205)
 * Added load balancer application profile [#208](https://github.com/vmware/go-vcloud-director/pull/208)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 2.3.0 (Unreleased)
 
 * Added edge gateway create/delete functions [#130](https://github.com/vmware/go-vcloud-director/issues/130).
+* Added global load balancer configuration support [#219](https://github.com/vmware/go-vcloud-director/pull/219)
 * Added load balancer service monitor [#196](https://github.com/vmware/go-vcloud-director/pull/196)
 * Added load balancer server pool [#205](https://github.com/vmware/go-vcloud-director/pull/205)
 * Added load balancer application profile [#208](https://github.com/vmware/go-vcloud-director/pull/208)
--- a/govcd/edgegateway.go
+++ b/govcd/edgegateway.go
@@ -9,7 +9,6 @@ import (
 	"crypto/rand"
 	"encoding/xml"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"net/url"
 	"regexp"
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 2.3.0 (Unreleased)
 
 * Added edge gateway create/delete functions [#130](https://github.com/vmware/go-vcloud-director/issues/130).
+* Added global load balancer configuration support [#219](https://github.com/vmware/go-vcloud-director/pull/219)
```
lvirbalas: "configuration" is rather generic if you don't know what this issue is about, what about adding more detail like:
`Added global load balancer configuration support (e.g. enable/disable)`

govcd/edgegateway.go
```diff
@@ -1076,3 +1076,88 @@ func (eGW *EdgeGateway) buildProxiedEdgeEndpointURL(optionalSuffix string) (stri
...
+	return loadBalancerConfig, nil
+}
+
+// UpdateLoadBalancer allows to update global load balancer configuration.
+// These are 4 fields only: LoadBalancer.Enabled, LoadBalancer.AccelerationEnabled,
+// LoadBalancer.Logging.Enable, LoadBalancer.Logging.LogLevel. They are represented in load balancer
+// global configuration tab in the UI.
+// All other fields are ignored and sent as they are in order to prevent load balancer configuration
+// corruption
+func (egw *EdgeGateway) UpdateLoadBalancer(lb *types.LoadBalancer) (*types.LoadBalancer, error) {
```
lvirbalas: Someone may trip on the function name. What about `UpdateLoadBalancerGlobal` to be more specific?

vbauzysvmware: Global? - I think it's wrong. It's only for 4 fields. We need to be more specific. I agree current name isn't enough and we need other ones `EnableLoadBalancer`, `UpdateLoadBalancerEnalblament`.. other options? UpdateGenericSomething should be used for main update functionality.

Didainius: Changed.

CHANGELOG.md
Didainius: Changed. I also added "edge gateway" in that sentence so that it is clearer.

govcd/edgegateway.go
```diff
@@ -1076,3 +1076,88 @@ func (eGW *EdgeGateway) buildProxiedEdgeEndpointURL(optionalSuffix string) (stri
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve load balancer after update: %s", err)
+	}
+
+	return updatedLb, nil
+}
+
+// validateUpdateLoadBalancer validates mandatory fields for global load balancer configuration
+// settings
+func validateUpdateLoadBalancer(lb *types.LoadBalancer) error {
```
vbauzysvmware: Kind'a strange. I thouhgh LoadBalancer.Enabled, LoadBalancer.AccelerationEnabled are main thing's

Didainius: Yeah, but these are `bool`. They have a default value (of false), I thought about using `*bool` to evaluate `nil` conditions, but then the constructs become quite ugly as you must use an intermediate var as you cannot do something like `&false`

govcd/edgegateway.go
Didainius: I must admit that `UpdateLoadBalancerGlobal` doesn't sound to precise as well as `UpdateLoadBalancer`, but saying that it is only `enable` is confusing as you can also tune logging options in this tab as per screenshot below. The `global` comes from tab name, but not sure what other names we could make:
* `UpdateLoadBalancerConfig` ?

If I split this to only `enable` then there will be more functions for logging and more calls which I'd say are not necessary.



image

govcd/edgegateway.go
vbauzysvmware: Yep, in such cases then we expose values which implicitly set default values and we request from user to have them, we kinda help them and also hide what they may explicitly set. I think topic for discussion for overall API.

govcd/edgegateway.go
vbauzysvmware: I agree enable is just partly say what is doing. Definitely against *Global. Also don't like *Config in this case at this also very abstract. We need brainstorm session :)

vbauzysvmware: I saw that UI names this Global configuration, but from API point of view is just flat fields of LB

govcd/edgegateway.go
```diff
@@ -1076,3 +1076,88 @@ func (eGW *EdgeGateway) buildProxiedEdgeEndpointURL(optionalSuffix string) (stri
...
+	currentLb, err := egw.GetLoadBalancer()
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve load balancer before update: %s", err)
+	}
+
+	// Modify only the global configuration settings
+	currentLb.Enabled = lb.Enabled
+	currentLb.AccelerationEnabled = lb.AccelerationEnabled
+	currentLb.Logging = lb.Logging
+	// Ommit the version as it is updated automatically with each put
```
dataclouder: "Ommit" -> "omit"

govcd/edgegateway.go
```diff
@@ -1076,3 +1076,88 @@ func (eGW *EdgeGateway) buildProxiedEdgeEndpointURL(optionalSuffix string) (stri
...
+	currentLb.Version = ""
+
+	// Push updated configuration
```
dataclouder: Should we have a check here to see whether the intended fields are already set as we want them?
If so, we could just return the current LB without further action

govcd/lb_test.go
```diff
@@ -106,19 +108,46 @@ func (vcd *TestVCD) Test_LB(check *C) {
...
 	fmt.Printf("Done\n")
 
-	buildLoadBalancer(ip1, ip2, vcd, check)
+	// Build load balancer
+	buildLoadBalancer(edge, ip1, ip2, vcd, check)
+
+	// Cache current load balancer settings for change validation in the end
+	beforeLb, beforeLbXml := testCacheLoadBalancer(edge, check)
+
+	// Enable load balancer globally
```
dataclouder: In the interest of failing quickly, can we move this operation to the top? That way, if this operation fails, we don't need to wait for the longer tasks to complete.

govcd/edgegateway_test.go
```diff
@@ -538,3 +542,111 @@ func (vcd *TestVCD) Test_UpdateNATRule(check *C) {
...
+	check.Assert(err, IsNil)
+
+	req := edge.client.NewRequest(nil, http.MethodGet, *reqUrl, nil)
+
+	resp, err := checkResp(edge.client.Http.Do(req))
+	check.Assert(err, IsNil)
+
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
```
dataclouder: Since we are bypassing `decodeBody`, could you add a call to `util.ProcessResponseOutput` here?
Or even better, could you make this operation an API call that avoids unmarshaling? We may need it for similar operations with other features.

govcd/lb_test.go
Didainius: I want to keep it there for a few reasons:
* If I enable load balancing before all the configuration is created (`buildLoadBalancer(edge, ip1, ip2, vcd, check)`) then there is less configuration available in load balancer and less chances to catch corrupt config (matching raw XMLs)
* `testCacheLoadBalancer` and `testCheckLoadBalancerConfig` `XML` validationg would not work in this case because majority of load balancer component cleanup is left for teardown part in the end of suite

govcd/edgegateway.go
Didainius: Changed. I need to kick bad grammar habits of my head.

govcd/edgegateway.go
Didainius: Very good spot. Done.

govcd/edgegateway.go
lvirbalas: Then, at least to me looks like `UpdateLoadBalancerGlobalConfig` would be most precise even though long. What do you say?

vbauzysvmware: It possible, but I don't like it, cause it's UI name, not API. We agree that govcd matches API and terraform UI namings.

govcd/edgegateway_test.go
Didainius: In fact the `ExecuteRequestWithCustomError` does it. There was a problem with reading `resp.Body` twice, but I think it is solved now.

govcd/edgegateway.go
lvirbalas: We found that NSX API calls it "General Load Balancer Parameters"
image (1)

govcd/edgegateway.go
Didainius: Actually initially when you get the load balancer it is disabled so the default boolean value `false` plays quite fair here.

govcd/edgegateway.go
lvirbalas: Decided:
```
UpdateLBGeneralParams(<four fields>)
GetLBGeneralParams() (*types.LBGeneralParamsWithXML, error)
```

govcd/edgegateway.go
vbauzysvmware: ok then, in this case.

PR #220, opened 2019-07-25
Load balancer syntax cleanup and bugfixes

Didainius:


govcd/lbservicemonitor.go
```diff
@@ -191,28 +195,32 @@ func validateCreateLBServiceMonitor(lbMonitorConfig *types.LBMonitor) error {
...
+
 	if lbMonitorConfig.ID == "" && lbMonitorConfig.Name == "" {
 		return fmt.Errorf("to read load balancer service monitor at least one of `ID`, `Name` fields must be specified")
 	}
 
 	return nil
 }
 
-func validateUpdateLBServiceMonitor(lbMonitorConfig *types.LBMonitor) error {
+func validateUpdateLbServiceMonitor(lbMonitorConfig *types.LbMonitor, egw *EdgeGateway) error {
```
lvirbalas: Should the EdgeGateway be the receiver as opposed to parameter for these few `validate` refactored functions?

Didainius: I thought of it. Can change. My point was to avoid further polluting EdgeGateway with functions which don't help being used separately.

govcd/lb_test.go
```diff
@@ -18,7 +18,7 @@ import (
 	. "gopkg.in/check.v1"
 )
 
-// Test_LB load balancer integration test
+// Test_Lb load balancer integration test
 // MANDATORY PREREQUISITE - manually enable load balancing for edge gateway with enabled acceleration
```
dataclouder: Perhaps we can remove the "manually enable", since the test now does it

govcd/lbappprofile_test.go
```diff
@@ -27,28 +27,28 @@ func (vcd *TestVCD) Test_LBAppProfile(check *C) {
...
 
 	// Lookup by both name and ID and compare that these are equal values
-	lbAppProfileByID, err := edge.ReadLBAppProfile(&types.LBAppProfile{ID: createdLbAppProfile.ID})
+	lbAppProfileByID, err := edge.GetLbAppProfile(&types.LbAppProfile{ID: createdLbAppProfile.ID})
 	check.Assert(err, IsNil)
 
-	lbPoolByName, err := edge.ReadLBAppProfile(&types.LBAppProfile{Name: createdLbAppProfile.Name})
+	lbPoolByName, err := edge.GetLbAppProfile(&types.LbAppProfile{Name: createdLbAppProfile.Name})
 	check.Assert(err, IsNil)
 	check.Assert(createdLbAppProfile.ID, Equals, lbPoolByName.ID)
```
dataclouder: There should be a check that `lbPoolByName` is not nil

govcd/lbappprofile.go
```diff
@@ -91,92 +91,100 @@ func (eGW *EdgeGateway) ReadLBAppProfile(lbAppProfileConfig *types.LBAppProfile)
...
 		"error while updating load balancer application profile : %s", lbAppProfileConfig, &types.NSXError{})
 	if err != nil {
 		return nil, err
 	}
 
-	readAppProfile, err := eGW.ReadLBAppProfileByID(lbAppProfileConfig.ID)
+	readAppProfile, err := egw.GetLbAppProfileById(lbAppProfileConfig.ID)
 	if err != nil {
 		return nil, fmt.Errorf("unable to retrieve application profile with ID (%s) after update: %s",
 			readAppProfile.ID, err)
```
dataclouder: If err is not nil, readAppProfile should be nil, and can't reference and ID

govcd/lbappprofile.go
```diff
@@ -11,59 +11,59 @@ import (
...
+	lbAppProfileID, err := extractNsxObjectIdFromPath(resp.Header.Get("Location"))
 	if err != nil {
 		return nil, err
 	}
 
-	readAppProfile, err := eGW.ReadLBAppProfileByID(lbAppProfileID)
+	readAppProfile, err := egw.GetLbAppProfileById(lbAppProfileID)
 	if err != nil {
 		return nil, fmt.Errorf("unable to retrieve application profile with ID (%s) after creation: %s",
 			readAppProfile.ID, err)
```
dataclouder: readAppProfile.ID seems to be accessing a nil pointer

Didainius: Yes. Changed it.

govcd/lbappprofile.go
Didainius: Yes. Changed it.

govcd/lbappprofile_test.go
Didainius: Added here, and to more places.

govcd/lb_test.go
Didainius: Done.

PR #221, opened 2019-07-26
Parse error of &types.NSXError{} for Load balancer Delete methods

Didainius:
At the moment load balancer components do not parse NSX error (&types.NSXError{}) if the Delete method fails. This does not give proper information to user without it.

Currently the provider gives such error:
```
vcd_lb_service_monitor.monitor: Destroying... [id=monitor-42]

Error: error deleting load balancer service monitor: unable to delete Service Monitor: API Error: 0:
```
After this PR it would be:
```
vcd_lb_service_monitor.monitor: Destroying... [id=monitor-42]

Error: error deleting load balancer service monitor: unable to delete service monitor: vShield Edge [LoadBalancer] Fail to delete objectId monitor-42 for it is used by pool-29 (API error: 14570)
```

### DIFF
--- a/govcd/lbserverpool_test.go
+++ b/govcd/lbserverpool_test.go
@@ -92,7 +92,7 @@ func (vcd *TestVCD) Test_LBServerPool(check *C) {
 
 	// Try to delete used service monitor and expect it to fail with nice error
 	err = edge.DeleteLbServiceMonitor(lbMon)
-	check.Assert(err, ErrorMatches, ".*Fail to delete objectId .* for it is used by .*")
+	check.Assert(err, ErrorMatches, `.*Fail to delete objectId .*\S+.* for it is used by .*`)
 
 	// Lookup by both name and ID and compare that these are equal values
 	lbPoolByID, err := edge.getLbServerPool(&types.LbPool{ID: createdLbPool.ID})
--- a/govcd/lbserverpool_test.go
+++ b/govcd/lbserverpool_test.go
@@ -90,6 +90,10 @@ func (vcd *TestVCD) Test_LBServerPool(check *C) {
 	// We created server pool successfully therefore let's add it to cleanup list
 	AddToCleanupList(TestLbServerPool, "lbServerPool", parentEntity, check.TestName())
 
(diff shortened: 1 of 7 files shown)


govcd/lbserverpool_test.go
```diff
@@ -90,6 +90,10 @@ func (vcd *TestVCD) Test_LBServerPool(check *C) {
 	// We created server pool successfully therefore let's add it to cleanup list
 	AddToCleanupList(TestLbServerPool, "lbServerPool", parentEntity, check.TestName())
 
+	// Try to delete used service monitor and expect it to fail with nice error
+	err = edge.DeleteLbServiceMonitor(lbMon)
+	check.Assert(err, ErrorMatches, ".*Fail to delete objectId .* for it is used by .*")
```
dataclouder: Replace `objectId .*` with `objectId .*\S+.*` to make sure that there is some ID returned.

govcd/lbvirtualserver_test.go
```diff
@@ -73,6 +73,16 @@ func (vcd *TestVCD) Test_LBVirtualServer(check *C) {
 	check.Assert(createdLbVirtualServer.ApplicationRuleIds, DeepEquals, lbVirtualServerConfig.ApplicationRuleIds)
 	check.Assert(createdLbVirtualServer.DefaultPoolId, Equals, lbVirtualServerConfig.DefaultPoolId)
 
+	// Try to delete child components and expect a a well parsed NSX error
+	err = edge.DeleteLbServiceMonitorById(serviceMonitorId)
+	check.Assert(err, ErrorMatches, ".*Fail to delete objectId .* for it is used by .*")
```
dataclouder: Make the regexp more aggressive by checking that there is actually an ID and an owner, as suggested above

govcd/lbserverpool_test.go
Didainius: Thanks. Improved and tested at least on 9.5 and 9.7

govcd/lbvirtualserver_test.go
Didainius: Thanks. Improved and tested at least on 9.5 and 9.7

PR #224, opened 2019-07-30
Add methods to find provider VDC entities

dataclouder:
Add method VCDClient.QueryProviderVdcs
Add method VCDClient.QueryNetworkPools
Add method VCDClient.QueryProviderVdcStorageProfiles

The main reason for these additions is to improve the configuration file generation, and allow us to create them using just the credentials, while the provider VDC is retrieved from the server.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -316,7 +316,7 @@ While running tests, the following environment variables can be used:
 * `GOVCD_CONFIG=/path/file`: sets an alternative configuration file for the tests.
    e.g.:  `GOVCD_CONFIG=/some/path/govcd_test_config.yaml go test -tags functional -timeout 0 .`
 * `GOVCD_DEBUG=1`: enable debug output on screen.
-* `TEST_VERBOSE=1`: shows execution details in unit tests.
+* `GOVCD_TEST_VERBOSE=1`: shows execution details in some tests.
 * `GOVCD_SKIP_VAPP_CREATION=1`: will not create the initial vApp. All tests that
    depend on a vApp availability will be skipped.
 * `GOVCD_TASK_MONITOR=show|log|simple_show|simple|log`: sets a task monitor function when running `task.WaitCompletion`
--- a/govcd/system.go
+++ b/govcd/system.go
@@ -464,6 +464,7 @@ func getExtension(client *Client) (*types.Extension, error) {
 }
 
 // QueryProviderVdcStorageProfileByName finds a provider VDC storage profile by name
+// Deprecated: use vcdClient.QueryProviderVdcStorageProfileByName instead
 func QueryProviderVdcStorageProfileByName(vcdCli *VCDClient, name string) ([]*types.QueryResultProviderVdcStorageProfileRecordType, error) {
 	results, err := vcdCli.QueryWithNotEncodedParams(nil, map[string]string{
 		"type":   "providerVdcStorageProfile",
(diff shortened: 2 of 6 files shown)


govcd/system_test.go
```diff
@@ -233,3 +234,82 @@ func (vcd *TestVCD) Test_QueryOrgVdcNetworkByName(check *C) {
...
+	providerVdcName := vcd.config.VCD.ProviderVdc.Name
+	networkPoolName := vcd.config.VCD.ProviderVdc.NetworkPool
+	storageProfileName := vcd.config.VCD.ProviderVdc.StorageProfile
+	if providerVdcName == "" {
+		check.Skip("Skipping Provider VDC query: no provider VDC was given")
+	}
+	providerVdcs, err := vcd.client.QueryProviderVdcs()
+	check.Assert(err, IsNil)
+
+	testVerbose := os.Getenv("TEST_VERBOSE") != ""
```
Didainius: Thinking if we should make "global" for testing and then potentially use everywhere. Currently I guess we print out (or not) in "random" fashion.

govcd/system.go
```diff
@@ -502,6 +505,94 @@ func QueryProviderVdcByName(vcdCli *VCDClient, name string) ([]*types.QueryResul
...
+// QueryNetworkPoolByName finds a network pool by name
+func (vcdClient *VCDClient) QueryNetworkPoolByName(name string) (*types.QueryResultNetworkPoolRecordType, error) {
+	results, err := vcdClient.QueryWithNotEncodedParams(nil, map[string]string{
+		"type":   "networkPool",
+		"filter": fmt.Sprintf("(name==%s)", name),
+	})
+	if err != nil {
+		return nil, err
+	}
+	for _, networkPool := range results.Results.NetworkPoolRecord {
```
vbauzysvmware: I am not definitely sure is right approach. Cause `name==%s` already do this. Though name may be regexp when result will be quite different. And if resource can have a few instances with same name? Same comment is valid for other changed Query functions!

Didainius: Yeah. I also started to think what can happen in case when multiple objects are allowed to have the same name. (IIRC so far we know independent disks can have the same names)

dataclouder: I know about independent disks, but in this case we know that the objects we are looking for have unique names.

I don't particularly care about having a Query*ByName with a single result at the moment, since these functions are unused. I proposed this rewrite because the original one looks wrong to me, because the name is deceiving and because it returns a `nil` error when not found.
 
My goal with this PR is to have a list of Provider VDCs, Network Pools, and Storage Profiles, which are needed for automation, and we lack the ability of fetching the list.

However, these functions will become important when (or if) we implement data sources for such entities, or when we implement data sources for other entities that need to be retrieved through Query functions. At that stage, we need a way of getting those entities by ID, and sometimes by name.

If there is no agreement on this function, I'll remove it and only keep the list functions only.

vbauzysvmware: For what purpose these functions could be used it's another story. One e.g. can be that returning attributes are different. If you want to now which ext network egtw is using - you have to use query result, cause direct GET won't provide such information. Going back regarding implementation, IMHO, isn't right if we query by name and expect one result when check len() and return result. Currectly it is doing same check against name and qeury return a few results it's choosing only first result. And such behaviour I don't like

dataclouder: I have removed the methods `vcdClient.QueryNetworkPoolByName`, `vcdClient.QueryProviderVdcStorageProfileByName`, and `vcdClient.QueryProviderVdcByName`, and the deprecation notice from the corresponding functions.

TESTING.md
```diff
@@ -316,7 +316,7 @@ While running tests, the following environment variables can be used:
 * `GOVCD_CONFIG=/path/file`: sets an alternative configuration file for the tests.
    e.g.:  `GOVCD_CONFIG=/some/path/govcd_test_config.yaml go test -tags functional -timeout 0 .`
 * `GOVCD_DEBUG=1`: enable debug output on screen.
-* `TEST_VERBOSE=1`: shows execution details in unit tests.
+* `GOVCD_TEST_VERBOSE=1`: shows execution details in some tests.
```
lvirbalas: Shall we add a changelog note that this has been renamed?

dataclouder: I don't think it's needed.

PR #229, opened 2019-08-12
Add networkupdate support for `VM`, introduce forced guest customization function

Didainius:
Ref https://github.com/terraform-providers/terraform-provider-vcd/issues/299, https://github.com/terraform-providers/terraform-provider-vcd/issues/300

This PR introduces a few things:
* Add function for network updates (`vm.UpdateNetworkConnectionSection()`)
* Add forced guest customization related functions

A few general lint fixes:
* Get rid of staticcheck lint error `SA1015: using time.Tick leaks the underlying ticker, consider using it only in endless functions, tests and the main package, and use time.NewTicker here` 
* Use http client with timeout in `lb_test.go`

### DIFF
--- a/govcd/lb_test.go
+++ b/govcd/lb_test.go
@@ -263,12 +263,14 @@ func checkLb(queryUrl string, expectedResponses []string, maxRetryTimeout int) e
 		iterations = maxRetryTimeout / 5
 	}
 
+	httpClient := &http.Client{Timeout: 5}
+
 	fmt.Printf("# Waiting for the virtual server to accept responses (%s interval x %d iterations)"+
 		"\n[_ = timeout, x = connection refused, ?(err) = unknown error, / = no nodes are up yet, "+
 		". = no response from all nodes yet]: ", sleepIntervalDuration.String(), iterations)
 	for i := 1; i <= iterations; i++ {
 		var resp *http.Response
-		resp, err = http.Get(queryUrl)
+		resp, err = httpClient.Get(queryUrl)
 		if err != nil {
 			switch {
 			case strings.Contains(err.Error(), "i/o timeout"):
--- a/govcd/lb_test.go
+++ b/govcd/lb_test.go
@@ -263,7 +263,7 @@ func checkLb(queryUrl string, expectedResponses []string, maxRetryTimeout int) e
 		iterations = maxRetryTimeout / 5
 	}
 
-	httpClient := &http.Client{Timeout: 5}
+	httpClient := &http.Client{Timeout: 5 * time.Second}
 
 	fmt.Printf("# Waiting for the virtual server to accept responses (%s interval x %d iterations)"+
 		"\n[_ = timeout, x = connection refused, ?(err) = unknown error, / = no nodes are up yet, "+
--- a/govcd/lb_test.go
+++ b/govcd/lb_test.go
@@ -248,63 +248,68 @@ func buildLb(edge EdgeGateway, node1Ip, node2Ip string, vcd *TestVCD, check *C) 
 // checkLb queries specified endpoint until it gets all responses in expectedResponses slice
 func checkLb(queryUrl string, expectedResponses []string, maxRetryTimeout int) error {
 	var err error
-	var iterations int
 	if len(expectedResponses) == 0 {
 		return fmt.Errorf("no expected responses specified")
 	}
 
+	retryTimeout := maxRetryTimeout
 	// due to the VMs taking long time to boot it needs to be at least 5 minutes
 	// may be even more in slower environments
-	sleepInterval := 5
-	sleepIntervalDuration := time.Duration(sleepInterval) * time.Second
 	if maxRetryTimeout < 5*60 { // 5 minutes
-		iterations = 60
-	} else {
-		iterations = maxRetryTimeout / 5
+		retryTimeout = 5 * 60 // 5 minutes
 	}
+
+	timeOutAfterInterval := time.Duration(retryTimeout) * time.Second
+	timeoutAfter := time.After(timeOutAfterInterval)
+	tick := time.NewTicker(time.Duration(5) * time.Second)
 
 	httpClient := &http.Client{Timeout: 5 * time.Second}
 
-	fmt.Printf("# Waiting for the virtual server to accept responses (%s interval x %d iterations)"+
+	fmt.Printf("# Waiting for the virtual server to accept responses (timeout after %s)"+
 		"\n[_ = timeout, x = connection refused, ?(err) = unknown error, / = no nodes are up yet, "+
-		". = no response from all nodes yet]: ", sleepIntervalDuration.String(), iterations)
-	for i := 1; i <= iterations; i++ {
-		var resp *http.Response
-		resp, err = httpClient.Get(queryUrl)
-		if err != nil {
-			switch {
-			case strings.Contains(err.Error(), "i/o timeout"):
-				fmt.Printf("_")
-			case strings.Contains(err.Error(), "connect: connection refused"):
-				fmt.Printf("x")
-			case strings.Contains(err.Error(), "connect: network is unreachable"):
-				fmt.Printf("/")
-			default:
-				fmt.Printf("?(%s)", err.Error())
-			}
-		}
+		". = no response from all nodes yet]: ", timeOutAfterInterval.String())
 
-		if err == nil {
-			fmt.Printf(".") // progress bar when waiting for responses from all nodes
-			body, _ := ioutil.ReadAll(resp.Body)
-			// check if the element is in the list
-			for index, value := range expectedResponses {
-				if value == string(body) {
-					expectedResponses = append(expectedResponses[:index], expectedResponses[index+1:]...)
-					if len(expectedResponses) > 0 {
-						fmt.Printf("\n# '%s' responded. Waiting for node(s) '%s': ",
-							value, strings.Join(expectedResponses, ","))
-					} else {
-						fmt.Printf("\n# Last node '%s' responded. Exiting\n", value)
-						return nil
+	for {
+		select {
+		case <-timeoutAfter:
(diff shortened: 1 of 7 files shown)


govcd/vm.go
```diff
@@ -85,6 +86,35 @@ func (vm *VM) GetNetworkConnectionSection() (*types.NetworkConnectionSection, er
 	return networkConnectionSection, err
 }
 
+// UpdateNetworkConnectionSection applies network configuration of types.NetworkConnectionSection for the VM
+// Runs synchronously, VM is ready for another operation after this function returns.
+func (vm *VM) UpdateNetworkConnectionSection(networks *types.NetworkConnectionSection) error {
+	if vm.VM.HREF == "" {
+		return fmt.Errorf("cannot refresh, Object is empty")
```
dataclouder: "refresh"?

govcd/lb_test.go
```diff
@@ -263,12 +263,14 @@ func checkLb(queryUrl string, expectedResponses []string, maxRetryTimeout int) e
 		iterations = maxRetryTimeout / 5
 	}
 
+	httpClient := &http.Client{Timeout: 5 * time.Second}
```
dataclouder: Could you add a comment explaining how this timeout affects the expected duration of the test?

I am talking about the estimate given with `Waiting for the virtual server to accept responses (%s interval x %d iterations)"`

govcd/vm.go
```diff
@@ -107,6 +137,35 @@ func (vm *VM) PowerOn() (Task, error) {
 
 }
 
+// PowerOnAndForceCustomization is a synchronous function which is equivalent to the functionality
+// one has in UI. It triggers customization which may be useful in some cases (like altering NICs)
+//
+// The VM _must_ be Undeployed for this action to actually work.
```
dataclouder: Can we detect the state of the VM to make sure this prerequisite is met?

govcd/vm_test.go
```diff
@@ -810,3 +810,101 @@ func (vcd *TestVCD) Test_VMPowerOnPowerOff(check *C) {
...
+	if vmName == "" {
+		check.Skip("skipping test because no VM is found")
+	}
+
+	vm, err := vcd.client.Client.FindVMByHREF(vmType.HREF)
+	check.Assert(err, IsNil)
+
+	networkBefore, err := vm.GetNetworkConnectionSection()
+	check.Assert(err, IsNil)
+	//check.Assert(network, IsNil)
```
dataclouder: leftover comment: can it be removed?

govcd/vm_test.go
```diff
@@ -810,3 +810,101 @@ func (vcd *TestVCD) Test_VMPowerOnPowerOff(check *C) {
...
+	check.Assert(err, IsNil)
+
+	// It may be that prebuilt VM was not booted before in the test vApp and it would still have
+	// a guest customization status 'GC_PENDING'. This is because initially VM has this flag set
+	// but while this flag is here the test cannot actually check if vm.PowerOnAndForceCustomization()
+	// gives any effect therefore we must "wait through" initial guest customization if it is in
+	// 'GC_PENDING' state.
+	custStatus, err := vm.GetGuestCustomizationStatus()
+	check.Assert(err, IsNil)
+	if custStatus == "GC_PENDING" {
```
dataclouder: should `CG_PENDING` be an item in `types.VAppStatuses`?

govcd/vm.go
```diff
@@ -85,6 +86,35 @@ func (vm *VM) GetNetworkConnectionSection() (*types.NetworkConnectionSection, er
 	return networkConnectionSection, err
 }
 
+// UpdateNetworkConnectionSection applies network configuration of types.NetworkConnectionSection for the VM
+// Runs synchronously, VM is ready for another operation after this function returns.
+func (vm *VM) UpdateNetworkConnectionSection(networks *types.NetworkConnectionSection) error {
```
vbauzysvmware: Usually Giuseppe request to have functions which returns tasks. @dataclouder

govcd/vm.go
vbauzysvmware: IMHO/ Yes, we should. VM should have status.

govcd/vm.go
```diff
@@ -107,6 +137,35 @@ func (vm *VM) PowerOn() (Task, error) {
 
 }
 
+// PowerOnAndForceCustomization is a synchronous function which is equivalent to the functionality
+// one has in UI. It triggers customization which may be useful in some cases (like altering NICs)
+//
+// The VM _must_ be Undeployed for this action to actually work.
+func (vm *VM) PowerOnAndForceCustomization() error {
```
vbauzysvmware: Same question regarding exposing task

govcd/vm.go
```diff
@@ -302,6 +361,43 @@ func (vm *VM) RunCustomizationScript(computername, script string) (Task, error)
 	return vm.Customize(computername, script, false)
 }
 
+func (vm *VM) GetGuestCustomizationStatus() (string, error) {
```
vbauzysvmware: Comment missing

govcd/vm.go
```diff
@@ -302,6 +361,43 @@ func (vm *VM) RunCustomizationScript(computername, script string) (Task, error)
...
+	_, err := vm.client.ExecuteRequest(vm.VM.HREF+"/guestcustomizationstatus", http.MethodGet,
+		types.MimeGuestCustomizationStatus, "error retrieving guest customization status: %s", nil, guestCustomizationStatus)
+
+	// The request was successful
+	return guestCustomizationStatus.GuestCustStatus, err
+}
+
+// BlockWhileGuestCustomizationStatus blocks until the customization status of VM exits unwantedStatus.
+// It sleeps 3 seconds between iterations and times out after timeOutAfterSeconds of seconds.
+func (vm *VM) BlockWhileGuestCustomizationStatus(unwantedStatus string, timeOutAfterSeconds int) error {
```
vbauzysvmware: Can we have validations for statuses and seconds?

govcd/vm.go
Didainius: Wow. That was way off. Fixed.

govcd/lb_test.go
Didainius: Yep. This simple loop was actually quite unstable in guessing timing. I put in a `time.Ticker` so that it does not lie and actually work as the time was set.

govcd/vm.go
Didainius: I can add a check, but the problem is that it doesn't actually fail to power on and god knows if it can be useful in any scenario. As far as I've investigated it simply never exits "GC_PENDING" if it wasn't un-deployed before using `PowerOnAndForceCustomization`.

govcd/vm_test.go
Didainius: Done.

govcd/vm_test.go
Didainius: No, it is not vApp status. It is VM guest customization status. It can be constants on their own. Added known statuses to constants.

govcd/vm.go
Didainius: Not 100% sure is we really know all the possible statuses. As far as time I can, but probably that could be limited to simply checking if it is not zero, nothing more

govcd/vm.go
vbauzysvmware: Now I lost between explanation and doc:  "The VM _must_ be Undeployed for this action to actually work." ?

govcd/vm.go
vbauzysvmware: So let's add atleast time validation, to not create some unlimited one - constant loop

govcd/vm.go
Didainius: Ok. On a second read that may not be too clear. Power on works anyway, but the VM never exits `GC_PENDING` state if it was not undeployed.

vbauzysvmware: doesn't `vm.status == undeployed` isn't sufficient?

Didainius: It may be. The problem is that if I enforce `vm.status == undeployed`, I'm not 100% sure it is the only case it would be useful.

govcd/vm.go
Didainius: Ok. What upper limit do we think makes sense? 30 mins?

govcd/vm.go
dataclouder: No issues with this code from my side

govcd/vm.go
dataclouder: no issue here

govcd/vm.go
dataclouder: I don't understand. If we can check the status, we can simply abort when the wanted status is not set.
What am I missing?

govcd/vm.go
vbauzysvmware: could be and also not less zero, I don't know how it would work with -30.

govcd/vm.go
Didainius: Added.

govcd/vm.go
Didainius: Added.

govcd/vm.go
Didainius: Added to be 4<X<7200  (minimum is 4, because first check kicks in after 3 seconds)
I left  quite big maximum as I do remember having some big Windows "guest machine customization" happening > 1hour from my experience. Hope this is enough.

lvirbalas: Not a big fan of timeouts hidden deep inside the code... Is there a way to have a configurable timeout parameter which would be used as a common one for more than one function? This configuration parameter could be hidden from the user to avoid complexity, however, in case of user issue, we could ask to increase it for troubleshooting.

Didainius: The timeout is here. `timeOutAfterSeconds int` is a parameter, it just has the validator now as guys wanted. From usage perspective at the moment I only really needed it for tests so far, but one can change it.

PR #230, opened 2019-08-13
Fix for external network serialization and deserialization

vbauzysvmware:
Changes to fix issue where external network description wasn't read. Issue noticed by Giuseppe.

Create needs to use two namespaces and `go` don't work with them, so namespaces are hardcoded. Read functionality needs seperate struct as API returns different look of xml from create in regards of namespaces(prefix)

### DIFF
--- a/govcd/externalnetwork_test.go
+++ b/govcd/externalnetwork_test.go
@@ -29,7 +29,7 @@ func (vcd *TestVCD) Test_ExternalNetworkGetByName(check *C) {
 }
 
 // Helper function that creates an external network to be used in other tests
-func (vcd *TestVCD) testCreateExternalNetwork(testName, networkName, dnsSuffix string) (skippingReason string, externalNetwork *types.ExternalNetworkCreate, task Task, err error) {
+func (vcd *TestVCD) testCreateExternalNetwork(testName, networkName, dnsSuffix string) (skippingReason string, externalNetwork *types.ExternalNetwork, task Task, err error) {
 
 	if vcd.skipAdminTests {
 		return fmt.Sprintf(TestRequiresSysAdminPrivileges, testName), externalNetwork, Task{}, nil
@@ -72,12 +72,9 @@ func (vcd *TestVCD) testCreateExternalNetwork(testName, networkName, dnsSuffix s
 		return fmt.Sprintf("More than one port group found with name '%s'", vcd.config.VCD.ExternalNetworkPortGroup), externalNetwork, Task{}, nil
 	}
 
-	externalNetwork = &types.ExternalNetworkCreate{
+	externalNetwork = &types.ExternalNetwork{
 		Name:        networkName,
 		Description: "Test Create External Network",
-		XmlnsVmext:  types.XMLNamespaceExtension,
-		XmlnsVcloud: types.XMLNamespaceVCloud,
-		Type:        types.MimeExternalNetwork,
 		Configuration: &types.NetworkConfiguration{
 			Xmlns: types.XMLNamespaceVCloud,
 			IPScopes: &types.IPScopes{
@@ -99,9 +96,9 @@ func (vcd *TestVCD) testCreateExternalNetwork(testName, networkName, dnsSuffix s
 				}},
 			FenceMode: "isolated",
 		},
-		VimPortGroupRefs: &types.VimObjectRefsCreate{
-			VimObjectRef: []*types.VimObjectRefCreate{
-				&types.VimObjectRefCreate{
+		VimPortGroupRefs: &types.VimObjectRefs{
+			VimObjectRef: []*types.VimObjectRef{
+				&types.VimObjectRef{
 					VimServerRef: &types.Reference{
 						HREF: vimServerHref,
 					},
--- a/govcd/externalnetwork_test.go
+++ b/govcd/externalnetwork_test.go
@@ -29,7 +29,7 @@ func (vcd *TestVCD) Test_ExternalNetworkGetByName(check *C) {
 }
 
 // Helper function that creates an external network to be used in other tests
-func (vcd *TestVCD) testCreateExternalNetwork(testName, networkName, dnsSuffix string) (skippingReason string, externalNetwork *types.ExternalNetwork, task Task, err error) {
+func (vcd *TestVCD) testCreateExternalNetwork(testName, networkName, dnsSuffix string) (skippingReason string, externalNetwork *types.ExternalNetworkCreate, task Task, err error) {
 
 	if vcd.skipAdminTests {
 		return fmt.Sprintf(TestRequiresSysAdminPrivileges, testName), externalNetwork, Task{}, nil
@@ -72,11 +72,12 @@ func (vcd *TestVCD) testCreateExternalNetwork(testName, networkName, dnsSuffix s
 		return fmt.Sprintf("More than one port group found with name '%s'", vcd.config.VCD.ExternalNetworkPortGroup), externalNetwork, Task{}, nil
 	}
 
-	externalNetwork = &types.ExternalNetwork{
+	externalNetwork = &types.ExternalNetworkCreate{
 		Name:        networkName,
 		Description: "Test Create External Network",
-		Xmlns:       types.XMLNamespaceExtension,
-		XmlnsVCloud: types.XMLNamespaceVCloud,
+		XmlnsVmext:  types.XMLNamespaceExtension,
+		XmlnsVcload: types.XMLNamespaceVCloud,
+		Type:        types.MimeExternalNetwork,
 		Configuration: &types.NetworkConfiguration{
 			Xmlns: types.XMLNamespaceVCloud,
 			IPScopes: &types.IPScopes{
@@ -98,9 +99,9 @@ func (vcd *TestVCD) testCreateExternalNetwork(testName, networkName, dnsSuffix s
 				}},
 			FenceMode: "isolated",
 		},
-		VimPortGroupRefs: &types.VimObjectRefs{
-			VimObjectRef: []*types.VimObjectRef{
-				&types.VimObjectRef{
+		VimPortGroupRefs: &types.VimObjectRefsCreate{
+			VimObjectRef: []*types.VimObjectRefCreate{
+				&types.VimObjectRefCreate{
 					VimServerRef: &types.Reference{
 						HREF: vimServerHref,
 					},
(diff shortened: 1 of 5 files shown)


types/v56/types.go
```diff
@@ -2317,25 +2328,55 @@ type VimObjectRefs struct {
...
+type VimObjectRefsCreate struct {
+	VimObjectRef []*VimObjectRefCreate `xml:"vmext:VimObjectRef"`
+}
+
+// Type: VMWExternalNetworkType
+// Namespace: http://www.vmware.com/vcloud/extension/v1.5
+// https://vdc-repo.vmware.com/vmwb-repository/dcr-public/7a028e78-bd37-4a6a-8298-9c26c7eeb9aa/09142237-dd46-4dee-8326-e07212fb63a8/doc/doc/types/VMWExternalNetworkType.html
+// Description: External network type.
+// Since: 1.0
+type ExternalNetworkCreate struct {
```
dataclouder: Why do we need two structures for the same entity? Do we have a similar need for other entities?

types/v56/types.go
```diff
@@ -2317,25 +2328,55 @@ type VimObjectRefs struct {
...
 	OperationKey     string                `xml:"operationKey,attr,omitempty"`
 	Name             string                `xml:"name,attr"`
-	Description      string                `xml:"vcloud:Description,omitempty"`
-	Configuration    *NetworkConfiguration `xml:"Configuration,omitempty"`
 	Link             []*Link               `xml:"Link,omitempty"`
-	VimPortGroupRefs *VimObjectRefs        `xml:"VimPortGroupRefs,omitempty"`
+	Description      string                `xml:"Description,omitempty"`
 	Tasks            *TasksInProgress      `xml:"Tasks,omitempty"`
+	Configuration    *NetworkConfiguration `xml:"Configuration,omitempty"`
+	VimPortGroupRef  *VimObjectRef         `xml:"VimPortGroupRef,omitempty"`
```
dataclouder: Why two properties of the same type? what's  in `VimPortGroupRef` and what in `VimPortGroupRefs`?

types/v56/types.go
vbauzysvmware: cause go still don't handle namespaces and I need them as create use two of them. Also get returns different look of xml in regards of namespaces. If you know better way, let me know

types/v56/types.go
vbauzysvmware: Info is here https://code.vmware.com/apis/287/vcloud#/doc/doc/landing-user_operations.html

types/v56/types.go
Didainius: Are you sure the same structure (old or new) cannot be used in both places?
This makes it  complicated.

vbauzysvmware: Please find solution better, I already spent a day figure out one working solution. In general go don't work with namespaces - request hangs up for 2 years minimum

types/v56/types.go
```diff
@@ -2317,25 +2328,55 @@ type VimObjectRefs struct {
...
+
+// Type: VMWExternalNetworkType
+// Namespace: http://www.vmware.com/vcloud/extension/v1.5
+// https://vdc-repo.vmware.com/vmwb-repository/dcr-public/7a028e78-bd37-4a6a-8298-9c26c7eeb9aa/09142237-dd46-4dee-8326-e07212fb63a8/doc/doc/types/VMWExternalNetworkType.html
+// Description: External network type.
+// Since: 1.0
+type ExternalNetworkCreate struct {
+	XMLName          xml.Name              `xml:"vmext:VMWExternalNetwork"`
+	XmlnsVmext       string                `xml:"xmlns:vmext,attr,omitempty"`
+	XmlnsVcload      string                `xml:"xmlns:vcloud,attr,omitempty"`
```
Didainius: Typo.
```suggestion
	XmlnsVcloud      string                `xml:"xmlns:vcloud,attr,omitempty"`
```

vbauzysvmware: Fixed

types/v56/types.go
```diff
@@ -2317,25 +2328,56 @@ type VimObjectRefs struct {
...
+type VimObjectRefsCreate struct {
+	VimObjectRef []*VimObjectRefCreate `xml:"vmext:VimObjectRef"`
+}
+
+// Type: VMWExternalNetworkType
+// Namespace: http://www.vmware.com/vcloud/extension/v1.5
+// https://vdc-repo.vmware.com/vmwb-repository/dcr-public/7a028e78-bd37-4a6a-8298-9c26c7eeb9aa/09142237-dd46-4dee-8326-e07212fb63a8/doc/doc/types/VMWExternalNetworkType.html
+// Description: External network type.
+// Since: 1.0
+type ExternalNetworkCreate struct {
```
Didainius: This struct can now become private.

govcd/externalnetwork_test.go
```diff
@@ -193,6 +191,11 @@ func (vcd *TestVCD) Test_CreateExternalNetwork(check *C) {
 	check.Assert(ipRange.EndAddress, Equals, "192.168.201.250")
 
 	check.Assert(newExternalNetwork.ExternalNetwork.Configuration.FenceMode, Equals, "isolated")
+	check.Assert(newExternalNetwork.ExternalNetwork.Description, Equals, "Test Create External Network")
+	check.Assert(newExternalNetwork.ExternalNetwork.VimPortGroupRef, NotNil)
+	check.Assert(newExternalNetwork.ExternalNetwork.VimPortGroupRef.VimObjectType, Equals, externalNetwork.VimPortGroupRefs.VimObjectRef[0].VimObjectType)
+	check.Assert(newExternalNetwork.ExternalNetwork.VimPortGroupRef.MoRef, Equals, externalNetwork.VimPortGroupRefs.VimObjectRef[0].MoRef)
+	check.Assert(newExternalNetwork.ExternalNetwork.VimPortGroupRef.VimServerRef.HREF, Equals, externalNetwork.VimPortGroupRefs.VimObjectRef[0].VimServerRef.HREF)
```
Didainius: Wouldn't it be a safer and more future-proof check if the below was used
```go
check.Assert(newExternalNetwork, DeepEquals, externalNetwork)
```

Note. You may need to "mask" (set for equal values) for just a few fields like XMLnamespace or some others to make it fully work.

types/v56/types.go
```diff
@@ -2308,6 +2308,17 @@ type VimObjectRef struct {
 	VimObjectType string     `xml:"VimObjectType"`
 }
 
+// Type: VimObjectRefType
+// Namespace: http://www.vmware.com/vcloud/extension/v1.5
+// https://vdc-repo.vmware.com/vmwb-repository/dcr-public/7a028e78-bd37-4a6a-8298-9c26c7eeb9aa/09142237-dd46-4dee-8326-e07212fb63a8/doc/doc/types/VimObjectRefsType.html
+// Description: Represents the Managed Object Reference (MoRef) and the type of a vSphere object.
+// Since: 0.9
+type VimObjectRefCreate struct {
```
Didainius: Could become private now.

types/v56/types.go
vbauzysvmware: Then I need to move it to another go file

types/v56/types.go
vbauzysvmware: Then I need to move it to another go file

vbauzysvmware: moved

govcd/externalnetwork_test.go
vbauzysvmware: to much diff between objects, impossible to use

types/v56/types.go
vbauzysvmware: moved

Didainius:
And I keep forgetting. Worth mentioning in CHANGELOG

vbauzysvmware:
> And I keep forgetting. Worth mentioning in CHANGELOG

added

govcd/system.go
```diff
@@ -479,17 +480,91 @@ func GetExternalNetwork(vcdClient *VCDClient, networkName string) (*ExternalNetw
...
+		Link             []*types.Link               `xml:"Link,omitempty"`
+		Description      string                      `xml:"vcloud:Description,omitempty"`
+		Tasks            *types.TasksInProgress      `xml:"Tasks,omitempty"`
+		Configuration    *types.NetworkConfiguration `xml:"vcloud:Configuration,omitempty"`
+		VimPortGroupRef  *vimObjectRefCreate         `xml:"VimPortGroupRef,omitempty"`
+		VimPortGroupRefs *vimObjectRefsCreate        `xml:"vmext:VimPortGroupRefs,omitempty"`
+		VCloudExtension  *types.VCloudExtension      `xml:"VCloudExtension,omitempty"`
+	}
+
+	externalNetwork := &externalNetworkCreate{}
```
lvirbalas: I request for a big comment explaining what is happening here and, most importantly, answering the question "Why!?"

vbauzysvmware: Create needs to use two namespaces and go don't work with them, so namespaces are hardcoded. Read functionality needs seperate struct as API returns different look of xml from create in regards of namespaces(prefix)

lvirbalas: Thanks for explanation, only my request was to add a comment to code :)

govcd/externalnetwork_test.go
```diff
@@ -193,6 +191,11 @@ func (vcd *TestVCD) Test_CreateExternalNetwork(check *C) {
 	check.Assert(ipRange.EndAddress, Equals, "192.168.201.250")
 
 	check.Assert(newExternalNetwork.ExternalNetwork.Configuration.FenceMode, Equals, "isolated")
+	check.Assert(newExternalNetwork.ExternalNetwork.Description, Equals, "Test Create External Network")
```
lvirbalas: @dataclouder Do I understand correctly that this PR is trying to solve an issue you hit where a change through UI in description of external network doesn't get noticed by Terraform? If yes:
1. Can you confirm that code from this PR solved it?
2. I believe we need a test case here which checks that after updating description it gets read correctly. I.e. not only that it is set correctly (which works even without this PR).

All in all, I'm asking for a test case which tests what the PR is fixing.

dataclouder: No. The issue was that, using the previous data structure, some fields in the external network were not converted to the Go structure when retrieved.
The creation was correct: it was setting everything the way we expected, but when retrieving the data, fields like description and vcenter details were not retrieved, although the external network had them set correctly.

Without this fix, neither data source or importing would work on the Terraform side.

There is already a test that checks for all these details. I can add an extra test and propose it for inclusion in this PR on Monday morning.

dataclouder: Upon further inspection, I see that the existing test already checks for everything we need.

govcd/system.go
vbauzysvmware: Added

govcd/externalnetwork_test.go
lvirbalas: Does this PR fix the issue where a change through UI in description of external network doesn't get read by Terraform? If yes, can we add a test case which checks whether such background change gets read afterwards correctly?

dataclouder: The PR fix the issue that was reported. 
The problem was that description was created correctly, but never retrieved, regardless of whether it was modified using API or UI.

lvirbalas: Synced with Vaidotas. As external network doesn't have update function yet, the test I was asking for is not timely to implement yet. On the other hand, he confirmed that the new asserts that are added below fail with old code and succeed with new code, so it's enough for now.

PR #231, opened 2019-08-20
Add handling of timeout

vbauzysvmware:
When network/env isn't stable sometime response are lost. Standard http go behaviour is wait for it for unlimited time. We saw that from time to time in cassini when using terraform tests that terraform hang ups.

Solutions is introduce configurable timeout when create client. This allows to avoid hang ups and throw time out errors.

Verified with local env - hangups doesn't happen.
Verified with cassini evn - hangups happen and timeout throws error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ IMPROVEMENTS:
 
 * Move methods for `AdminOrg`, `AdminCatalog`, `AdminVdc` to new files `adminorg.go`,
  `admincatalog.go`, `adminvdc.go`.
+* Added default value for http timeout(600s) which is configurable
 
 BUGS FIXED:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ IMPROVEMENTS:
 
 * Move methods for `AdminOrg`, `AdminCatalog`, `AdminVdc` to new files `adminorg.go`,
  `admincatalog.go`, `adminvdc.go`.
-* Added default value for http timeout(600s) which is configurable
+* Added default value for HTTP timeout (600s) which is configurable
 
 BUGS FIXED:
 
--- a/govcd/api_vcd.go
+++ b/govcd/api_vcd.go
@@ -88,7 +88,7 @@ func (vcdCli *VCDClient) vcdauthorize(user, pass, org string) error {
 
 // NewVCDClient initializes VMware vCloud Director client with reasonable defaults.
 // It accepts functions of type VCDClientOption for adjusting defaults.
-func NewVCDClient(vcdEndpoint url.URL, insecure bool, timeout int, options ...VCDClientOption) *VCDClient {
+func NewVCDClient(vcdEndpoint url.URL, insecure bool, options ...VCDClientOption) *VCDClient {
 	// Setting defaults
 	vcdClient := &VCDClient{
 		Client: Client{
@@ -102,7 +102,7 @@ func NewVCDClient(vcdEndpoint url.URL, insecure bool, timeout int, options ...VC
 					Proxy:               http.ProxyFromEnvironment,
 					TLSHandshakeTimeout: 120 * time.Second,
 				},
-				Timeout: time.Duration(timeout) * time.Second,
+				Timeout: 600 * time.Second,
 			},
 			MaxRetryTimeout: 60, // Default timeout in seconds for Client
 		},
@@ -165,6 +165,14 @@ func WithMaxRetryTimeout(timeoutSeconds int) VCDClientOption {
 func WithAPIVersion(version string) VCDClientOption {
 	return func(vcdClient *VCDClient) error {
 		vcdClient.Client.APIVersion = version
+		return nil
+	}
+}
+
+// WithTimeout allows to override default http timeout
+func WithHttpTimeout(timeout int64) VCDClientOption {
+	return func(vcdClient *VCDClient) error {
+		vcdClient.Client.Http.Timeout = time.Duration(timeout) * time.Second
 		return nil
 	}
 }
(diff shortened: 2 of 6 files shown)


govcd/api_vcd.go
```diff
@@ -88,7 +88,7 @@ func (vcdCli *VCDClient) vcdauthorize(user, pass, org string) error {
 
 // NewVCDClient initializes VMware vCloud Director client with reasonable defaults.
 // It accepts functions of type VCDClientOption for adjusting defaults.
-func NewVCDClient(vcdEndpoint url.URL, insecure bool, options ...VCDClientOption) *VCDClient {
+func NewVCDClient(vcdEndpoint url.URL, insecure bool, timeout int, options ...VCDClientOption) *VCDClient {
```
Didainius: I would propose to set a reasonable default timeout (probably the 600 you're aiming for) and create a `WithTimeout(timeoutSeconds int) VCDClientOption`
similar to `func WithMaxRetryTimeout(timeoutSeconds int) VCDClientOption {`

That way you wouldn't break the `func NewVCDClient(` signature.

vbauzysvmware: done

govcd/api_vcd.go
```diff
@@ -167,3 +168,11 @@ func WithAPIVersion(version string) VCDClientOption {
 		return nil
 	}
 }
+
+// WithTimeout allows to override default http timeout
```
Didainius:
```suggestion
// WithHttpTimeout allows to override default http timeout
```

govcd/sample_govcd_test_config.json
```diff
@@ -7,7 +7,8 @@
 		"sysOrg": "System",
 		"//": "(Optional) In some cases the vCloud Director SDK must wait. It can be",
 		"//": "customized using the below var. If it is not set SDK assumes a default value.",
-		"maxRetryTimeout": 60
+		"maxRetryTimeout": 60,
+    "httpTimeout" : 600
```
Didainius: Looks to be misaligned.

govcd/sample_govcd_test_config.json
```diff
@@ -21,7 +22,7 @@
 			"name": "mycat",
 			"//": "One item in the catalog. It will be used to compose test vApps. Some tests",
 			"//": "rely on it being Photon OS. It is is not Photon OS - the tests will be skipped",
-			
+
```
Didainius: Not sure if there is a reason for this empty line.

govcd/sample_govcd_test_config.json
vbauzysvmware: in UI shows normally

govcd/api_vcd.go
vbauzysvmware: fixed

govcd/sample_govcd_test_config.json
vbauzysvmware: fixed

govcd/api_vcd.go
```diff
@@ -102,6 +102,7 @@ func NewVCDClient(vcdEndpoint url.URL, insecure bool, options ...VCDClientOption
 					Proxy:               http.ProxyFromEnvironment,
 					TLSHandshakeTimeout: 120 * time.Second,
 				},
+				Timeout: 600 * time.Second,
```
dataclouder: In the space of three lines we have three timeouts, all without comments. I think we should document all three of them.

govcd/sample_govcd_test_config.json
dataclouder: This line uses spaces, while the others use tabs

govcd/api_vcd.go
vbauzysvmware: Added

govcd/sample_govcd_test_config.json
vbauzysvmware: image
now ok

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ IMPROVEMENTS:
 
 * Move methods for `AdminOrg`, `AdminCatalog`, `AdminVdc` to new files `adminorg.go`,
  `admincatalog.go`, `adminvdc.go`.
+* Added default value for http timeout(600s) which is configurable
```
lvirbalas: "http timeout(600s)" -> "HTTP timeout (600s)"

vbauzysvmware:
> One polishing ask, and conditional LGTM:
> 
>     * I assume that 600s does _not_ apply to uploading OVAs? It's very easy to spend more than 10 mins uploading? If it doesn't, feel free to merge.

It's applying for one of upload requests and requests are divided, which is configured by upload size

jdavidson2021:
Do you know if this change was ever picked up by the Terraform VCD Provider?  We are still having this catalog timeout issue on Terraform VCD Provider 3.6.  It looks like they are using the go-vcloud-director 2.1 if I am tracing this down correctly.  I see in the go-vcloud-director release notes that this fix wasn't introduced until go-vcloud-director 2.4.0.

jdavidson2021:
> Do you know if this change was ever picked up by the Terraform VCD Provider? We are still having this catalog timeout issue on Terraform VCD Provider 3.6. It looks like they are using the go-vcloud-director 2.1 if I am tracing this down correctly. I see in the go-vcloud-director release notes that this fix wasn't introduced until go-vcloud-director 2.4.0.

Looks like the Terraform VCD Provider 3.6 is using go-vcloud-director 2.15.0, but we still have this catalog "still creating" issue and it isn't erroring out after the default 600s timeout.

Still creating... [2h39m21s elapsed]

vbauzysvmware:
Hi @jdavidson2021 

VCD provider version 3.6 has the latest govcd. And timeout was added a long time ago. By default http response timeout is 600 seconds. 

Yet in your case I assume is not HTTP timeout issue. What you see is that VCD task is running and the provider is waiting for it to finish by fetching it in a repetitive way. To tell you more, we need more information, as currently, it's my assumption with the information I have at the moment.

PR #232, opened 2019-08-21
Fix for removing shared catalogs when removing org.

vbauzysvmware:
Ref: https://github.com/terraform-providers/terraform-provider-vcd/issues/312

Change which validates that catalog is from same organization and valid to delete. Issue is with API return structure, where shared catalogs are showed as own with no distinction.

Couldn't make verify test as we don't have functionality yet to create shared folders. Verified manually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ BUGS FIXED:
 
 * Fix bug in AdminOrg.Update, where OrgGeneralSettings would not update correctly if it contained only one property
 * Fix bug in External network creation and get when description wasn't populated.
+* Fix bug in Org delete where would delete other organization shared catalogs.
 
 ## 2.3.1 (Jul 29, 2019)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ BUGS FIXED:
 
 * Fix bug in AdminOrg.Update, where OrgGeneralSettings would not update correctly if it contained only one property
 * Fix bug in External network creation and get when description wasn't populated.
-* Fix bug in Org delete where would delete other organization shared catalogs.
+* Fix bug in Org Delete, which would remove catalogs shared from other organizations.
 
 ## 2.3.1 (Jul 29, 2019)
 
--- a/govcd/adminorg.go
+++ b/govcd/adminorg.go
@@ -389,7 +389,7 @@ func (adminOrg *AdminOrg) removeCatalogs() error {
 // isCatalogFromSameOrg checks if catalog is in same Org. Shared catalogs form other Org are showed as normal one
 // in some API responses.
 func isCatalogFromSameOrg(adminOrg *AdminOrg, catalogName string) (bool, error) {
-	foundCatalogs, err := adminOrg.FinAdminCatalogByName(catalogName, adminOrg.AdminOrg.Name)
+	foundCatalogs, err := adminOrg.FindCatalogRecordTypes(catalogName)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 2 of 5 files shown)


govcd/adminorg.go
```diff
@@ -362,23 +363,58 @@ func (adminOrg *AdminOrg) removeAllOrgNetworks() error {
...
+	}
+
+	if len(foundCatalogs) == 1 {
+		return true, nil
+	}
+	return false, nil
+}
+
+// FinCatalogByName uses catalog name and Org name to return CatalogRecordType information.
+func (adminOrg *AdminOrg) FinAdminCatalogByName(name, orgName string) ([]*types.CatalogRecordType, error) {
```
dataclouder: Three problems:
1. The method name is misspelled. It should be `Find`, not `Fin`.
2. According to our conventions, the method should indicate what object is retrieving. Not `FindAdminCatalog` but`FindCatalogRecordTypes` (I believe we can skip the `ByName` part)
3. Why do we pass orgName? Isn't the org name already in the receiver `adminOrg` data?

govcd/adminorg_test.go
```diff
@@ -0,0 +1,38 @@
...
+ */
+package govcd
+
+import (
+	"fmt"
+	. "gopkg.in/check.v1"
+)
+
+// Creates a Catalog and then verify that finds it
+func (vcd *TestVCD) Test_Client_FinAdminCatalogByName(check *C) {
```
dataclouder: Change function name to reflect suggestion about main function above

govcd/query.go
```diff
@@ -31,6 +31,16 @@ func (vcdCli *VCDClient) Query(params map[string]string) (Results, error) {
 	return getResult(&vcdCli.Client, req)
 }
 
+func (client *Client) QueryWithNotEncodedParams(params map[string]string, notEncodedParams map[string]string) (Results, error) {
```
dataclouder: Missing description.
What's the difference between this function and the one a few lines below with the same name but different receiver?
(`func (vcdCli *VCDClient) QueryWithNotEncodedParams(params map[string]string, notEncodedParams map[string]string)` )

govcd/adminorg_test.go
```diff
@@ -0,0 +1,38 @@
...
+	// just imitate wait
+	err = adminOrg.Refresh()
+	check.Assert(err, IsNil)
+
+	findRecords, err := adminOrg.FinAdminCatalogByName(catalogName, adminOrg.AdminOrg.Name)
+	check.Assert(err, IsNil)
+	check.Assert(findRecords, NotNil)
+	check.Assert(len(findRecords), Equals, 1)
+	check.Assert(findRecords[0].Name, Equals, catalogName)
+	check.Assert(findRecords[0].OrgName, Equals, adminOrg.AdminOrg.Name)
```
dataclouder: This test proves that the function finds a catalog, but not that the catalog is shared, which is the point of this PR.
Is there a way of testing the shared catalog scenario ?

govcd/query.go
vbauzysvmware: Yes, for convenience

govcd/adminorg_test.go
vbauzysvmware: I don't care about it shared, I care about if catalog belongs same ORG. Fully verify unit test would create two orgs, catalogs and make one of shared for other Orgs. And when delete  Org and expects it didn't touch catalog not belonged to other org.

govcd/adminorg_test.go
vbauzysvmware: fixed

govcd/adminorg.go
vbauzysvmware: fixed

govcd/adminorg_test.go
dataclouder: What I mean is: do we have a way of testing a negative finding? Perhaps using one of the tests where we create other catalogs.

vbauzysvmware: we can't create shared catalog, we can only find not existing one

govcd/query.go
Didainius: Can one wrap another if it is a convenience function?

govcd/adminorg.go
```diff
@@ -362,23 +367,58 @@ func (adminOrg *AdminOrg) removeAllOrgNetworks() error {
...
-	for _, catalogs := range adminOrg.AdminOrg.Catalogs.Catalog {
-		// Get Catalog HREF
-		catalogHREF := adminOrg.client.VCDHREF
-		catalogHREF.Path += "/admin/catalog/" + strings.Split(catalogs.HREF, "/api/admin/catalog/")[1] //gets id
-		req := adminOrg.client.NewRequest(map[string]string{
-			"force":     "true",
-			"recursive": "true",
-		}, http.MethodDelete, catalogHREF, nil)
-		_, err := checkResp(adminOrg.client.Http.Do(req))
+	for _, catalog := range adminOrg.AdminOrg.Catalogs.Catalog {
```
Didainius: Could you fix the comment for this function. It's not valid godoc

govcd/adminorg.go
```diff
@@ -362,23 +367,58 @@ func (adminOrg *AdminOrg) removeAllOrgNetworks() error {
...
+		return false, err
+	}
+
+	if len(foundCatalogs) == 1 {
+		return true, nil
+	}
+	return false, nil
+}
+
+// FinCatalogByName uses catalog name and Org name to return AdminCatalogRecord information.
```
Didainius: Invalid comment.

vbauzysvmware: fixed

govcd/adminorg.go
vbauzysvmware: fixed

govcd/query.go
vbauzysvmware: done

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ BUGS FIXED:
 
 * Fix bug in AdminOrg.Update, where OrgGeneralSettings would not update correctly if it contained only one property
 * Fix bug in External network creation and get when description wasn't populated.
+* Fix bug in Org delete where would delete other organization shared catalogs.
```
dataclouder: "Org delete where would delete other organization shared catalogs" -> "Org Delete, which would delete catalogs shared from other organizations"

PR #233, opened 2019-08-23
Escape query API calls to handle special characters correctly

vbauzysvmware:
Ref: https://github.com/terraform-providers/terraform-provider-vcd/issues/306

Fix issue within queries, where name wasn't escaped and result would throw API errors.


CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ BUGS FIXED:
 
 * Fix bug in AdminOrg.Update, where OrgGeneralSettings would not update correctly if it contained only one property
 * Fix bug in External network creation and get when description wasn't populated.
+* Fix bug in VDC creation when name with space throw error
```
lvirbalas: "throw error" -> "caused an error"

PR #234, opened 2019-08-26
Improve attach tests

vbauzysvmware:
This improves attach media tests.
Issue we got on VCD 10 version quite often is that checking state of detached media was updated with lag. So added a few refresh for test not to fail.

### DIFF
--- a/govcd/vm_test.go
+++ b/govcd/vm_test.go
@@ -495,7 +495,6 @@ func (vcd *TestVCD) Test_HandleInsertOrEjectMedia(check *C) {
 
 	//verify
 	vcd.waitUntilEjected(err, vm, check)
-
 	check.Assert(isMediaInjected(vm.VM.VirtualHardwareSection.Item), Equals, false)
 }
 
@@ -584,13 +583,7 @@ func (vcd *TestVCD) Test_InsertOrEjectMedia(check *C) {
 
 	//verify
 	// VCD is slow to change representative state
-	for i := 1; i < 5; i++ {
-		err = vm.Refresh()
-		check.Assert(err, IsNil)
-		if isMediaInjected(vm.VM.VirtualHardwareSection.Item) == false {
-			break
-		}
-	}
+	vcd.waitUntilEjected(err, vm, check)
 	check.Assert(isMediaInjected(vm.VM.VirtualHardwareSection.Item), Equals, false)
 }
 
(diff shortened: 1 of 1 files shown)


vbauzysvmware:
> Shouldn't this be solved in actual SDK code, not tests?

Please elaborate

Didainius:
I mean the current function `func (vcd *TestVCD) waitUntilEjected(err error, vm *VM, check *C)` is created and used for tests to handle such scenario. How does one solve this problem using our SDK (not in tests)?

vbauzysvmware:
It doesn't solve any SDK issue - is another question/discussion if there is any.

govcd/vm_test.go
```diff
@@ -494,11 +494,22 @@ func (vcd *TestVCD) Test_HandleInsertOrEjectMedia(check *C) {
...
 
 	//verify
-	err = vm.Refresh()
-	check.Assert(err, IsNil)
+	vcd.waitUntilEjected(err, vm, check)
 	check.Assert(isMediaInjected(vm.VM.VirtualHardwareSection.Item), Equals, false)
 }
 
+// waitUntilEjected checks status until ready. We need this cause VCD is slow to change representative state
+func (vcd *TestVCD) waitUntilEjected(err error, vm *VM, check *C) {
```
dataclouder: While this function makes the test pass, it doesn't solve the problem.
The problem is that with some recent vCD versions, the refresh doesn't return quickly enough.
Tests should certify that the SDK woks as expected. If they don't, we should fix the root cause, not change the test to adapt it to the symptoms/

govcd/vm_test.go
```diff
@@ -494,11 +494,22 @@ func (vcd *TestVCD) Test_HandleInsertOrEjectMedia(check *C) {
 	check.Assert(err, IsNil)
 
 	//verify
-	err = vm.Refresh()
-	check.Assert(err, IsNil)
+	vcd.waitUntilEjected(err, vm, check)
 	check.Assert(isMediaInjected(vm.VM.VirtualHardwareSection.Item), Equals, false)
 }
 
+// waitUntilEjected checks status until ready. We need this cause VCD is slow to change representative state
```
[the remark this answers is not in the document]
vbauzysvmware: fixed

govcd/vm_test.go
vbauzysvmware: Well, issue what refresh return to quickly enough.
In general I think it's VCD bug and we should report it.
Workaround in SDK would be possible if we in detach function would fetch refresh untill VCD database updated.

lvirbalas:
p.s. Please also stop using short abbreviated commit messages, a reminder that we have an agreement to use the following guide for commit messages:
https://chris.beams.io/posts/git-commit/

vbauzysvmware:
> p.s. Please also stop using short abbreviated commit messages, a reminder that we have an agreement to use the following guide for commit messages:
> https://chris.beams.io/posts/git-commit/

Where I did that?

PR #236, opened 2019-08-26
Add helper function which handles eject with refresh

vbauzysvmware:
Ref: Issue we got on VCD 10 version quite often is that checking state of detached media was updated with lag. This issue comes only in unit tests due how tests verify eject of media. In terraform we don't read state after eject of media.

I added helper function which helps handling such case. Though current test work with lower level of abstraction function and couldn't be fixed in SDK level, only on unit test level when we need.

### DIFF
--- a/govcd/vm.go
+++ b/govcd/vm.go
@@ -555,7 +555,8 @@ func (vm *VM) HandleInsertMedia(org *Org, catalogName, mediaName string) (Task, 
 }
 
 // Helper function which finds media, calls EjectMedia, waits for task to complete and answer question.
-// Also waits until vm status refreshes - this added as current VCD version has lag in status update.
+// Also waits until VM status refreshes - this added as current vCD version has lag in status update.
+// answerYes - handles question risen when VM is running. True value enforces ejection.
 func (vm *VM) HandleEjectMediaAndAnswer(org *Org, catalogName, mediaName string, answerYes bool) (*VM, error) {
 	task, err := vm.HandleEjectMedia(org, catalogName, mediaName)
 	if err != nil {
--- a/govcd/vm.go
+++ b/govcd/vm.go
@@ -555,7 +555,7 @@ func (vm *VM) HandleInsertMedia(org *Org, catalogName, mediaName string) (Task, 
 }
 
 // HandleEjectMediaAndAnswer helper function which finds media, calls EjectMedia, waits for task to complete and answer question.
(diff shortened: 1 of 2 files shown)


govcd/vm.go
```diff
@@ -554,6 +554,43 @@ func (vm *VM) HandleInsertMedia(org *Org, catalogName, mediaName string) (Task,
 	})
 }
 
+// Helper function which finds media, calls EjectMedia, waits for task to complete and answer question.
+// Also waits until vm status refreshes - this added as current VCD version has lag in status update.
```
lvirbalas: 1.) Description of what `answerYes` does needed.

2.) Proper capitalisation needed:
"vm" -> "VM"
"VCD" -> "vCD"

govcd/vm.go
```diff
@@ -554,6 +554,43 @@ func (vm *VM) HandleInsertMedia(org *Org, catalogName, mediaName string) (Task,
 	})
 }
 
+// Helper function which finds media, calls EjectMedia, waits for task to complete and answer question.
```
Didainius: Valid go doc comment begins with function name

govcd/vm.go
vbauzysvmware: fixed

govcd/vm.go
```diff
@@ -554,6 +554,44 @@ func (vm *VM) HandleInsertMedia(org *Org, catalogName, mediaName string) (Task,
...
+// Also waits until VM status refreshes - this added as current vCD version has lag in status update.
```
Didainius: The wording "current vCD version" will become unclear after a few months.

govcd/vm.go
vbauzysvmware: fixed

govcd/vm.go
vbauzysvmware: fixed

govcd/vm.go
```diff
@@ -554,6 +554,45 @@ func (vm *VM) HandleInsertMedia(org *Org, catalogName, mediaName string) (Task,
 	})
 }
 
+// Helper function which finds media, calls EjectMedia, waits for task to complete and answer question.
+// Also waits until vm status refreshes - this added as current VCD version has lag in status update.
```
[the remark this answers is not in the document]
vbauzysvmware: changed with exact versions.

PR #237, opened 2019-08-29
Clean up lint errors

Didainius:
This repo contains quite a few lint reported problems. This PR has 0 impact on functionality, but cleans up quite a few lint errors.

Both govcd and Terraform tests pass:

This is the list of fixed lint errors:
```bash
govcd/catalog_test.go:188:31: Error return value of `uploadTask.ShowUploadProgress` is not checked (errcheck)
govcd/catalog_test.go:376:31: Error return value of `uploadTask.ShowUploadProgress` is not checked (errcheck)
govcd/lb_test.go:72:23: Error return value of `vapp.BlockWhileStatus` is not checked (errcheck)
govcd/media_test.go:78:31: Error return value of `uploadTask.ShowUploadProgress` is not checked (errcheck)
govcd/media_test.go:190:19: Error return value of `mediaItem.Refresh` is not checked (errcheck)
govcd/vdc_test.go:162:23: Error return value of `vapp.BlockWhileStatus` is not checked (errcheck)
govcd/disk.go:99:36: sloppyLen: len(disk.Disk.Tasks.Task) <= 0 can be len(disk.Disk.Tasks.Task) == 0 (gocritic)
govcd/entity_test.go:131:15: S1025: should use String() instead of fmt.Sprintf (gosimple)
govcd/entity_test.go:147:15: S1025: should use String() instead of fmt.Sprintf (gosimple)
govcd/entity_test.go:159:15: S1025: should use String() instead of fmt.Sprintf (gosimple)
govcd/lb_test.go:148:2: S1023: redundant `return` statement (gosimple)
govcd/lb_test.go:348:13: S1005: should omit value from range; this loop is equivalent to `for index := range ...` (gosimple)
govcd/vm.go:592:6: S1002: should omit comparison to bool constant, can be simplified to `!isMediaInjected(vm.VM.VirtualHardwareSection.Item)` (gosimple)
types/v56/types.go:1182:2: S1023: redundant `return` statement (gosimple)
util/logging_test.go:16:2: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (gosimple)
util/tar.go:128:29: SA1024: cutset contains duplicate characters (staticcheck)
```

### DIFF
--- a/util/tar.go
+++ b/util/tar.go
@@ -125,7 +125,7 @@ func sanitizedName(filename string) string {
 		filename = filename[2:]
 	}
 	filename = strings.TrimLeft(filename, "\\/.")
-	filename = strings.TrimLeft(filename, "../")
+	filename = strings.TrimLeft(filename, "./")
 	filename = strings.Replace(filename, "../../", "../", -1)
 	return strings.Replace(filename, "..\\", "", -1)
 }
--- a/govcd/catalog_test.go
+++ b/govcd/catalog_test.go
@@ -185,8 +185,8 @@ func (vcd *TestVCD) Test_UploadOvf_ShowUploadProgress_works(check *C) {
 	r, w, _ := os.Pipe()
 	os.Stdout = w
 
-	uploadTask.ShowUploadProgress()
-
+	err = uploadTask.ShowUploadProgress()
+	check.Assert(err, IsNil)
 	w.Close()
 	//read stdin
 	result, _ := ioutil.ReadAll(r)
@@ -373,8 +373,8 @@ func (vcd *TestVCD) Test_CatalogUploadMediaImage_ShowUploadProgress_works(check 
 	r, w, _ := os.Pipe()
 	os.Stdout = w
 
-	uploadTask.ShowUploadProgress()
-
+	err = uploadTask.ShowUploadProgress()
+	check.Assert(err, IsNil)
 	w.Close()
 	//read stdin
 	result, _ := ioutil.ReadAll(r)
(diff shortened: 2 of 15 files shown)


util/tar.go
```diff
@@ -125,7 +125,7 @@ func sanitizedName(filename string) string {
 		filename = filename[2:]
 	}
 	filename = strings.TrimLeft(filename, "\\/.")
-	filename = strings.TrimLeft(filename, "../")
+	filename = strings.TrimLeft(filename, "./")
```
lvirbalas: Out of all these improvements, this one looks like it is changing functionality?
a. Old behaviour is "../my.file" -> "my.file"
b. New behaviour, it seems, would leave the string "../my.file" intact

Didainius: I believe it should not as per go doc https://golang.org/pkg/strings/#TrimLeft

> TrimLeft returns a slice of the string s with all leading Unicode code points contained in cutset removed. 

It does not matter how many characters (of the same name) are in _cutset_. It would clean all characters (unicode code points) in the cutset.

Here is the official example from godoc. https://play.golang.org/p/LBj5Z_hQpMT
`fmt.Print(strings.TrimLeft("¡¡¡Hello, Gophers!!!", "!¡"))`
gives output (despite only one "!" char listed):
`Hello, Gophers!!!`

Here are also some random example from me:
https://play.golang.org/p/6UjTznm3wnk 


I also included your example into tests and it still passes:
$ go test -v -run "TestSanitizedName" .
=== RUN   TestSanitizedName
--- PASS: TestSanitizedName (0.00s)
PASS
ok      github.com/vmware/go-vcloud-director/v2/util    0.015s

lvirbalas: Ergh, that's one trim function definition to trip on (I mean the way it uses cutsets)! Thanks for clarifying.

PR #239, opened 2019-09-06
Add Get* functions for network and edge gateway

dataclouder:
* Add GetEdgeGatewayBy{Name|Id|NameOrId}
* Add GetVdcNetworkBy{Name|Id|NameOrId}
* Deprecate FindVDCNetwork and FindEdgeGateway
* Replace deprecated functions with new methods throughout the library
* Add tests for new Get* methods

This work is preparation for network and edge gateway data sources in terraform-provider-vcd

### DIFF
--- a/govcd/orgvdcnetwork.go
+++ b/govcd/orgvdcnetwork.go
@@ -89,7 +89,7 @@ func (orgVdcNet *OrgVDCNetwork) Delete() (Task, error) {
 
 // Looks for an Org Vdc network and, if found, will delete it.
 func RemoveOrgVdcNetworkIfExists(vdc Vdc, networkName string) error {
-	network, err := vdc.GetVdcNetworkByName(networkName, false)
+	network, err := vdc.GetOrgVdcNetworkByName(networkName, false)
 	if err == nil {
 		task, err := network.Delete()
 		if err != nil {
--- a/govcd/orgvdcnetwork.go
+++ b/govcd/orgvdcnetwork.go
@@ -89,8 +89,8 @@ func (orgVdcNet *OrgVDCNetwork) Delete() (Task, error) {
 
 // Looks for an Org Vdc network and, if found, will delete it.
 func RemoveOrgVdcNetworkIfExists(vdc Vdc, networkName string) error {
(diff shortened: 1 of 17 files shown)


govcd/vdc.go
```diff
@@ -196,6 +197,71 @@ func (vdc *Vdc) FindVDCNetwork(network string) (OrgVDCNetwork, error) {
...
+			}
+		}
+	}
+
+	return nil, ErrorEntityNotFound
+}
+
+// GetVdcNetworkByNameOrId returns a VDC Network reference if either the network name or ID matches an existing one.
+// If no valid external network is found, it returns an empty ExternalNetwork reference and an error
+func (vdc *Vdc) GetVdcNetworkByNameOrId(identifier string, refresh bool) (*OrgVDCNetwork, error) {
```
lvirbalas: Hi @dataclouder , really sorry for confusion - I know we discussed this, but probably it's better to not discuss these kind of things at Friday 4pm... Now that I saw the whole spectrum in PR, to me it looks that these Org network related functions are better to be called `GetOrgVdcNetworkBy...` after all. Though longer, but more intuitive, as you don't need to process in your head what kind of VDC is it referring to.

dataclouder: Renamed

govcd/orgvdcnetwork.go
```diff
@@ -89,8 +89,8 @@ func (orgVdcNet *OrgVDCNetwork) Delete() (Task, error) {
 
 // Looks for an Org Vdc network and, if found, will delete it.
 func RemoveOrgVdcNetworkIfExists(vdc Vdc, networkName string) error {
-	network, err := vdc.FindVDCNetwork(networkName)
-	if err == nil && network != (OrgVDCNetwork{}) {
+	network, err := vdc.GetOrgVdcNetworkByName(networkName, false)
+	if err == nil {
```
Didainius: Wouldn't this "mask" the `err`? If the error was not nil - it would not try to remove and return`nil` I think while in fact there could have been some error during deletion.

dataclouder: Added more error handling, and an explicit test

PR #241, opened 2019-09-11
Add support for NAT rules using proxied NSX-V API

Didainius:
This PR aims to add support for methods to allow configuring Edge Gateway NAT rules using proxied NSX-V API.

It introduces a few new methods: `CreateNsxvNatRule`, `UpdateNsxvNatRule`, `GetNsxvNatRuleById` and `DeleteNsxvNatRuleById`.

The creation of above mentioned NAT rules requires to specify a vNic (network interface on edge gateway). To easen lookup of vNic and replicating similar behaviour as vCD UI two more methods are added: `GetVnicIndexByNetworkNameAndType` and `GetNetworkNameAndTypeByVnicIndex`. They allow to convert between vNic index number and network type and network name.

It should also be compatible with subinterfaces when we implement them in other places: https://github.com/terraform-providers/terraform-provider-vcd/issues/321

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ guest properties [#235](https://github.com/vmware/go-vcloud-director/pull/235)
 * Added methods `vapp.SetProductSectionList` and `vapp.GetProductSectionList` allowing to manipulate
 vApp guest properties [#235](https://github.com/vmware/go-vcloud-director/pull/235)
 * Added method GetStorageProfileByHref
+* Added methods `CreateNsxvNatRule()`, `UpdateNsxvNatRule()`, `GetNsxvNatRuleById()`, `DeleteNsxvNatRuleById()` [#xx]()
+* Added methods `GetVnicIndexFromNetworkNameType()` and `GetNetworkNameTypeFromVnicIndex()` [#xx]()
 
 IMPROVEMENTS:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ vApp guest properties [#235](https://github.com/vmware/go-vcloud-director/pull/2
 * Added method GetStorageProfileByHref
 * Added methods `CreateNsxvNatRule()`, `UpdateNsxvNatRule()`, `GetNsxvNatRuleById()`, `DeleteNsxvNatRuleById()`
 which use the proxied NSX-V API for handling NAT rules [#241](https://github.com/vmware/go-vcloud-director/pull/241)
-* Added methods `GetVnicIndexFromNetworkNameType()` and `GetNetworkNameTypeFromVnicIndex()` [#241](https://github.com/vmware/go-vcloud-director/pull/241)
+* Added methods `GetVnicIndexByNetworkNameAndType()` and `GetNetworkNameAndTypeByVnicIndex()` [#241](https://github.com/vmware/go-vcloud-director/pull/241)
 
 IMPROVEMENTS:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ guest properties [#235](https://github.com/vmware/go-vcloud-director/pull/235)
 vApp guest properties [#235](https://github.com/vmware/go-vcloud-director/pull/235)
 * Added method GetStorageProfileByHref
 * Added methods `CreateNsxvNatRule()`, `UpdateNsxvNatRule()`, `GetNsxvNatRuleById()`, `DeleteNsxvNatRuleById()`
-which use the proxied NSX-V API for handling NAT rules [#241](https://github.com/vmware/go-vcloud-director/pull/241)
+which use the proxied NSX-V API of advanced edge gateway for handling NAT rules [#241](https://github.com/vmware/go-vcloud-director/pull/241)
 * Added methods `GetVnicIndexByNetworkNameAndType()` and `GetNetworkNameAndTypeByVnicIndex()` [#241](https://github.com/vmware/go-vcloud-director/pull/241)
 
 IMPROVEMENTS:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,8 +28,9 @@ guest properties [#235](https://github.com/vmware/go-vcloud-director/pull/235)
 * Added methods `vapp.SetProductSectionList` and `vapp.GetProductSectionList` allowing to manipulate
 vApp guest properties [#235](https://github.com/vmware/go-vcloud-director/pull/235)
 * Added method GetStorageProfileByHref
-* Added methods `CreateNsxvNatRule()`, `UpdateNsxvNatRule()`, `GetNsxvNatRuleById()`, `DeleteNsxvNatRuleById()` [#xx]()
-* Added methods `GetVnicIndexFromNetworkNameType()` and `GetNetworkNameTypeFromVnicIndex()` [#xx]()
+* Added methods `CreateNsxvNatRule()`, `UpdateNsxvNatRule()`, `GetNsxvNatRuleById()`, `DeleteNsxvNatRuleById()`
+which use the proxied NSX-V API for handling NAT rules [#241](https://github.com/vmware/go-vcloud-director/pull/241)
+* Added methods `GetVnicIndexFromNetworkNameType()` and `GetNetworkNameTypeFromVnicIndex()` [#241](https://github.com/vmware/go-vcloud-director/pull/241)
 
 IMPROVEMENTS:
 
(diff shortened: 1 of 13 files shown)


govcd/edgegateway.go
```diff
@@ -1167,3 +1167,131 @@ func validateUpdateLBGeneralParams(logLevel string) error {
...
+		return nil, err
+	}
+
+	return vnicConfig, nil
+}
+
+// GetVnicIndexFromNetworkNameType returns *int of vNic index for specified network name and network type
+// networkType one of: 'internal', 'uplink', 'trunk', 'subinterface'
+// networkName cannot be empty
+func (egw *EdgeGateway) GetVnicIndexFromNetworkNameType(networkName, networkType string) (*int, error) {
```
lvirbalas: I suggest naming these new methods with "By" instead "From" and with an "And", i.e.: GetVnicIndex**By**NetworkName**And**Type

This would make these names consistent/similar to methods like:
GetCatalog**By**Name**Or**Id

CC: @dataclouder

Didainius: Fixed

Didainius:
>     1. That NAT rule(s) created by this method may be updated via H5 UI and vice versa (that these operations don't collide anymore).
> 
Yes, they can be updated. II have manually tested it using terraform resources and "read" capabilities, which hint updates.

>     2. That it will be possible from Terraform side to create a DNAT rule on an _external_ network with an _org_ admin connection (with current DNAT rule that's not possible though in UI it is).

Yes, it must be so. It does not try to access any other resources outside of edge gateway. If the edge gateway has an external network attached - it uses it.

govcd/nat_test.go
```diff
@@ -0,0 +1,142 @@
...
+// 4. Deletes the rule
+// 5. Validates that the rule was deleted
+func testNsxvNat(natRule *types.EdgeNatRule, vcd *TestVCD, check *C, edge EdgeGateway) {
+	createdNatRule, err := edge.CreateNsxvNatRule(natRule)
+	check.Assert(err, IsNil)
+
+	parentEntity := vcd.org.Org.Name + "|" + vcd.vdc.Vdc.Name + "|" + vcd.config.VCD.EdgeGateway
+	AddToCleanupList(createdNatRule.ID, "nsxvNatRule", parentEntity, check.TestName())
+
+	gotNatRule, err := edge.GetNsxvNatRuleById(createdNatRule.ID)
```
vbauzysvmware: `gotNatRule` missing check it isn't nil

govcd/nat_test.go
```diff
@@ -0,0 +1,142 @@
...
+	check.Assert(err, IsNil)
+	check.Assert(gotNatRule, DeepEquals, createdNatRule)
+	check.Assert(gotNatRule.ID, Equals, createdNatRule.ID)
+
+	// Set ID and update nat rule with description
+	natRule.ID = gotNatRule.ID
+	natRule.Description = "Description for NAT rule"
+	updatedNatRule, err := edge.UpdateNsxvNatRule(natRule)
```
vbauzysvmware: `updateNatRule` missing check it isn't nil

types/v56/types.go
```diff
@@ -1818,6 +1818,27 @@ type LbVirtualServer struct {
...
+// // https://code.vmware.com/docs/6900/vcloud-director-api-for-nsx-programming-guide
+type EdgeNatRule struct {
+	XMLName           xml.Name `xml:"natRule"`
+	ID                string   `xml:"ruleId,omitempty"`
+	RuleType          string   `xml:"ruleType,omitempty"`
+	RuleTag           string   `xml:"ruleTag,omitempty"`
+	Action            string   `xml:"action,omitempty"`
+	Vnic              *int     `xml:"vnic,omitempty"`
+	OriginalAddress   string   `xml:"originalAddress,omitempty"`
+	TranslatedAddress string   `xml:"translatedAddress,omitempty"`
```
vbauzysvmware: Is it really all fields can be empty?

types/v56/types.go
```diff
@@ -2671,3 +2692,52 @@ type AdminCatalogRecord struct {
 	Link                    *Link     `xml:"Link,omitempty"`
 	Vdc                     *Metadata `xml:"Metadata,omitempty"`
 }
+
+// EdgeGatewayVnics is a data structure holding information of vNic configuration in NSX-V edge
+// gateway
+type EdgeGatewayVnics struct {
+	XMLName xml.Name `xml:"vnics"`
```
vbauzysvmware: Are sure that all fields can't be empty?

govcd/nat.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	// Location header should look similar to:
+	// [/network/edges/edge-1/nat/config/rules/197157]
+	natRuleId, err := extractNsxObjectIdFromPath(resp.Header.Get("Location"))
+	if err != nil {
+		return nil, err
+	}
+
+	time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
Didainius: I managed to leave this while using it for some debugging. Thanks. Will remove at all.

types/v56/types.go
Didainius: I only use a couple of them here in "read only" fashion. It never failed but I added tags to majority.

types/v56/types.go
Didainius: Added just a few.

govcd/nat_test.go
Didainius: Done.

govcd/nat_test.go
Didainius: Done.

govcd/nat.go
```diff
@@ -0,0 +1,203 @@
...
+
+// GetNsxvNatRuleById retrieves types.EdgeNatRule by NAT rule ID as shown in the UI using proxied
+// NSX-V API.
+// It returns and error `ErrorEntityNotFound` if the NAT rule is now found.
+func (egw *EdgeGateway) GetNsxvNatRuleById(id string) (*types.EdgeNatRule, error) {
+	return egw.getNsxvNatRule(&types.EdgeNatRule{ID: id})
+}
+
+func (egw *EdgeGateway) deleteNsxvNatRule(natRuleConfig *types.EdgeNatRule) error {
+	err := validateDeleteNsxvNatRule(natRuleConfig, egw)
```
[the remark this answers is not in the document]
Didainius: No more.

govcd/nat.go
```diff
@@ -0,0 +1,203 @@
...
+
+	readNatRule, err := egw.GetNsxvNatRuleById(natRuleConfig.ID)
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve NAT rule with ID (%s) after update: %s",
+			readNatRule.ID, err)
+	}
+	return readNatRule, nil
+}
+
+func (egw *EdgeGateway) getNsxvNatRule(natRuleConfig *types.EdgeNatRule) (*types.EdgeNatRule, error) {
```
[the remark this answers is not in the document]
Didainius: I'm a fan of symmetric functions, removing it now.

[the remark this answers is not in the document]
Didainius: No more.

types/v56/types.go
vbauzysvmware: In general we match API description, which are mandatory, which aren't. It isn't decision from our point of view.

types/v56/types.go
vbauzysvmware: In general we match API description, which are mandatory, which aren't. It isn't decision from our point of view.

Didainius: Yes, but this one is used for read-only purposes and I care only about a few fields. I very much doubt we will ever need a "POST" capability. This is edge gateway network card configuration.

types/v56/types.go
Didainius: Yes, but only these are mandatory. NSX can default everything else.

types/v56/types.go
vbauzysvmware: I find full types.go of previously inserted types were was used minimally which are wrongly matched with required/optional and  as result create unexpected behaviour when you start to use them. Just a tip to avoid that

CHANGELOG.md
```diff
@@ -28,6 +28,9 @@ guest properties [#235](https://github.com/vmware/go-vcloud-director/pull/235)
 * Added methods `vapp.SetProductSectionList` and `vapp.GetProductSectionList` allowing to manipulate
 vApp guest properties [#235](https://github.com/vmware/go-vcloud-director/pull/235)
 * Added method GetStorageProfileByHref
+* Added methods `CreateNsxvNatRule()`, `UpdateNsxvNatRule()`, `GetNsxvNatRuleById()`, `DeleteNsxvNatRuleById()`
```
vbauzysvmware: Maybe comment should state that this for adcanced edge gtw

Didainius: Added

PR #242, opened 2019-09-17
Update OrgNetwork structure

dataclouder:
The reason for this addition (VimPortGroupRef) is that when
we manipulate DHCP records, the API complains about this element
being missed. We don't need to add anything inside VimPortGroupRef,
but it needs to be there or we can't add DHCP definitions.

Note that, before this addition, DHCP was created automatically
using the leftwovers of static IP pool. For example, if the static
IP pool was 192.168.0.2-192.168.0.100, the DHCP pool, was
192.168.0.101-192.168.0.254.

Note that this change doesn't affect go-vcloud-director, as the network operations keeps working as before, but it is vital for Terraform provider, where DHCP settings need to be enforced (even when there are none) because the automatic behavior affects terraform state.

### DIFF
--- a/types/v56/types.go
+++ b/types/v56/types.go
@@ -322,21 +322,22 @@ type InstantiationParams struct {
 // Description: Represents an Org VDC network in the vCloud model.
 // Since: 5.1
 type OrgVDCNetwork struct {
-	XMLName       xml.Name              `xml:"OrgVdcNetwork"`
-	Xmlns         string                `xml:"xmlns,attr,omitempty"`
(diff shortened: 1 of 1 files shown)


dataclouder:
> LGTM and a question:
> 
> > Note that, before this addition, DHCP was created automatically
> > using the leftwovers of static IP pool
> 
> Does it mean that with this VimPortGroupRef added DHCP does _not_ get created automatically?

Only if we pass explicitly an empty DHCP structure when creating. Before this change, such DHCP structure (empty of filled) was rejected if we tried to add it.

PR #245, opened 2019-09-27
Amendments to media datasource

vbauzysvmware:
Changes needed for Media read/data sources.
* new Media and MediaRecord and Get functions for them
* Deprecated MedIaItem and FindMedia* functions


CHANGELOG.md
```diff
@@ -35,6 +35,8 @@ which use the proxied NSX-V API of advanced edge gateway for handling NAT rules
 * Added methods `Client.GetVMByHref` `Vapp.GetVAMByName` and related `GetVMById`, `GetVAMByNameOrId`
 * Deprecated methods `Client.FindVMByHREF`, `Vdc.FindVMByName`, `Vdc.FindVAppByID`, and `Vdc.FindVAppByName`
 * Added methods `Vm.GetGuestCustomizationSection` and `Vm.SetGuestCustomizationSection`  
+* Made method `GetBareEntityUuid` public
```
lvirbalas: Suggestion:
```
* Converted private method `getBareEntityUuid` to a public `GetBareEntityUuid`
```

vbauzysvmware: Added

CHANGELOG.md
```diff
@@ -35,6 +35,8 @@ which use the proxied NSX-V API of advanced edge gateway for handling NAT rules
 * Added methods `Client.GetVMByHref` `Vapp.GetVAMByName` and related `GetVMById`, `GetVAMByNameOrId`
 * Deprecated methods `Client.FindVMByHREF`, `Vdc.FindVMByName`, `Vdc.FindVAppByID`, and `Vdc.FindVAppByName`
 * Added methods `Vm.GetGuestCustomizationSection` and `Vm.SetGuestCustomizationSection`  
+* Converted private method `getBareEntityUuid` to a public `GetBareEntityUuid`
+* Added new method `QueryMediaImage`
```
dataclouder: I see a new Query method, but I was expecting a set of `Get*` methods like we have done in the other entities for which we provide data sources.
I'd like to see uniformity, and have methods
`Catalog.GetMediaImageByName`,  `Catalog.GetMediaImageById`, `Catalog.GetMediaImageByNameOrId`
as wrappers around the current methods.
When I look for a media image, I want to know that there is a method named in a consistent manner that gives me the expected results.

Moreover, we need to deprecate `FindMediaImage` and replace all its usage  everywhere

govcd/entity_test.go
```diff
@@ -71,6 +71,9 @@ func (vapp *VApp) id() string   { return vapp.VApp.ID }
 func (vm *VM) name() string { return vm.VM.Name }
 func (vm *VM) id() string   { return vm.VM.ID }
 
+func (catItem *Media) name() string { return catItem.Media.Name }
```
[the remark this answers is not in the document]
vbauzysvmware: fixed

govcd/media_test.go
```diff
@@ -196,3 +239,232 @@ func (vcd *TestVCD) Test_RefreshMediaImage(check *C) {
...
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
+
+	mediaItem, err = vcd.vdc.FindMediaImage(itemName)
+	check.Assert(err, IsNil)
+	check.Assert(mediaItem, Equals, MediaItem{})
+
+}
+
+func (vcd *TestVCD) Test_QueryMediaImage(check *C) {
```
lvirbalas: After the recent commit, I believe this should be name `Test_QueryMedia`?

CHANGELOG.md
lvirbalas: They were added just recently I believe, @vbauzysvmware ?

@vbauzysvmware also after that recent commit this should state `QueryMedia` not `QueryMediaImage` anymore? Please check for more places where old "QueryMediaImage" name might have remained unchanged.

govcd/media_test.go
```diff
@@ -196,3 +239,232 @@ func (vcd *TestVCD) Test_RefreshMediaImage(check *C) {
...
+
+// Tests System function Delete by creating media and deleting it after.
+func (vcd *TestVCD) Test_DeleteMediaImage(check *C) {
+	fmt.Printf("Running: %s\n", check.TestName())
+
+	skipWhenMediaPathMissing(vcd, check)
+	itemName := `Test_DeleteMediaImage`
+
+	if vcd.config.VCD.Org == "" {
+		check.Skip("Test_GetMedia: Org name not given")
```
dataclouder: `Test_GetMedia` should be `TestDeleteMediaImage`

govcd/media_test.go
```diff
@@ -196,3 +239,232 @@ func (vcd *TestVCD) Test_RefreshMediaImage(check *C) {
...
+		return
+	}
+	if vcd.config.VCD.Catalog.Name == "" {
+		check.Skip("Test_GetMedia: Catalog name not given")
```
dataclouder: `Test_GetMedia` should be `TestDeleteMediaImage`

govcd/media_test.go
```diff
@@ -196,3 +239,232 @@ func (vcd *TestVCD) Test_RefreshMediaImage(check *C) {
...
+}
+
+func (vcd *TestVCD) Test_QueryMedia(check *C) {
+	fmt.Printf("Running: %s\n", check.TestName())
+
+	skipWhenMediaPathMissing(vcd, check)
+
+	testQueryMediaName := "testQueryMedia"
+	if vcd.config.VCD.Org == "" {
+		check.Skip("Test_GetMedia: Org name not given")
```
dataclouder: `Test_GetMedia` should be `TestQueryMedia`

govcd/media_test.go
```diff
@@ -196,3 +239,232 @@ func (vcd *TestVCD) Test_RefreshMediaImage(check *C) {
...
+		return
+	}
+	if vcd.config.VCD.Catalog.Name == "" {
+		check.Skip("Test_GetMedia: Catalog name not given")
```
dataclouder: `Test_GetMedia` should be `TestQueryMedia`

govcd/media_test.go
```diff
@@ -196,3 +239,232 @@ func (vcd *TestVCD) Test_RefreshMediaImage(check *C) {
...
+	check.Assert(media, IsNil)
+}
+
+// Tests System function Refresh
+func (vcd *TestVCD) Test_RefreshMediaRecord(check *C) {
+	fmt.Printf("Running: %s\n", check.TestName())
+
+	skipWhenMediaPathMissing(vcd, check)
+	if vcd.config.VCD.Org == "" {
+		check.Skip("Test_GetMedia: Org name not given")
```
dataclouder: `Test_GetMedia` should be `TestRefreshMediaRecord`

govcd/media_test.go
```diff
@@ -196,3 +239,232 @@ func (vcd *TestVCD) Test_RefreshMediaImage(check *C) {
...
+		return
+	}
+	if vcd.config.VCD.Catalog.Name == "" {
+		check.Skip("Test_GetMedia: Catalog name not given")
```
dataclouder: `Test_GetMedia` should be `TestRefreshMediaRecord`

govcd/media_test.go
```diff
@@ -196,3 +239,232 @@ func (vcd *TestVCD) Test_RefreshMediaImage(check *C) {
...
+	check.Assert(err, IsNil)
+	check.Assert(catalog, NotNil)
+
+	itemName := "TestGetMedia"
+	uploadTask, err := catalog.UploadMediaImage(itemName, "upload from test", vcd.config.Media.MediaPath, 1024)
+	check.Assert(err, IsNil)
+	err = uploadTask.WaitTaskCompletion()
+	check.Assert(err, IsNil)
+
+	AddToCleanupList(itemName, "mediaCatalogImage", vcd.org.Org.Name+"|"+vcd.config.VCD.Catalog.Name, "Test_GetMedia")
```
dataclouder: Why not using the existing media item from `vcd.config.Media` instead of uploading a new one?

govcd/media.go
```diff
@@ -354,3 +413,220 @@ func (mediaItem *MediaItem) Refresh() error {
...
+	_, err := media.client.ExecuteRequest(url, http.MethodGet,
+		"", "error retrieving media: %s", nil, media.Media)
+
+	return err
+}
+
+// GetMediaByHref finds a Media by HREF
+// On success, returns a pointer to the Media structure and a nil error
+// On failure, returns a nil pointer and an error
+func (cat *Catalog) GetMediaByHref(mediaItemHref string) (*Media, error) {
```
dataclouder: Should `mediaItemHref` be `mediaRecordHref`?

govcd/media.go
```diff
@@ -354,3 +413,220 @@ func (mediaItem *MediaItem) Refresh() error {
...
+	if entity == nil {
+		return nil, err
+	}
+	return entity.(*Media), err
+}
+
+// GetMediaByHref finds a Media by HREF
+// On success, returns a pointer to the Media structure and a nil error
+// On failure, returns a nil pointer and an error
+func (adminCatalog *AdminCatalog) GetMediaByHref(mediaItemHref string) (*Media, error) {
```
dataclouder: Should `mediaItemHref` be `mediaRecordHref`?

govcd/metadata.go
```diff
@@ -247,6 +251,109 @@ func (mediaItem *MediaItem) DeleteMetadata(key string) error {
 
 // DeleteMetadataAsync() function calls private function deleteMetadata() with mediaItem.client and mediaItem.MediaItem.HREF
 // which deletes metadata depending on key provided as input from media item.
+// Deprecated: Use MediaRecord.DeleteMetadataAsync
 func (mediaItem *MediaItem) DeleteMetadataAsync(key string) (Task, error) {
 	return deleteMetadata(mediaItem.vdc.client, key, mediaItem.MediaItem.HREF)
 }
+
+// GetMetadata() function calls private function getMetadata() with mediaItem.client and mediaItem.MediaItem.HREF
```
dataclouder: `mediaItem*` should be `mediaRecord`

Same for the other functions below.

govcd/media.go
```diff
@@ -19,21 +19,48 @@ import (
 	"github.com/vmware/go-vcloud-director/v2/util"
 )
 
+// Deprecated: use MediaRecord
```
dataclouder: This name has been replaced somewhere, but not consistently.
It is still used in non-deprecated functions and in comments.

Please make sure that we use the name `mediaItem` or named composed with it only in deprecated functions and comments. Otherwise, we'll have a confusing situation in our hands.

govcd/media_test.go
```diff
@@ -196,3 +239,232 @@ func (vcd *TestVCD) Test_RefreshMediaImage(check *C) {
...
+	check.Assert(err, IsNil)
+
+	uploadTask, err := catalog.UploadMediaImage(itemName, "upload from test", vcd.config.Media.MediaPath, 1024)
+	check.Assert(err, IsNil)
+	err = uploadTask.WaitTaskCompletion()
+	check.Assert(err, IsNil)
+
+	AddToCleanupList(itemName, "mediaCatalogImage", vcd.config.VCD.Catalog.Name+"|"+vcd.vdc.Vdc.Name, "Test_UploadMediaImage")
+
+	mediaItem, err := vcd.vdc.FindMediaImage(itemName)
```
dataclouder: Using a deprecated function

govcd/media_test.go
vbauzysvmware: fixed

govcd/media_test.go
vbauzysvmware: fixed

govcd/media_test.go
vbauzysvmware: fixed

govcd/media_test.go
vbauzysvmware: fixed

govcd/media_test.go
vbauzysvmware: fixed

govcd/media_test.go
vbauzysvmware: fixed

govcd/media.go
vbauzysvmware: changed to mediaHref

govcd/metadata.go
vbauzysvmware: updated

govcd/media.go
vbauzysvmware: If you find anything else please state, cause I scanned I don't see usage.

govcd/media_test.go
vbauzysvmware: These are tests for deprecated functions, no need any changes here.

govcd/media.go
dataclouder: I am not talking about the type, but the type name. In `media.go` only I can find the following functions that are using a MediaItem.
`UploadMediaImage` `executeUpload`, `RemoveMediaImageIfExists`, `RemoveMediaIfExists`

There are many more in the rest of the code

govcd/media_test.go
vbauzysvmware: udpated

govcd/media.go
vbauzysvmware: All not deprecated methods now don't have mediaItem anymore

govcd/media_test.go
dataclouder: The stipulation when we started to write the `Get*` methods was to prove the new functionality by replacing every call to the deprecated methods with the new ones.

We have done that for the retrieval methods of
* Org
* Catalog
* Catalog Item
* vApp
* VM
* External Network
* Networks
* Edge Gateway
* Vdc

I'd like the media item to follow the same path.

vbauzysvmware: Could you directly state was is missing?

dataclouder: media_test.go and vdc_test.go are still using deprecated functions.

vbauzysvmware: updated

govcd/media_test.go
```diff
@@ -112,37 +121,150 @@ func (vcd *TestVCD) Test_UploadMediaImage_error_withSameItem(check *C) {
...
+	check.Assert(media, Not(Equals), Media{})
 
-	task, err := mediaItem.Delete()
+	task, err := media.Delete()
 	check.Assert(err, IsNil)
 	err = task.WaitTaskCompletion()
 	check.Assert(err, IsNil)
 
-	mediaItem, err = vcd.vdc.FindMediaImage(itemName)
+	media, err = catalog.GetMediaByName(itemName, true)
```
dataclouder: This occurrence of `media` is never used.

govcd/metadata.go
```diff
@@ -247,6 +251,109 @@ func (mediaItem *MediaItem) DeleteMetadata(key string) error {
...
+	return mediaRecord, nil
+}
+
+// AddMetadataAsync() function calls private function addMetadata() with MediaRecord.client and MediaRecord.MediaRecord.HREF
+// which adds metadata key/value pair provided as input.
+func (mediaRecord *MediaRecord) AddMetadataAsync(key string, value string) (Task, error) {
+	return addMetadata(mediaRecord.client, key, value, mediaRecord.MediaRecord.HREF)
+}
+
+// DeleteMetadata() function calls deletes metadata depending on key provided as input from media item.
```
dataclouder:
```suggestion
// DeleteMetadata deletes metadata depending on key provided as input from media item.
```

govcd/metadata.go
```diff
@@ -247,6 +251,109 @@ func (mediaItem *MediaItem) DeleteMetadata(key string) error {
...
+	}
+	err = task.WaitTaskCompletion()
+	if err != nil {
+		return fmt.Errorf("error completing delete metadata for media item task: %s", err)
+	}
+
+	return nil
+}
+
+// DeleteMetadataAsync() function calls private function deleteMetadata() with MediaRecord.client and MediaRecord.MediaRecord.HREF
```
dataclouder:
```suggestion
// DeleteMetadataAsync calls private function deleteMetadata() with MediaRecord.client and MediaRecord.MediaRecord.HREF
```

govcd/metadata.go
```diff
@@ -247,6 +251,109 @@ func (mediaItem *MediaItem) DeleteMetadata(key string) error {
...
 	return deleteMetadata(mediaItem.vdc.client, key, mediaItem.MediaItem.HREF)
 }
+
+// GetMetadata() function calls private function getMetadata() with MediaRecord.client and MediaRecord.MediaRecord.HREF
+// which returns a *types.Metadata struct for provided media item input.
+func (mediaRecord *MediaRecord) GetMetadata() (*types.Metadata, error) {
+	return getMetadata(mediaRecord.client, mediaRecord.MediaRecord.HREF)
+}
+
+// AddMetadata() function adds metadata key/value pair provided as input.
```
dataclouder:
```suggestion
// AddMetadata adds metadata key/value pair provided as input.
```

govcd/metadata.go
```diff
@@ -247,6 +251,109 @@ func (mediaItem *MediaItem) DeleteMetadata(key string) error {
 
 // DeleteMetadataAsync() function calls private function deleteMetadata() with mediaItem.client and mediaItem.MediaItem.HREF
 // which deletes metadata depending on key provided as input from media item.
+// Deprecated: Use MediaRecord.DeleteMetadataAsync
 func (mediaItem *MediaItem) DeleteMetadataAsync(key string) (Task, error) {
 	return deleteMetadata(mediaItem.vdc.client, key, mediaItem.MediaItem.HREF)
 }
+
+// GetMetadata() function calls private function getMetadata() with MediaRecord.client and MediaRecord.MediaRecord.HREF
```
dataclouder:
```suggestion
// GetMetadata calls private function getMetadata() with MediaRecord.client and MediaRecord.MediaRecord.HREF
```

govcd/metadata.go
```diff
@@ -247,6 +251,109 @@ func (mediaItem *MediaItem) DeleteMetadata(key string) error {
...
+
+	err = mediaRecord.Refresh()
+	if err != nil {
+		return nil, fmt.Errorf("error refreshing media item: %s", err)
+	}
+
+	return mediaRecord, nil
+}
+
+// AddMetadataAsync() function calls private function addMetadata() with MediaRecord.client and MediaRecord.MediaRecord.HREF
```
dataclouder:
```suggestion
// AddMetadataAsync calls private function addMetadata() with MediaRecord.client and MediaRecord.MediaRecord.HREF
```

govcd/media_test.go
```diff
@@ -112,37 +121,150 @@ func (vcd *TestVCD) Test_UploadMediaImage_error_withSameItem(check *C) {
...
 	check.Assert(err, IsNil)
 
-	AddToCleanupList(itemName, "mediaImage", vcd.org.Org.Name+"|"+vcd.vdc.Vdc.Name, "Test_UploadMediaImage")
+	AddToCleanupList(itemName, "mediaCatalogImage", vcd.org.Org.Name+"|"+vcd.config.VCD.Catalog.Name, "Test_DeleteMediaImage")
 
-	mediaItem, err := vcd.vdc.FindMediaImage(itemName)
+	media, err := catalog.GetMediaByName(itemName, true)
 	check.Assert(err, IsNil)
-	check.Assert(mediaItem, Not(Equals), MediaItem{})
+	check.Assert(media, Not(Equals), Media{})
```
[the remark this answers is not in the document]
vbauzysvmware: added

govcd/media_test.go
```diff
@@ -112,37 +121,150 @@ func (vcd *TestVCD) Test_UploadMediaImage_error_withSameItem(check *C) {
...
+	check.Assert(err, IsNil)
+	err = uploadTask.WaitTaskCompletion()
+	check.Assert(err, IsNil)
+
+	AddToCleanupList(itemName, "mediaCatalogImage", vcd.org.Org.Name+"|"+vcd.config.VCD.Catalog.Name, "Test_RefreshMediaImage")
+
+	media, err := catalog.GetMediaByName(itemName, true)
+	check.Assert(err, IsNil)
+	check.Assert(media, NotNil)
+	check.Assert(media, Not(Equals), Media{})
```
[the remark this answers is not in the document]
vbauzysvmware: added

govcd/media_test.go
```diff
@@ -152,47 +274,66 @@ func (vcd *TestVCD) Test_FindMediaAsCatalogItem(check *C) {
...
-	AddToCleanupList(itemName, "mediaImage", vcd.org.Org.Name+"|"+vcd.vdc.Vdc.Name, "Test_UploadMediaImage")
+	AddToCleanupList(itemName, "mediaCatalogImage", vcd.org.Org.Name+"|"+vcd.config.VCD.Catalog.Name, "Test_RefreshMediaImage")
 
-	mediaItem, err := vcd.vdc.FindMediaImage(itemName)
+	mediaRecord, err := catalog.QueryMedia(itemName)
 	check.Assert(err, IsNil)
-	check.Assert(mediaItem, NotNil)
-	check.Assert(mediaItem, Not(Equals), MediaItem{})
+	check.Assert(mediaRecord, NotNil)
+	check.Assert(mediaRecord, Not(Equals), MediaRecord{})
```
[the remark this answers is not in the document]
vbauzysvmware: added

govcd/metadata_test.go
```diff
@@ -294,26 +303,26 @@ func (vcd *TestVCD) Test_DeleteMetadataOnMediaItem(check *C) {
...
 	check.Assert(err, IsNil)
 
-	mediaItem, err := vcd.vdc.FindMediaImage(itemName)
+	mediaRecord, err := catalog.QueryMedia(itemName)
 	check.Assert(err, IsNil)
-	check.Assert(mediaItem, NotNil)
-	check.Assert(mediaItem, Not(Equals), MediaItem{})
-	check.Assert(mediaItem.MediaItem.Name, Equals, itemName)
+	check.Assert(mediaRecord, NotNil)
+	check.Assert(mediaRecord, Not(Equals), MediaRecord{})
```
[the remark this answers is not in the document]
vbauzysvmware: There is check one line belowe

govcd/metadata_test.go
```diff
@@ -343,23 +354,23 @@ func (vcd *TestVCD) Test_AddMetadataOnMediaItem(check *C) {
...
```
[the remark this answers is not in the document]
vbauzysvmware: There is check one line belowe

govcd/catalog_test.go
```diff
@@ -425,20 +456,20 @@ func (vcd *TestVCD) Test_CatalogDeleteMediaImage(check *C) {
...
 	check.Assert(err, IsNil)
 
-	AddToCleanupList(itemName, "mediaImage", vcd.org.Org.Name+"|"+vcd.vdc.Vdc.Name, "Test_UploadCatalogMediaImage")
+	AddToCleanupList(itemName, "mediaCatalogImage", vcd.org.Org.Name+"|"+vcd.config.VCD.Catalog.Name, "Test_CatalogDeleteMediaImage")
 
-	mediaItem, err := vcd.vdc.FindMediaImage(itemName)
+	mediaRecord, err := catalog.QueryMedia(itemName)
 	check.Assert(err, IsNil)
-	check.Assert(mediaItem, Not(Equals), MediaItem{})
+	check.Assert(mediaRecord, Not(Equals), MediaRecord{})
```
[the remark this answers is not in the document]
vbauzysvmware: added

govcd/media_test.go
```diff
@@ -152,47 +274,66 @@ func (vcd *TestVCD) Test_FindMediaAsCatalogItem(check *C) {
...
-	catalogItem, err := FindMediaAsCatalogItem(vcd.org, vcd.config.VCD.Catalog.Name, itemName)
-	check.Assert(err, IsNil)
-	check.Assert(catalogItem, Not(Equals), CatalogItem{})
-	check.Assert(catalogItem.CatalogItem.Name, Equals, itemName)
+	check.Assert(media.MediaRecord.Name, Equals, testQueryMediaName)
+	check.Assert(media.MediaRecord.HREF, Not(Equals), "")
 
+	// find Invalid media
+	media, err = catalog.QueryMedia("INVALID")
+	check.Assert(err, Equals, ErrorEntityNotFound)
```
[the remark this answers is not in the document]
vbauzysvmware: updated in all occurrences

govcd/media_test.go
vbauzysvmware: udpated

govcd/metadata.go
vbauzysvmware: updated all occurrences

govcd/metadata.go
vbauzysvmware: updated all occurrences

govcd/metadata_test.go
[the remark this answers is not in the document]
dataclouder: `mediaRecord` is a pointer. We should not compare it to an object

govcd/metadata.go
vbauzysvmware: updated all occurrences

govcd/metadata.go
vbauzysvmware: updated all occurrences

govcd/metadata.go
vbauzysvmware: updated all occurrences

govcd/metadata_test.go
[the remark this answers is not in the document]
vbauzysvmware: removed

govcd/metadata_test.go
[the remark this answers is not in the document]
vbauzysvmware: removed

govcd/media_test.go
vbauzysvmware: updated

govcd/catalog_test.go
```diff
@@ -425,20 +456,20 @@ func (vcd *TestVCD) Test_CatalogDeleteMediaImage(check *C) {
...
+	task, err := mediaRecord.Delete()
 	check.Assert(err, IsNil)
 	err = task.WaitTaskCompletion()
 	check.Assert(err, IsNil)
 
-	mediaItem, err = vcd.vdc.FindMediaImage(itemName)
-	check.Assert(err, IsNil)
-	check.Assert(mediaItem, Equals, MediaItem{})
+	mediaRecord, err = catalog.QueryMedia(itemName)
+	check.Assert(err, Equals, ErrorEntityNotFound)
```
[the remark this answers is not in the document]
vbauzysvmware: updated

govcd/media.go
```diff
@@ -354,3 +413,220 @@ func (mediaItem *MediaItem) Refresh() error {
...
+}
+
+// GetMediaByHref finds a Media by HREF
+// On success, returns a pointer to the Media structure and a nil error
+// On failure, returns a nil pointer and an error
+func (cat *Catalog) GetMediaByHref(mediaHref string) (*Media, error) {
+
+	media := NewMedia(cat.client)
+
+	_, err := cat.client.ExecuteRequest(mediaHref, http.MethodGet,
```
dataclouder: If we get an explicit response here (instead of `_`), we can check whether the status code is 404, and in that case return `ErrEntityNotFound`. This way, `GetMediaById` doesn't need to be changed (apart from returning the correct errors on parsing failures) and the entity test can be used.

govcd/media.go
```diff
@@ -354,3 +413,220 @@ func (mediaItem *MediaItem) Refresh() error {
...
+		util.Logger.Printf("[Error] parsing bareID from mediaId %s: %s", mediaId, err)
+		return nil, ErrorEntityNotFound
+	}
+	if mediaBareId == "" {
+		util.Logger.Printf("[Error] parsing bareID from mediaId %s - empty bareID returned", mediaId)
+		return nil, ErrorEntityNotFound
+	}
+	mediaHREF.Path += fmt.Sprintf("/media/%s", mediaBareId)
+	return cat.GetMediaByHref(mediaHREF.String())
+}
```
lvirbalas: The `GetMediaById` function does not behave by the contract:
- It returns `ErrorEntityNotFound` if ID is **malformed**, but does _not_ return it if entity does not exist
- It does _not_ return `ErrorEntityNotFound` if entity does not exist, because `GetMediaByHref` does not check for that

I believe this implementation causes a rippling effect further down, like the suggestion to remove a line from the generic test, which checks for when entity doesn't exist and throws `ErrorEntityNotFound`.

So:
- Either we refactor _all_ `Get...Href` functions to start returning `ErrorEntityNotFound` correctly, which should be represented by a 404 result from API (assumption);
- Or we refactor _this_ function to return `ErrorEntityNotFound` correctly, which probably could be done one of two ways:
a. By copying approach of the `GetMediaByName` function above;
b. By calling `queryMediaWithFilter` with "catalog" and "id" filters set.

CC: @dataclouder @Didainius

govcd/media.go
```diff
@@ -354,3 +413,220 @@ func (mediaItem *MediaItem) Refresh() error {
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	mediaHREF := cat.client.VCDHREF
+
+	mediaBareId, err := getBareEntityUuid(mediaId)
+	if err != nil {
+		util.Logger.Printf("[Error] parsing bareID from mediaId %s: %s", mediaId, err)
+		return nil, ErrorEntityNotFound
```
dataclouder: simply return the error received from getBareEntityUuid

govcd/media.go
```diff
@@ -354,3 +413,220 @@ func (mediaItem *MediaItem) Refresh() error {
...
+	}
+	if mediaBareId == "" {
+		util.Logger.Printf("[Error] parsing bareID from mediaId %s - empty bareID returned", mediaId)
+		return nil, ErrorEntityNotFound
```
dataclouder: Please change to an error defining the failure, similar to the log message above

govcd/media.go
vbauzysvmware: * when Id don't exist  we get - Response status 403 Forbidden
* variant a - possible, but nonsense as this requires for 100 media make 101 request in worst case
* variant b - needs further investigation as I already stumble on that queries don't work with ID in some cases

govcd/media.go
vbauzysvmware: tx, changed

govcd/media.go
vbauzysvmware: tx, changed

govcd/media.go
vbauzysvmware: We get `explicit` Response status 403 Forbidden - which is quite common in API in such cases. If we agree - check would be simple.

dataclouder: I was expecting to get a 404.
I'd like to see an explicit test for this.

vbauzysvmware: Usually it's security feature to return 403. 404 already tells information.

govcd/media.go
lvirbalas: OK, so (a) is not viable, because:

> with catalogItem you have Name, but not real media ID - you have to call every catalogItem to find ID

Hence, we agreed to pursue extending the `GetMediaByHref` function to return `ErrorEntityNotFound` on `403`. If it works as expected, later we can refactor all `Get...ByHref` to return `ErrorEntityNotFound`.

vbauzysvmware: Change added

govcd/media.go
```diff
@@ -354,3 +414,224 @@ func (mediaItem *MediaItem) Refresh() error {
...
+			}
+		}
+	}
+	return nil, ErrorEntityNotFound
+}
+
+// GetMediaById finds a Media by ID
+// On success, returns a pointer to the Media structure and a nil error
+// On failure, returns a nil pointer and an error
+func (cat *Catalog) GetMediaById(mediaId string, refresh bool) (*Media, error) {
```
dataclouder: If the implementation doesn't require looping through the internal items, the refresh argument is not needed.

vbauzysvmware: good catch - removed

govcd/media.go
lvirbalas: It was not. Recent commit started returning `ErrorEntityNotFound` on parsing error:
https://github.com/vbauzysvmware/go-vcloud-director/blob/changes-for-media/govcd/media.go#L487

We discussed at the beginning that this is not right.

lvirbalas: It is a bad practice to assume knowledge about the calling functions from within a different function. Such assumptions lead to bad design choices and poorly factorized code. This is especially import when dealing with public functions.

`GetById` returns `cat.GetMediaByHref(mediaHREF.String())`. `GetMediaByHref` has been refactored to return `ErrorEntityNotFound`, hence returning of `ErrorEntityNotFound` on parsing error shouldn't be needed.

dataclouder: The error that we get here is "the parsing went wrong". We can't return any "not found".
GetMediaByNameOrId assumes we have implemented things according to the same principles.
If you want to go ahead with this implementation, please take that into account.

lvirbalas: After further investigation (we really need to converge)...

@dataclouder the generic `getEntityByNameOrId` is a "blind" function - it _doesn't understand_ whether it is given an ID or name and tries to call `getById` first and then `getByName`. I realize that we need this function for backwards compatibility when switching identifier from name to ID, but now it starts to bite:

```
	entity, byIdErr = getById(identifier, refresh)
	if byIdErr == nil {
		// Found by ID
		return entity, nil
	}
	if IsNotFound(byIdErr) {
		// Not found by ID, try by name
		entity, byNameErr = getByName(identifier, false)
		return entity, byNameErr
	} else {
		// On any other error, we return it
		return nil, byIdErr
	}
```

If `GetMediaById` is provided with a wrong ID (i.e. name) we could make it:
a. Return a generic parsing error. This is correct behavior from `GetMediaById` perspective, but then we need to fix the generic `getEntityByNameOrId`, because now `getEntityByNameOrId` would **not** try to search by `getByName` as it is only doing that if `if IsNotFound(byIdErr)` (!) I personally don't see how to make `getEntityByNameOrId` smarter at this point, because even the user could use name in such format like our IDs.
b. Return `ErrorEntityNotFound`. This is not ideal, but after more thought an incorrectly provided ID (like a name instead of ID) does imply it won't be found, so we're only one logic step ahead. Then this complies with what the generic `getEntityByNameOrId` expects.
c. Introduce and return a specific `ErrorParsing` and then check for that in the generic `getEntityByNameOrId` function.

lvirbalas: We agreed that we'll try one more variant before going the (b) or (c) ways:

d. Try to use `queryMediaItemsWithFilter` with catalog filter and loop/match with the help of `equalIds(...)`.

Risk: query sometimes doesn't return results without a clear reason, we have seen that in some cases with filtering PVDCs.

vbauzysvmware:
Code updated - using query method

vbauzysvmware:
Thank you both for your help.

PR #247, opened 2019-09-30
Add support for firewall rule handling using NSX-V edge gateway API

Didainius:
This PR adds:
* Support for firewall handling using NSX-V API: `CreateNsxvFirewall()`, `UpdateNsxvFirewall()`, `GetNsxvFirewallById()`, `DeleteNsxvFirewallById()` and `GetAllNsxvFirewallRules()`
* Support for enabling/disabling and altering default firewall rule as edge gateway methods: `GetFwGeneralParams()`, `UpdateFwGeneralParams()`
* Additionally `GetAnyVnicIndexByNetworkName()` to for easier interface (vNic) lookup in edge gateway
* New variation of execution function `ExecuteParamRequestWithCustomError()` which adds query parameter support on top of `ExecuteRequestWithCustomError()`
* New types `EdgeFirewallRule` and `FwGeneralParamsWithXml` for firewall manipulation

**Note** Full test suite passed on vCD 9.5 and 10


govcd/edgegateway.go
```diff
@@ -1144,20 +1144,99 @@ func (egw *EdgeGateway) UpdateLBGeneralParams(enabled, accelerationEnabled, logg
...
 	}
 
 	return updatedLb, nil
 }
 
+// GetFwGeneralParams retrieves firewall configuration and can be used
+// to alter master configuration options. These are 3 fields only:
+// FwGeneralParamsWithXml.Enabled, FwGeneralParamsWithXml.DefaultPolicy.LoggingEnabled and
+// FwGeneralParamsWithXml.DefaultPolicy.Action
+func (egw *EdgeGateway) GetFwGeneralParams() (*types.FwGeneralParamsWithXml, error) {
```
vbauzysvmware: GetFwGeneralParams -> GetFwGeneralConfig - maybe better?

govcd/edgegateway.go
```diff
@@ -1144,20 +1144,99 @@ func (egw *EdgeGateway) UpdateLBGeneralParams(enabled, accelerationEnabled, logg
...
+
+	return firewallConfig, nil
+}
+
+// UpdateFwGeneralParams allows to update firewall configuration.
+// It accepts three fields (Enabled, DefaultLoggingEnabled, DefaultAction) and uses
+// them to construct types.FwGeneralParamsWithXml without altering other options to prevent config
+// corruption.
+// They are represented in firewall configuration page in the UI.
+func (egw *EdgeGateway) UpdateFwGeneralParams(enabled, defaultLoggingEnabled bool, defaultAction string) (*types.FwGeneralParamsWithXml, error) {
```
vbauzysvmware: Change func name to config from params if agree on that.

govcd/edgegateway.go
```diff
@@ -1144,20 +1144,99 @@ func (egw *EdgeGateway) UpdateLBGeneralParams(enabled, accelerationEnabled, logg
...
+	if !egw.HasAdvancedNetworking() {
+		return nil, fmt.Errorf("only advanced edge gateway supports load balancing")
+	}
+
+	if defaultAction != "accept" && defaultAction != "deny" {
```
vbauzysvmware: Maybe is more convenient to ask user of function to provide bool for parameters `defaultAction` cause it can be only two options. Challenge here is to provide meaningful name - maybe `defaultActionAccept` then true or false.

govcd/edgegateway.go
```diff
@@ -1144,20 +1144,99 @@ func (egw *EdgeGateway) UpdateLBGeneralParams(enabled, accelerationEnabled, logg
...
+		return nil, fmt.Errorf("default action must be either 'accept' or 'deny'")
+	}
+
+	// Retrieve firewall latest configuration
+	currentFw, err := egw.GetFwGeneralParams()
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve firewall config before update: %s", err)
```
vbauzysvmware: Again func name is `params` but in error is config :-D

govcd/edgegateway.go
```diff
@@ -1199,7 +1278,40 @@ func (egw *EdgeGateway) GetVnicIndexByNetworkNameAndType(networkName, networkTyp
 	if err != nil {
 		return nil, fmt.Errorf("cannot retrieve vNic configuration: %s", err)
 	}
-	return GetVnicIndexByNetworkNameAndType(networkName, networkType, vnics)
+	return getVnicIndexByNetworkNameAndType(networkName, networkType, vnics)
+}
+
+// GetAnyVnicIndexByNetworkName returns *int of vNic index and network type by network name
```
vbauzysvmware: I would like to see explanation of vNic is and how it is mapping to edge gtw context.

govcd/edgegateway.go
```diff
@@ -1199,7 +1278,40 @@ func (egw *EdgeGateway) GetVnicIndexByNetworkNameAndType(networkName, networkTyp
...
+// networkName cannot be empty
+// networkType will be one of: 'internal', 'uplink', 'trunk', 'subinterface'
+//
+// Warning: this function assumes that there are no duplicate network names attached. If it is so
```
vbauzysvmware: What we can do about it?

govcd/nsxv_firewall.go
```diff
@@ -0,0 +1,235 @@
...
+	var resp *http.Response
+	if aboveRuleId == "" {
+		resp, err = egw.client.ExecuteRequestWithCustomError(httpPath, http.MethodPost, types.AnyXMLMime,
+			"error creating firewall rule: %s", firewallRuleRequest, &types.NSXError{})
+	} else {
+		errString := fmt.Sprintf("error creating firewall rule (aboveRuleId: %s): %%s", aboveRuleId)
+		resp, err = egw.client.ExecuteParamRequestWithCustomError(httpPath, params, http.MethodPost, types.AnyXMLMime,
+			errString, firewallRuleConfig, &types.NSXError{})
+	}
+
```
vbauzysvmware: not needed empty line

govcd/nsxv_firewall.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	httpPath, err := egw.buildProxiedEdgeEndpointURL(types.EdgeCreateFirewallPath + "/" + firewallRuleConfig.ID)
+	if err != nil {
+		return nil, fmt.Errorf("could not get Edge Gateway API endpoint: %s", err)
+	}
+
+	// Result should be 204, if not we expect an error of type types.NSXError
```
vbauzysvmware: Comment it's ok, but in following code we don't check for status?

govcd/nsxv_firewall.go
```diff
@@ -0,0 +1,235 @@
...
+	httpPath, err := egw.buildProxiedEdgeEndpointURL(types.EdgeFirewallPath)
+	if err != nil {
+		return nil, fmt.Errorf("could not get Edge Gateway API endpoint: %s", err)
+	}
+
+	firewallRuleResponse := &responseEdgeFirewallRules{}
+
+	// This query returns all application rules as the API does not have filtering options
+	_, err = egw.client.ExecuteRequest(httpPath, http.MethodGet, types.AnyXMLMime,
+		"unable to read firewall rule: %s", nil, firewallRuleResponse)
```
vbauzysvmware:
```suggestion
		"unable to read firewall rules: %s", nil, firewallRuleResponse)
```

govcd/nsxv_firewall.go
```diff
@@ -0,0 +1,235 @@
...
+
+// GetNsxvFirewallById retrieves types.EdgeFirewallRule by firewall rule ID as shown in the UI using proxied
+// NSX-V API.
+// It returns and error `ErrorEntityNotFound` if the firewall rule is not found
+func (egw *EdgeGateway) GetNsxvFirewallById(id string) (*types.EdgeFirewallRule, error) {
+	if err := validateGetNsxvFirewall(id, egw); err != nil {
+		return nil, err
+	}
+
+	httpPath, err := egw.buildProxiedEdgeEndpointURL(types.EdgeFirewallPath)
```
vbauzysvmware: why `GetAllNsxvFirewallRules` wasn't used ?

govcd/nsxv_firewall.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not get Edge Gateway API endpoint: %s", err)
+	}
+
+	firewallRuleResponse := &responseEdgeFirewallRules{}
+
+	// This query returns all application rules as the API does not have filtering options
+	_, err = egw.client.ExecuteRequest(httpPath, http.MethodGet, types.AnyXMLMime,
+		"unable to read firewall rule: %s", nil, firewallRuleResponse)
```
vbauzysvmware:
```suggestion
		"unable to read firewall rules: %s", nil, firewallRuleResponse)
```

govcd/nsxv_firewall_test.go
```diff
@@ -0,0 +1,136 @@
...
+ */
+
+package govcd
+
+import (
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+	. "gopkg.in/check.v1"
+)
+
+func (vcd *TestVCD) Test_NsxvFirewallRule(check *C) {
```
vbauzysvmware: I am not fond of tests which tests few methods. Usually test scope is one main function.

govcd/api_vcd_test.go
```diff
@@ -997,6 +997,31 @@ func (vcd *TestVCD) removeLeftoverEntities(entity CleanupEntity) {
...
+		}
+
+		orgName, vdcName, edgeName := splitParent(entity.Parent, "|")
+
+		_, _, edge, err := getOrgVdcEdgeByNames(vcd, orgName, vdcName, edgeName)
+		if err != nil {
+			vcd.infoCleanup("removeLeftoverEntries: [ERROR] %s \n", err)
+		}
+
+		err = edge.DeleteNsxvFirewallById(entity.Name)
```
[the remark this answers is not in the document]
lvirbalas: I agree, as "delete firewall" sounds to be something bigger.

govcd/nsxv_firewall.go
```diff
@@ -0,0 +1,235 @@
...
+	XMLName           xml.Name                 `xml:"firewall"`
+	Version           string                   `xml:"version"`
+	EdgeFirewallRules requestEdgeFirewallRules `xml:"firewallRules"`
+}
+
+// CreateNsxvFirewall creates firewall rule using proxied NSX-V API. It is a synchronuous operation.
+// It returns an object with all fields populated (including ID)
+// If aboveRuleId is not empty, it will send a query parameter aboveRuleId= which instructs NSX to
+// place this rule above the specified rule ID
+func (egw *EdgeGateway) CreateNsxvFirewall(firewallRuleConfig *types.EdgeFirewallRule, aboveRuleId string) (*types.EdgeFirewallRule, error) {
```
lvirbalas: Similarly to the Terraform PR, I believe all the related functions in govcd should contain `...FirewallRule` at the end. We are dealing with firewall rules here, as per comment above, hence we should name the function as such.

Note: `GetFwGeneralParams()`, `UpdateFwGeneralParams()` are probably exceptions, because they do deal with the whole FW config at once.

CC: @vbauzysvmware @dataclouder

vbauzysvmware: I made same note in my review, also recommend to change from `Params` to `Config`

govcd/edgegateway.go
dataclouder: I prefer `Upfdate...` as it makes clear that it is modifying the entity.

govcd/api_vcd_test.go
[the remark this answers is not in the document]
Didainius: I had initially created it so, then shortened. Switched back to have `Rule` everywhere.

govcd/edgegateway.go
Didainius: Done.

govcd/edgegateway.go
Didainius: `UpdateFwGeneralConfig` it is

govcd/edgegateway.go
Didainius: This function has become "config".

govcd/edgegateway.go
Didainius: This could be a bit lengthy instruction. In fact the unit test has a sample XML config how it looks like. https://github.com/vmware/go-vcloud-director/pull/247/files#diff-5ae225082b593473c248452f5c3b8b2cR297

I have adjusted the comment to just touch the point that XML contains a mapping of vNic (network interfaces on edge gateway) and network names and their types.

govcd/edgegateway.go
Didainius: In fact I could not make this happen during my tests, but still left this note as this is how the search algorithm behaves in the case it had it. The unit test `Test_getVnicIndexFromNetworkNameType` also has a fake XML to test it does so.

govcd/nsxv_firewall.go
Didainius: Removed

govcd/nsxv_firewall.go
Didainius: Done.

govcd/nsxv_firewall_test.go
Didainius: Could be. I can try to split them into multiple. The only reasons for this is to save on test suite time. (For example to do a TestDelete it would still need to be created)
@dataclouder, @lvirbalas your thoughts?

govcd/nsxv_firewall.go
Didainius: This is changed. `GetFwGeneralParams()` and `UpdateFwGeneralParams()` have become `GetFwGeneralConfig()` and `UpdateFwGeneralConfig()` respectively.

govcd/api_vcd_test.go
```diff
@@ -1171,11 +1196,11 @@ func Test_splitParent(t *testing.T) {
 	}
 }
 
-// testGetLBGeneralParamsXML is used for additional validation that modifying load balancer
-// does not change any single field. It returns a string of whole load balancer configuration
-func testGetLBGeneralParamsXML(edge EdgeGateway, check *C) string {
+// testGetEdgeEndpointXML is used for additional validation that modifying edge gateway endpoint
```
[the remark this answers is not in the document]
Didainius: Indeed. This was some misplacement. Moved.

govcd/nsxv_firewall_test.go
lvirbalas: Both approaches are valid and have their own place. However, a composite test is capable of finding deeper and unexpected bugs than a simple test. This test is creating multiple rules and testing the order of them, hence I'm in favor of leaving it as is.

govcd/nsxv_firewall.go
Didainius: Good point. `GetAllNsxvFirewallRules` was done a lot later in the coding cycle. Changed.

Didainius:
Acceptance tests passed on 9.5 and 10

govcd/edgegateway.go
Didainius: Open to discussion - surely I can do that but I am not fully sure we are winning a lot here.

The var `defaultActionDeny = true` or `defaultActionDeny = false` is not that much more readable than `defaultAction = accept` or `defaultAction = deny` and I would still have to convert it behind the scenes

cc @dataclouder @lvirbalas

lvirbalas: For me it's more clear with "accept" and "deny", as they are standard in FW world.

CHANGELOG.md
```diff
@@ -35,6 +35,11 @@ which use the proxied NSX-V API of advanced edge gateway for handling NAT rules
 * Added methods `Client.GetVMByHref` `Vapp.GetVAMByName` and related `GetVMById`, `GetVAMByNameOrId`
 * Deprecated methods `Client.FindVMByHREF`, `Vdc.FindVMByName`, `Vdc.FindVAppByID`, and `Vdc.FindVAppByName`
 * Added methods `Vm.GetGuestCustomizationSection` and `Vm.SetGuestCustomizationSection`  
+* Added methods `CreateNsxvFirewall()`, `UpdateNsxvFirewall()`, `GetNsxvFirewallById()`, `DeleteNsxvFirewallById()`
```
lvirbalas: Not renamed after the `Rule` change?

CHANGELOG.md
```diff
@@ -35,6 +35,11 @@ which use the proxied NSX-V API of advanced edge gateway for handling NAT rules
...
+which use the proxied NSX-V API of advanced edge gateway for handling firewall rules [#247](https://github.com/vmware/go-vcloud-director/pull/247)
+* Added methods `GetFwGeneralParams()`, `UpdateFwGeneralParams()` for changing global firewall settings [#247](https://github.com/vmware/go-vcloud-director/pull/247)
```
lvirbalas: Method names need updating after a recent commit.

BTW, is "Fw" abbreviation ours or we are mimicking vCD API? I like it more with "Firewall" as in the rules' functions, as then you can search the whole code base for "Firewall" and you'll get all occurrences (and life showed that we do that with regexps when evaluating the state of code base). To make the function names shorter we could drop the word "General" and just use, say, `UpdateFirewallConfig`. Then all firewall functions would be named consistently. Your thoughts?

CC @vbauzysvmware @dataclouder

Didainius: It is me. I was relating it to the similar functions `UpdateLb...` Can drop it easily.

CHANGELOG.md
Didainius: Obviously. Fixed...

CHANGELOG.md
lvirbalas: If you agree, let's use the full word "Firewall" everywhere. If not, let's use the "Fw" everywhere, but, most importantly, let's be consistent :)

vbauzysvmware: I can live with Fw as with Lb -in time you get use to it. Personally prefer longer names. `General` comes from UI and API - so woudn't so keen to Drop.

govcd/edgegateway.go
vbauzysvmware: So Dainius you have option to choose :-D

CHANGELOG.md
Didainius: Done.

govcd/edgegateway.go
Didainius: 3 vs 1. Sorry :) leaving it as is

CHANGELOG.md
lvirbalas: Thanks!

Would you also share a link to vCD API docs about the `FwGeneralParamsWithXml` type. Just want to understand where the naming comes from.

Didainius: This comes more from my thinking. I needed to separate the firewall rule configuration from the regular firewall rule. Here was the word `General` coming from. Again it all started with load balancers. The `WithXml` suffix is to flag  that it carries the raw XML part of all rules inside.

`FirewallConfigWithXml` then?

lvirbalas: IMHO looks cleaner and consistent, but granted you have reasons for "GeneralParams".

Didainius: `FirewallConfigWithXml` it is.

govcd/nsxv_firewall.go
```diff
@@ -0,0 +1,225 @@
...
+		return nil, fmt.Errorf("unable to retrieve firewall rule with ID (%s) after update: %s",
+			readFirewallRule.ID, err)
+	}
+	return readFirewallRule, nil
+}
+
+// GetNsxvFirewallRuleById retrieves types.EdgeFirewallRule by firewall rule ID as shown in the UI using proxied
+// NSX-V API.
+// It returns and error `ErrorEntityNotFound` if the firewall rule is not found
+func (egw *EdgeGateway) GetNsxvFirewallRuleById(id string) (*types.EdgeFirewallRule, error) {
```
vbauzysvmware: Get and delete use internal IDs or UI IDs. From comment I am not sure

Didainius: Thanks. Improved comments.

PR #248, opened 2019-10-01
Debugging client timeout problems, not intended to merge

Didainius:


lvirbalas:
We agreed that this may be a good amendment to handling HTTP connectivity issues, as the scope of effect is only on connection layer. This addition works nicely below the mutex mechanism in Terraform, which works at higher layer.

Action items:
- Reduce the 10 minute timeout to 90 seconds and export it as parameter to user (just in case)
- Add the retry (as in this PoC) on connection failure, but refactor to have it in a single function
- Do **not** touch the file upload functionality, as retrying chunk uploads may lead to unwanted effects

Didainius:
I'm closing this pull request as this solution actually needs a better look and potentially use an already existing library (like `go-retryablehttp`) which handles retries and body closures properly instead of trying to reinvent the bike.

Issue #250, opened 2019-10-03
Need the ability to set root disk size on VM creation


cheald:
We're deploying a tiny .ova which we need to then increase the root disk size for on initial VM deployment. Right now, govcd can attach independent disks, but doesn't seem to have a way to manage the size of existing internally-attached disks.

I've got a PR for this one incoming shortly.

vbauzysvmware:
PR with solution is merged.

PR #251, opened 2019-10-03
Add change disk size

cheald:
## Description

Related issue: #250

govcd doesn't currently provide a mechanism for managing the RasdItemsList attached to a VM. While any number of items could be managed, the one I'm immediately interested in is the ability to change disk sizes. This PR adds `vm.ChangeDiskSize(index, size)`, which permits this action.

I plan to submit a sibling PR to the terraform provider allowing users to specify the root disk size.

I've written a test, but I don't have infrastructure set up to execute it. However, I've tested this against my 9.1 vCD environment and it works well.

### DIFF
--- a/govcd/vm.go
+++ b/govcd/vm.go
@@ -394,6 +394,70 @@ func (vm *VM) ChangeMemorySize(size int) (Task, error) {
 		types.MimeRasdItem, "error changing memory size: %s", newMem)
 }
 
+func (vm *VM) ChangeDiskSize(index int, size int) (Task, error) {
+	err := vm.Refresh()
+	if err != nil {
+		return Task{}, fmt.Errorf("error refreshing VM before running customization: %v", err)
+	}
+
+	items := &types.RasdItemsList{
+		Type:        types.MimeRasdItemList,
+		HREF:        vm.VM.HREF + "/virtualHardwareSection/disks",
+		Xmlns:       types.XMLNamespaceVCloud,
+		XmlnsRasd:   types.XMLNamespaceRASD,
+		XmlnsVCloud: types.XMLNamespaceVCloud,
+		XmlnsXsi:    types.XMLNamespaceXSI,
+		Items:       make([]*types.OVFItem, 0),
+		Link: &types.Link{
+			HREF: vm.VM.HREF + "/virtualHardwareSection/disks",
+			Rel:  "edit",
+			Type: types.MimeRasdItemList,
+		},
+	}
+
+	for _, item := range vm.VM.VirtualHardwareSection.Item {
+		newItem := &types.OVFItem{
+			XmlnsRasd:       types.XMLNamespaceRASD,
+			XmlnsVCloud:     types.XMLNamespaceVCloud,
+			XmlnsXsi:        types.XMLNamespaceXSI,
+			XmlnsVmw:        types.XMLNamespaceVMW,
+			Address:         item.Address,
+			AddressOnParent: &item.AddressOnParent,
+			Description:     item.Description,
+			ElementName:     item.ElementName,
+			InstanceID:      item.InstanceID,
+			Parent:          &item.Parent,
+			ResourceSubType: item.ResourceSubType,
+			ResourceType:    item.ResourceType,
+		}
+
+		if item.HostResource != nil && len(item.HostResource) > 0 {
+			resource := item.HostResource[0]
+			newItem.HostResource = &types.VirtualHardwareHostResource{
+				XmlnsVCloud:       types.XMLNamespaceVCloud,
+				BusSubType:        resource.BusSubType,
+				BusType:           resource.BusType,
+				Capacity:          resource.Capacity,
+				Disk:              resource.Disk,
+				OverrideVmDefault: resource.OverrideVmDefault,
+				StorageProfile:    resource.StorageProfile,
+			}
+
+			if *newItem.AddressOnParent == index && newItem.ResourceType == types.ResourceTypeDisk {
+				newItem.HostResource.Capacity = size
+			}
+		}
+
+		items.Items = append(items.Items, newItem)
+	}
+	apiEndpoint, _ := url.ParseRequestURI(vm.VM.HREF)
+	apiEndpoint.Path += "/virtualHardwareSection/disks"
+
+	// Return the task
+	return vm.client.ExecuteTaskRequest(apiEndpoint.String(), http.MethodPut,
+		types.MimeRasdItemList, "error changing disk size: %s", items)
+}
+
 func (vm *VM) RunCustomizationScript(computername, script string) (Task, error) {
 	return vm.Customize(computername, script, false)
 }
--- a/govcd/vm.go
+++ b/govcd/vm.go
@@ -394,10 +394,11 @@ func (vm *VM) ChangeMemorySize(size int) (Task, error) {
 		types.MimeRasdItem, "error changing memory size: %s", newMem)
 }
 
+// ChangeDiskSize alters the Capacity (in megabytes) of a non-independent disk attached to the VM. Disk sizes may only be increased.
 func (vm *VM) ChangeDiskSize(index int, size int) (Task, error) {
 	err := vm.Refresh()
 	if err != nil {
-		return Task{}, fmt.Errorf("error refreshing VM before running customization: %v", err)
+		return Task{}, fmt.Errorf("error refreshing VM before running customization: %s", err)
 	}
 
 	items := &types.RasdItemsList{
--- a/govcd/vm.go
+++ b/govcd/vm.go
@@ -394,8 +394,22 @@ func (vm *VM) ChangeMemorySize(size int) (Task, error) {
 		types.MimeRasdItem, "error changing memory size: %s", newMem)
 }
 
+// GetDiskHardwareItem finds the VirtualHardwareItem for a disk by bus (Address belonging to the Parent item) and unit (AddressOnParent)
+func (vm *VM) GetDiskHardwareItem(busID, unitID int) *types.VirtualHardwareItem {
+	for _, item := range vm.VM.VirtualHardwareSection.Item {
+		if item.ResourceType == types.ResourceTypeDisk && item.AddressOnParent == unitID {
+			for _, possibleParent := range vm.VM.VirtualHardwareSection.Item {
+				if possibleParent.InstanceID == item.Parent {
+					return item
+				}
+			}
+		}
+	}
+	return nil
+}
+
 // ChangeDiskSize alters the Capacity (in megabytes) of a non-independent disk attached to the VM. Disk sizes may only be increased.
-func (vm *VM) ChangeDiskSize(index int, sizeInMegabytes int) (Task, error) {
+func (vm *VM) ChangeDiskSize(bus, unit int, sizeInMegabytes int) (Task, error) {
 	err := vm.Refresh()
 	if err != nil {
 		return Task{}, fmt.Errorf("error refreshing VM before running customization: %s", err)
@@ -416,8 +430,10 @@ func (vm *VM) ChangeDiskSize(index int, sizeInMegabytes int) (Task, error) {
 		},
 	}
 
-	// Counter for the number of attached disks we've seen
-	diskCount := -1
+	disk := vm.GetDiskHardwareItem(bus, unit)
+	if disk == nil {
+		return Task{}, fmt.Errorf("error refreshing VM before running customization: couldn't find disk at bus %d unit %d", bus, unit)
+	}
 
 	for _, item := range vm.VM.VirtualHardwareSection.Item {
 		newItem := &types.OVFItem{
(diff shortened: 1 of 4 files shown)


govcd/vm_test.go
```diff
@@ -652,6 +652,48 @@ func (vcd *TestVCD) Test_AnswerVmQuestion(check *C) {
...
 
+func (vcd *TestVCD) Test_VMChangeRootDiskSize(check *C) {
+	if vcd.skipVappTests {
+		check.Skip("Skipping test because vapp was not successfully created at setup")
+	}
+	vapp := vcd.findFirstVapp()
+	existingVm, vmName := vcd.findFirstVm(vapp)
+	if vmName == "" {
+		check.Skip("skipping test because no VM is found")
+	}
```
dataclouder: There should be a check about the successful retrieval of the vApp before searching for the VM

govcd/vm_test.go
```diff
@@ -652,6 +652,48 @@ func (vcd *TestVCD) Test_AnswerVmQuestion(check *C) {
...
+
+	currentDisk := 0
+	if nil != vcd.vapp.VApp.Children.VM[0] && nil != vcd.vapp.VApp.Children.VM[0].VirtualHardwareSection && nil != vcd.vapp.VApp.Children.VM[0].VirtualHardwareSection.Item {
```
dataclouder: Please use `variableName` compared to `value` and not the other way around.
If you run [staticcheck](https://staticcheck.io), you may get more similar advice

govcd/vm_test.go
```diff
@@ -652,6 +652,48 @@ func (vcd *TestVCD) Test_AnswerVmQuestion(check *C) {
...
+		for _, item := range vcd.vapp.VApp.Children.VM[0].VirtualHardwareSection.Item {
+			if item.ResourceType == types.ResourceTypeDisk && item.AddressOnParent == 0 {
+				currentDisk = item.HostResource[0].Capacity
+				break
+			}
+		}
+	}
+	check.Assert(0, Not(Equals), currentDisk)
```
dataclouder: Same as above. First the variable, and then the value

govcd/vm_test.go
```diff
@@ -652,6 +652,48 @@ func (vcd *TestVCD) Test_AnswerVmQuestion(check *C) {
...
+	if vcd.skipVappTests {
+		check.Skip("Skipping test because vapp was not successfully created at setup")
+	}
+	vapp := vcd.findFirstVapp()
+	existingVm, vmName := vcd.findFirstVm(vapp)
+	if vmName == "" {
+		check.Skip("skipping test because no VM is found")
+	}
+
+	currentDisk := 0
```
dataclouder: This variable seems to refer to capacity, rather than a disk. Please use a name that suggests its role.

types/v56/types.go
```diff
@@ -1439,16 +1440,37 @@ type OVFItem struct {
...
+	ResourceSubType string                       `xml:"rasd:ResourceSubType,omitempty"`
+	ResourceType    int                          `xml:"rasd:ResourceType"`
+
+	VirtualQuantity int   `xml:"rasd:VirtualQuantity"`
+	Weight          int   `xml:"rasd:Weight"`
+	CoresPerSocket  *int  `xml:"vmw:CoresPerSocket,omitempty"`
+	Link            *Link `xml:"vcloud:Link"`
+}
+
+// Used to write a list of RASD items, generally for modifying disks attached to a VM
```
dataclouder: Please add the URL of the API docs page from where this structure was taken, and a note about what RASD stands for.

govcd/vm_test.go
```diff
@@ -652,6 +652,48 @@ func (vcd *TestVCD) Test_AnswerVmQuestion(check *C) {
...
+	if nil != vcd.vapp.VApp.Children.VM[0] && nil != vcd.vapp.VApp.Children.VM[0].VirtualHardwareSection && nil != vcd.vapp.VApp.Children.VM[0].VirtualHardwareSection.Item {
+		for _, item := range vcd.vapp.VApp.Children.VM[0].VirtualHardwareSection.Item {
```
dataclouder: I'd like to see some comments explaining what are the expectations of this test, or what prerequisites we should meet to make the test pass.
When I run it, I get a failure when it expects to see `currentDisk` containing a non-zero value.

govcd/vm_test.go
```diff
@@ -652,6 +652,48 @@ func (vcd *TestVCD) Test_AnswerVmQuestion(check *C) {
...
+	check.Assert(err, IsNil)
+
+	task, err := vm.ChangeDiskSize(0, currentDisk+1)
+	check.Assert(err, IsNil)
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
+	check.Assert(task.Task.Status, Equals, "success")
+
+	foundItem := false
+	if nil != vm.VM.VirtualHardwareSection.Item {
```
dataclouder: Also here, a few comments explaining what we are doing and what we expect to find would help evaluate this test better.

cheald:
I'll fix these up. As I noted, I don't have the infrastructure set up to run the test suite, so I'm not sure how to best develop it. I wrote the test as a hopeful do-90%-of-the-work and hopefully you all can help me get it over the line. I'll clean up the style issues, I may need an assist with the actual test functionality.

I cribbed the form of the test from `Test_VMChangeCPUCountWithCore`.

govcd/vm.go
```diff
@@ -394,6 +394,70 @@ func (vm *VM) ChangeMemorySize(size int) (Task, error) {
 		types.MimeRasdItem, "error changing memory size: %s", newMem)
 }
 
+func (vm *VM) ChangeDiskSize(index int, size int) (Task, error) {
+	err := vm.Refresh()
+	if err != nil {
+		return Task{}, fmt.Errorf("error refreshing VM before running customization: %v", err)
```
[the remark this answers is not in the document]
cheald: I can switch this up, but %v is consistent with the other invocations in the file (in fact, that block was copied from `ChangeMemorySize`)

govcd/vm_test.go
cheald: Will add. A similar check should be added to `Test_VMChangeCPUCountWithCore`, where this was taken from.

govcd/vm_test.go
cheald: This is again a consequence of copying the style from `Test_VMChangeCPUCountWithCore`. Would you prefer the staticcheck style, or consistency with the other tests?

govcd/vm_test.go
cheald: This may be a consequence of the OVA used to construct the VM. I'm manually testing this with an Ubuntu OVA with a 10GB root disk, which I can then modify to be larger. If you could provide more information about the error and any disks attached to the test OVA, that would be helpful.

govcd/vm.go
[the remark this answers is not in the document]
dataclouder: Please change it. I am slowly refactoring old code where this style was used.

govcd/vm_test.go
dataclouder: Just a few hours ago I updated go-vcloud-director with the removal of all these so called "Yoda comparisons". You may want to pull and merge from master again.

cheald: Oh, excellent. I'll rebase my PR in addition to the style changes.

govcd/vm_test.go
dataclouder: What we use is a standard Photon OVA (http://dl.bintray.com/vmware/photon/3.0/GA/ova/:photon-hw11-3.0-26156e2.ova). 

```
START: vm_test.go:655: TestVCD.Test_VMChangeRootDiskSize
vm_test.go:674:
    check.Assert(0, Not(Equals), currentDisk)
... obtained int = 0
... expected int = 0

FAIL: vm_test.go:655: TestVCD.Test_VMChangeRootDiskSize
```

cheald: Thanks. I'll get a VM spun up with that to test against.

dataclouder: You can look at `createTestVapp` in api_vcd_test.go to see how the initial testing vApp is created for our suite.

cheald:
I've fixed the test and hopefully addressed the style issues. Additionally, I rebased onto latest master. Let me know if there's further changes needed.

cheald:
In looking at the corresponding tests for the TF provider, it looks like `AddressOnParent == 0` isn't the correct check for the root disk. Instead, the *n*th attached disk found may be more appropriate. I've also noticed that InstanceID is always 2000 for that disk, but I don't know if that's spec or not. Do you have any insight?

govcd/vm.go
```diff
@@ -394,6 +394,77 @@ func (vm *VM) ChangeMemorySize(size int) (Task, error) {
...
 }
 
+// ChangeDiskSize alters the Capacity (in megabytes) of a non-independent disk attached to the VM. Disk sizes may only be increased.
+func (vm *VM) ChangeDiskSize(index int, sizeInMegabytes int) (Task, error) {
+	err := vm.Refresh()
+	if err != nil {
+		return Task{}, fmt.Errorf("error refreshing VM before running customization: %s", err)
+	}
+
+	items := &types.RasdItemsList{
```
vbauzysvmware:
```suggestion
	rasdItems := &types.RasdItemsList{
```

govcd/vm.go
```diff
@@ -394,6 +394,77 @@ func (vm *VM) ChangeMemorySize(size int) (Task, error) {
...
+			Rel:  "edit",
+			Type: types.MimeRasdItemList,
+		},
+	}
+
+	// Counter for the number of attached disks we've seen
+	diskCount := -1
+
+	for _, item := range vm.VM.VirtualHardwareSection.Item {
+		newItem := &types.OVFItem{
```
vbauzysvmware:
```suggestion
		newOvfItem := &types.OVFItem{
```

govcd/vm.go
```diff
@@ -394,6 +394,77 @@ func (vm *VM) ChangeMemorySize(size int) (Task, error) {
...
+			HREF: vm.VM.HREF + "/virtualHardwareSection/disks",
+			Rel:  "edit",
+			Type: types.MimeRasdItemList,
+		},
+	}
+
+	// Counter for the number of attached disks we've seen
+	diskCount := -1
+
+	for _, item := range vm.VM.VirtualHardwareSection.Item {
```
vbauzysvmware:
```suggestion
	for _, virtualHardwareItem := range vm.VM.VirtualHardwareSection.Item {
```

cheald:
I'll make those changes, but I'll have a few more - I think there's some issue with the way HostResources are currently unmarshaled, which is causing some problems when I try to use this from the TF plugin with an additional disk attached. I'll resolve those soon.

vbauzysvmware:
@cheald Be careful with XML - vCD API requires exact right order of elements it expects.

cheald:
Aye, I ran into that earlier.

I think I've got the issue sorted now. The problem is:

1. HostResource belongs to the rasd namespace, but its attributes belong to the vcloud namespace
2. Specifying `vcloud` for the attributes' namespace causes parsing to fail, because vCD passes them as the aliased namespace `ns10` (at least in my 9.1 install).
3. ...but we need them to render in the vcloud namespace when writing a request.

My solution was to add `VirtualHardwareHostResourceForWrite`, which is identical to `VirtualHardwareHostResource` except for the XML tagging. If there's a better way to do it, I'd like to, but I'm at the limits of my ability with Go's XML generation capabilities.

I've got this running locally against the TF plugin with unit tests, so I'm optimistic it'll work. I'm going to PR an updated set of commits for the TF plugin, with a gomod replacement for this branch, which should lay out tests for the whole kit and caboodle.

vbauzysvmware:
Hi, thank you for changes.
Golang doesn't work with different namespaces. So you choose one and another one just hardwire:
```

**Comment by dataclouder** [comment]
I gave the PR another go, and I've got a failure.
```
go test -tags 'functional' -timeout 180m -check.vv  -check.f Test_VMChangeRootDiskSize .
START: api_vcd_test.go:372: TestVCD.SetUpSuite
Running on vCD https://10.150.170.30/api
as user administrator@System
PASS: api_vcd_test.go:372: TestVCD.SetUpSuite	20.154s

START: vm_test.go:656: TestVCD.Test_VMChangeRootDiskSize
vm_test.go:679:
    check.Assert(err, IsNil)
... value *errors.errorString = &errors.errorString{s:"error changing disk size: API Error: 400: [ 6ab9a023-06c8-48bb-a114-96c56c5d8408 ] The hard drives of the virtual machine \"Photon OS\" may not be modified, once created."} ("error changing disk size: API Error: 400: [ 6ab9a023-06c8-48bb-a114-96c56c5d8408 ] The hard drives of the virtual machine \"Photon OS\" may not be modified, once created.")

FAIL: vm_test.go:656: TestVCD.Test_VMChangeRootDiskSize
```
If you know of an OVA that I can use as a template to make this test succeed, please share.

cheald:
Hm. I wonder if it was allowed  because of my older vCD version. I'm using this live via Ubuntu, so it at least

Kube-ASY:
I'm curious if https://github.com/vmware/go-vcloud-director/pull/272 is a successor of this PR?

Didainius:
> I'm curious if #272 is a successor of this PR?

Yes

vbauzysvmware:
I think we can close this PR when #272 is ready.

vbauzysvmware:
PR with needed functionality is merged - please use new functionality.

PR #253, opened 2019-10-11
Deprecated find disk functions and added new get Disk functions

vbauzysvmware:
Changes for independent disk datasource/read/import in Terraform.
* Deprecated methods `VDC.FindDiskByHREF` and `FindDiskByHREF`
* Added methods `VDC.GetDiskByHref` `VDC.GetDiskByName` and related `GetDiskById`, `GetDiskByNameOrId`

### DIFF
--- a/govcd/disk.go
+++ b/govcd/disk.go
@@ -227,13 +227,20 @@ func (disk *Disk) Delete() (Task, error) {
 func (disk *Disk) Refresh() error {
 	util.Logger.Printf("[TRACE] Disk refresh, HREF: %s\n", disk.Disk.HREF)
 
-	fetchedDisk, err := FindDiskByHREF(disk.client, disk.Disk.HREF)
+	if *disk == (Disk{}) {
+		return fmt.Errorf("cannot refresh, Object is empty")
+	}
+
+	unmarshalledDisk := &types.Disk{}
+
+	_, err := disk.client.ExecuteRequest(disk.Disk.HREF, http.MethodGet,
+		"", "error refreshing independent disk: %s", nil, unmarshalledDisk)
 	if err != nil {
 		return err
 	}
+	disk.Disk = unmarshalledDisk
 
-	disk.Disk = fetchedDisk.Disk
-
+	// The request was successful
 	return nil
 }
 
@@ -287,6 +294,7 @@ func (disk *Disk) AttachedVM() (*types.Reference, error) {
 }
 
 // Find an independent disk by disk href in VDC
+// Deprecated: Use VDC.GetDiskByHref()
 func (vdc *Vdc) FindDiskByHREF(href string) (*Disk, error) {
 	util.Logger.Printf("[TRACE] VDC find disk By HREF: %s\n", href)
 
@@ -294,6 +302,7 @@ func (vdc *Vdc) FindDiskByHREF(href string) (*Disk, error) {
 }
 
 // Find an independent disk by VDC client and disk href
+// Deprecated: Use VDC.GetDiskByHref()
 func FindDiskByHREF(client *Client, href string) (*Disk, error) {
 	util.Logger.Printf("[TRACE] Find disk By HREF: %s\n", href)
 
@@ -338,4 +347,80 @@ func (vdc *Vdc) QueryDisk(diskName string) (DiskRecord, error) {
 	}
 
 	return *newDisk, nil
+}
+
+// GetDiskByHref finds a Disk by HREF
+// On success, returns a pointer to the Disk structure and a nil error
+// On failure, returns a nil pointer and an error
+func (vdc *Vdc) GetDiskByHref(diskHref string) (*Disk, error) {
+	util.Logger.Printf("[TRACE] Get Disk By Href: %s\n", diskHref)
+	Disk := NewDisk(vdc.client)
+
+	_, err := vdc.client.ExecuteRequest(diskHref, http.MethodGet,
+		"", "error retrieving Disk: %s", nil, Disk.Disk)
+	if err != nil {
+		return nil, err
+	}
+	return Disk, nil
+}
+
+// GetDiskByName finds a Disk by Name
+// On success, returns a pointer to the Disk structure and a nil error
+// On failure, returns a nil pointer and an error
+func (vdc *Vdc) GetDiskByName(diskName string, refresh bool) (*Disk, error) {
+	util.Logger.Printf("[TRACE] Get Disk By Name: %s\n", diskName)
+	if refresh {
+		err := vdc.Refresh()
+		if err != nil {
+			return nil, err
+		}
+	}
+	for _, resourceEntities := range vdc.Vdc.ResourceEntities {
+		for _, resourceEntity := range resourceEntities.ResourceEntity {
+			if resourceEntity.Name == diskName && resourceEntity.Type == "application/vnd.vmware.vcloud.disk+xml" {
+				return vdc.GetDiskByHref(resourceEntity.HREF)
+			}
+		}
+	}
+	return nil, ErrorEntityNotFound
+}
+
+// GetDiskById finds a Disk by ID
+// On success, returns a pointer to the Disk structure and a nil error
+// On failure, returns a nil pointer and an error
+func (vdc *Vdc) GetDiskById(diskId string, refresh bool) (*Disk, error) {
+	util.Logger.Printf("[TRACE] Get Disk By Id: %s\n", diskId)
+	if refresh {
+		err := vdc.Refresh()
+		if err != nil {
+			return nil, err
+		}
+	}
+	diskHREF := vdc.client.VCDHREF
+
+	DiskBareId, err := getBareEntityUuid(diskId)
+	if err != nil {
+		util.Logger.Printf("[Error] parsing bareID from diskId %s: %s", diskId, err)
+		return nil, ErrorEntityNotFound
+	}
+	if DiskBareId == "" {
+		util.Logger.Printf("[Error] parsing bareID from diskId %s - empty bareID returned", diskId)
+		return nil, ErrorEntityNotFound
+	}
+	diskHREF.Path += fmt.Sprintf("/disk/%s", DiskBareId)
+	return vdc.GetDiskByHref(diskHREF.String())
+}
+
+// GetDiskByNameOrId finds a Disk by Name or ID
+// On success, returns a pointer to the Disk structure and a nil error
+// On failure, returns a nil pointer and an error
+func (vdc *Vdc) GetDiskByNameOrId(identifier string, refresh bool) (*Disk, error) {
+	util.Logger.Printf("[TRACE] Get Disk By Name or Id: %s\n", identifier)
+	getByName := func(name string, refresh bool) (interface{}, error) { return vdc.GetDiskByName(name, refresh) }
+	getById := func(id string, refresh bool) (interface{}, error) { return vdc.GetDiskById(id, refresh) }
+	entity, err := getEntityByNameOrId(getByName, getById, identifier, refresh)
+	if entity == nil {
+		return nil, err
+	}
+	return entity.(*Disk), err
 }
--- a/govcd/disk.go
+++ b/govcd/disk.go
@@ -365,11 +365,11 @@ func (vdc *Vdc) GetDiskByHref(diskHref string) (*Disk, error) {
 }
 
 // GetDisksByName finds one or more Disks by Name
-// On success, returns a pointer to the Disk array and a nil error
+// On success, returns a pointer to the Disk list and a nil error
 // On failure, returns a nil pointer and an error
 func (vdc *Vdc) GetDisksByName(diskName string, refresh bool) (*[]Disk, error) {
 	util.Logger.Printf("[TRACE] Get Disk By Name: %s\n", diskName)
-	var diskArray []Disk
+	var diskList []Disk
 	if refresh {
 		err := vdc.Refresh()
 		if err != nil {
@@ -383,14 +383,14 @@ func (vdc *Vdc) GetDisksByName(diskName string, refresh bool) (*[]Disk, error) {
 				if err != nil {
 					return nil, err
 				}
-				diskArray = append(diskArray, *disk)
+				diskList = append(diskList, *disk)
 			}
 		}
 	}
-	if len(diskArray) == 0 {
+	if len(diskList) == 0 {
 		return nil, ErrorEntityNotFound
 	}
-	return &diskArray, nil
+	return &diskList, nil
 }
 
 // GetDiskById finds a Disk by ID
--- a/govcd/disk.go
+++ b/govcd/disk.go
@@ -358,7 +358,7 @@ func (vdc *Vdc) GetDiskByHref(diskHref string) (*Disk, error) {
 	Disk := NewDisk(vdc.client)
 
 	_, err := vdc.client.ExecuteRequest(diskHref, http.MethodGet,
-		"", "error retrieving Disk: %s", nil, Disk.Disk)
+		"", "error retrieving Disk: %#v", nil, Disk.Disk)
 	if err != nil && strings.Contains(err.Error(), "MajorErrorCode:403") {
 		return nil, ErrorEntityNotFound
 	}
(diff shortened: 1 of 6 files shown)


govcd/disk.go
```diff
@@ -227,13 +227,20 @@ func (disk *Disk) Delete() (Task, error) {
...
 
-	fetchedDisk, err := FindDiskByHREF(disk.client, disk.Disk.HREF)
+	if *disk == (Disk{}) {
+		return fmt.Errorf("cannot refresh, Object is empty")
+	}
+
+	unmarshalledDisk := &types.Disk{}
+
+	_, err := disk.client.ExecuteRequest(disk.Disk.HREF, http.MethodGet,
+		"", "error refreshing independent disk: %s", nil, unmarshalledDisk)
```
lvirbalas: Suggestion: use `GetDiskByHref` instead of the low-level `ExecuteRequest`, so all disk HTTP request logic goes through this single function.

vbauzysvmware: GetDiskByHref requires VDC, which we don't have here

lvirbalas: Don't we know VDC from Disk?

vbauzysvmware: If I would figure you vdc name, then I need to make request/search to fetch VDC object and then I can request that function

lvirbalas: Yeah, sounds introducing a hop - an additional call where we can live without it.

govcd/disk.go
```diff
@@ -339,3 +348,79 @@ func (vdc *Vdc) QueryDisk(diskName string) (DiskRecord, error) {
...
+
+	_, err := vdc.client.ExecuteRequest(diskHref, http.MethodGet,
+		"", "error retrieving Disk: %s", nil, Disk.Disk)
+	if err != nil {
+		return nil, err
+	}
+	return Disk, nil
+}
+
+// GetDiskByName finds a Disk by Name
```
dataclouder: This function (and `GetDiskByNameOrId`) may return more than one item. Unlike the other entities we have changed so far, disk allows name duplicates. Only search by ID or HREF guarantee returning a single item.
I suggest either removing the functions that can return duplicates or renaming and modify them to return an array, and then deal with the consequences.

govcd/disk_test.go
```diff
@@ -621,3 +608,64 @@ func (vcd *TestVCD) Test_QueryDisk(check *C) {
 	}
 
 }
+
+// Tests Disk retrieval by name, by ID, and by a combination of name and ID
```
dataclouder: if we decide to change the behavior of `GetDiskByName`, or to remove it, this testing function will not work anymore and the tests performed by `testFinderGetGenericEntities` should be performed explicitly.

govcd/disk.go
vbauzysvmware: From terraform side - we don't allow to create disk with same name, but govcd allows. We have options change `GetName` and `GetDiskByNameOrId` to throw error when two disk with same name exist. Or allow them to return arrays of results.

dataclouder: Users can create duplicate disks with other means. Using the GUI (either Flash or HTML5) there is no check for uniqueness.

vbauzysvmware: So we can have only these functions and slightly deviate from our *get patter
Get`Disks`ByName which returns aray
Get`Disk`ById
Get`Disk`ByHref

dataclouder: Fine with me. @lvirbalas and @Didainius may want to comment.

lvirbalas: I really like this idea of using plural in a function name which returns an array, it can be generalised as `Get<X>sByName`

Go ahead with the implementation.

@dataclouder please update coding guidelines:
- To reflect this agreement
- To write up `Query...` functions which are to be defined when `/query` API is called

vbauzysvmware: Code implemented

govcd/disk_test.go
vbauzysvmware: test changed

govcd/disk.go
```diff
@@ -339,3 +348,73 @@ func (vdc *Vdc) QueryDisk(diskName string) (DiskRecord, error) {
...
+
+	_, err := vdc.client.ExecuteRequest(diskHref, http.MethodGet,
+		"", "error retrieving Disk: %s", nil, Disk.Disk)
+	if err != nil {
+		return nil, err
+	}
+	return Disk, nil
+}
+
+// GetDisksByName finds a Disks by Name
```
dataclouder:
```suggestion
// GetDisksByName finds one or more Disks by Name
```

vbauzysvmware: updated

govcd/disk_test.go
```diff
@@ -46,16 +46,15 @@ func (vcd *TestVCD) Test_CreateDisk(check *C) {
 	check.Assert(task.Task.Owner.Type, Equals, types.MimeDisk)
 	diskHREF := task.Task.Owner.HREF
 
-	// Defer prepend the disk info to cleanup list until the function returns
-	defer PrependToCleanupList(fmt.Sprintf("%s|%s", diskCreateParamsDisk.Name, diskHREF), "disk", "", check.TestName())
+	AddToCleanupList(diskCreateParamsDisk.Name, "disk", diskHREF, check.TestName())
```
dataclouder: Two problems:
1. Why changing `defer Prepend` to `Add`? There is a reason for this usage, which becomes clear when running the whole test suite.
2. I don't like this change from the previous usage. We are using the `parent` field to store the HREF.
If we need a more complex structure, let's propose it, but using a field for what is not will make the code harder to understand.

govcd/disk_test.go
```diff
@@ -621,3 +608,70 @@ func (vcd *TestVCD) Test_QueryDisk(check *C) {
...
+	check.Assert(err, IsNil)
+
+	check.Assert(task.Task.Owner.Type, Equals, types.MimeDisk)
+	diskHREF = task.Task.Owner.HREF
+
+	AddToCleanupList(diskCreateParamsDisk.Name, "disk", diskHREF, check.TestName())
+
+	err = vcd.vdc.Refresh()
+	check.Assert(err, IsNil)
+	diskArray, err = vcd.vdc.GetDisksByName(diskName, false)
```
dataclouder: This value of `err` is not checked

govcd/entity_test.go
```diff
@@ -71,6 +71,9 @@ func (vapp *VApp) id() string   { return vapp.VApp.ID }
 func (vm *VM) name() string { return vm.VM.Name }
 func (vm *VM) id() string   { return vm.VM.ID }
 
+func (disk *Disk) name() string { return disk.Disk.Name }
```
dataclouder: If we don't use `testFinderGetGenericEntity` for Disk, implementing the `genericEntityInterface` is not needed

govcd/disk.go
```diff
@@ -339,3 +348,73 @@ func (vdc *Vdc) QueryDisk(diskName string) (DiskRecord, error) {
...
+// On success, returns a pointer to the Disk array and a nil error
```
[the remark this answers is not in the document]
vbauzysvmware: updated

govcd/disk.go
```diff
@@ -339,3 +348,73 @@ func (vdc *Vdc) QueryDisk(diskName string) (DiskRecord, error) {
...
+// On failure, returns a nil pointer and an error
+func (vdc *Vdc) GetDisksByName(diskName string, refresh bool) (*[]Disk, error) {
+	util.Logger.Printf("[TRACE] Get Disk By Name: %s\n", diskName)
+	var diskArray []Disk
```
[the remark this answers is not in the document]
vbauzysvmware: updated

govcd/disk_test.go
```diff
@@ -621,3 +608,70 @@ func (vcd *TestVCD) Test_QueryDisk(check *C) {
...
+
+	AddToCleanupList(diskCreateParamsDisk.Name, "disk", diskHREF, check.TestName())
+
+	// Wait for disk creation complete
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
+
+	err = vcd.vdc.Refresh()
+	check.Assert(err, IsNil)
+	diskArray, err := vcd.vdc.GetDisksByName(diskName, false)
```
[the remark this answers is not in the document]
vbauzysvmware: updated

govcd/disk_test.go
vbauzysvmware: updated

govcd/entity_test.go
vbauzysvmware: removed

govcd/disk_test.go
vbauzysvmware: 1. To be consistency as we use `Add` everywhere and `prepend` looked suspicious regarding behaviour against `Add`
2. Yes you are right in that sense. I can change it to `AddToCleanupList(diskHREF, "disk", "", check.TestName())`

govcd/disk.go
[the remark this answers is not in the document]
lvirbalas: I just cannot not share this (written in a really funny way):
https://medium.com/@marty.stepien/arrays-vs-slices-bonanza-in-golang-fa8d32cd2b7c

govcd/disk_test.go
vbauzysvmware: updated

govcd/disk.go
```diff
@@ -339,3 +348,73 @@ func (vdc *Vdc) QueryDisk(diskName string) (DiskRecord, error) {
...
+	util.Logger.Printf("[TRACE] Get Disk By Id: %s\n", diskId)
+	if refresh {
+		err := vdc.Refresh()
+		if err != nil {
+			return nil, err
+		}
+	}
+	diskHREF := vdc.client.VCDHREF
+
+	DiskBareId, err := getBareEntityUuid(diskId)
```
Didainius: Any reason for capital var naming?

govcd/entity_test.go
```diff
@@ -219,6 +218,5 @@ func (vcd *TestVCD) testFinderGetGenericEntity(def getterTestDefinition, check *
 	ge, err = def.getByNameOrId(invalidEntityId, false)
 	entity8 := ge.(genericEntity)
 	check.Assert(err, NotNil)
-	check.Assert(IsNotFound(err), Equals, true)
```
Didainius: Is there a specific reason to remove this check (and the above one)? I think other tests using this `testFinderGetGenericEntity` rely and satisfy this depedency. Aren't we "relieving" this for all entities?

govcd/entity_test.go
```diff
@@ -209,7 +209,6 @@ func (vcd *TestVCD) testFinderGetGenericEntity(def getterTestDefinition, check *
 	ge, err = def.getById(invalidEntityId, false)
 	entity7 := ge.(genericEntity)
 	check.Assert(err, NotNil)
-	check.Assert(IsNotFound(err), Equals, true)
```
Didainius: Same question as in line 222

govcd/disk.go
```diff
@@ -339,3 +348,73 @@ func (vdc *Vdc) QueryDisk(diskName string) (DiskRecord, error) {
...
+// On failure, returns a nil pointer and an error
+func (vdc *Vdc) GetDiskById(diskId string, refresh bool) (*Disk, error) {
+	util.Logger.Printf("[TRACE] Get Disk By Id: %s\n", diskId)
+	if refresh {
+		err := vdc.Refresh()
+		if err != nil {
+			return nil, err
+		}
+	}
+	diskHREF := vdc.client.VCDHREF
```
Didainius: Either I got it wrong, either this var name implies some readability problem. Is this really a diskHREF stored? Is it not becoming a diskHREF below in line 418 (`diskHREF.Path += fmt.Sprintf("/disk/%s", DiskBareId)`)

govcd/disk.go
```diff
@@ -227,13 +227,20 @@ func (disk *Disk) Delete() (Task, error) {
 func (disk *Disk) Refresh() error {
 	util.Logger.Printf("[TRACE] Disk refresh, HREF: %s\n", disk.Disk.HREF)
 
-	fetchedDisk, err := FindDiskByHREF(disk.client, disk.Disk.HREF)
+	if *disk == (Disk{}) {
```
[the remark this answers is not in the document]
vbauzysvmware: I change to href, just for note, if disk is nil you won't access function :)

govcd/disk.go
vbauzysvmware: fixed

govcd/entity_test.go
vbauzysvmware: It questionable. My decision was to remove cause only specific one object can use that, all other entitities can't comply as Get*ByHref is direct call and would return 403 and not found error.

govcd/disk.go
vbauzysvmware: Moved to better place

govcd/disk.go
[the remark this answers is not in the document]
Didainius: Thanks. Oddly enough I think it can be accessed if it was defined like this:
`var myDisk *Disk`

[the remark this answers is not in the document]
vbauzysvmware: Hmm, not follow you - we just checking that receiver isn't empty. I updated check to be more robust. Will push after runing the tests.

govcd/entity_test.go
dataclouder: This check **must not** be removed.
`GetDisk*` functions will not use it. And if other entities fail this check, it means that they were not well defined.

The failure here must return a ErrEntityNotFound because the code should not reach `Get*ByHREF`. If it does, the implementation should be reviewed.

govcd/disk.go
dataclouder: This code can fail easily.  I prefer the following.
```suggestion
	if disk.Disk != nil && disk.Disk.HREF != "" {
```

govcd/entity_test.go
vbauzysvmware: GetDisk don't use this generic method, but media has same approach. Difference is that no all GetHref methods works in the way that fetch all elements and filter out by ==href. Normally you just call API ir return error if Href is wrong.

govcd/disk.go
vbauzysvmware: changed

govcd/entity_test.go
dataclouder: Please show me where you think it is inappropriate and I will see if the test can be improved. But until further agreement, this test works as defined and we cannot change it.

vbauzysvmware: GetExternalNetworkById vs GetMediaById

dataclouder: GetMediaById can be improved, as commented in its PR. 
If there is still a problem with it, it will be solved in such place. As for the current PR, this check should be reinstated.

govcd/entity_test.go
```diff
@@ -78,7 +78,7 @@ func (vm *VM) id() string   { return vm.VM.ID }
 // Get invalid name or ID
 // To use this function, the entity must satisfy the interface genericEntity
 // and within the caller it must define the getter functions
-//
+//AddToCleanupList(diskCreateParamsDisk.Name
```
dataclouder: This is not a requirement to run these tests. Please remove.

govcd/entity_test.go
vbauzysvmware: reverted

govcd/entity_test.go
vbauzysvmware: reverted

govcd/entity_test.go
vbauzysvmware: removed

vbauzysvmware:
code updated

govcd/disk_test.go
```diff
@@ -683,3 +687,53 @@ func (vcd *TestVCD) Test_GetDisks(check *C) {
 	check.Assert(len(*diskList), Equals, 2)
 
 }
+
+// Tests Disk list retrieval by name, by ID
+func (vcd *TestVCD) Test_GetDiskByHref(check *C) {
+	fmt.Printf("Running: %s\n", check.TestName())
+
+	if vcd.config.VCD.Disk.Size == 0 {
+		check.Skip("Skipping test because disk size is <= 0")
```
dataclouder: please remove "<="

PR #255, opened 2019-10-22
Add QueryDisks

vbauzysvmware:
* Add QueryDisks using name
Function needed to handle independent disk read/import by Id

### DIFF
--- a/govcd/disk.go
+++ b/govcd/disk.go
@@ -350,6 +350,31 @@ func (vdc *Vdc) QueryDisk(diskName string) (DiskRecord, error) {
 	return *newDisk, nil
 }
 
+// Find independent disks using disk name. Returns list of VMRecord query type
+func (vdc *Vdc) QueryDisks(diskName string) (*[]*types.DiskRecordType, error) {
+
+	if diskName == "" {
+		return nil, fmt.Errorf("disk name can not be empty")
+	}
+
+	typeMedia := "disk"
+	if vdc.client.IsSysAdmin {
+		typeMedia = "adminDisk"
+	}
+
+	results, err := vdc.QueryWithNotEncodedParams(nil, map[string]string{"type": typeMedia, "filter": "name==" + url.QueryEscape(diskName)})
+	if err != nil {
+		return nil, fmt.Errorf("error querying disk %#v", err)
+	}
+
+	diskResults := results.Results.DiskRecord
+	if vdc.client.IsSysAdmin {
+		diskResults = results.Results.AdminDiskRecord
+	}
+
+	return &diskResults, nil
+}
+
 // GetDiskByHref finds a Disk by HREF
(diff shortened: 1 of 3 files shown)


govcd/disk.go
```diff
@@ -350,6 +350,31 @@ func (vdc *Vdc) QueryDisk(diskName string) (DiskRecord, error) {
 	return *newDisk, nil
 }
 
+// Find independent disks using disk name. Returns list of VMRecord query type
+func (vdc *Vdc) QueryDisks(diskName string) (*[]*types.DiskRecordType, error) {
+
+	if diskName == "" {
+		return nil, fmt.Errorf("disk name can not be empty")
```
dataclouder: "can not" => "can't"

govcd/disk.go
```diff
@@ -350,6 +350,31 @@ func (vdc *Vdc) QueryDisk(diskName string) (DiskRecord, error) {
...
+	}
+
+	typeMedia := "disk"
+	if vdc.client.IsSysAdmin {
+		typeMedia = "adminDisk"
+	}
+
+	results, err := vdc.QueryWithNotEncodedParams(nil, map[string]string{"type": typeMedia, "filter": "name==" + url.QueryEscape(diskName)})
+	if err != nil {
+		return nil, fmt.Errorf("error querying disk %#v", err)
```
dataclouder: "%#v" => "%s"

govcd/disk.go
```diff
@@ -350,6 +350,31 @@ func (vdc *Vdc) QueryDisk(diskName string) (DiskRecord, error) {
...
+	}
+
+	diskResults := results.Results.DiskRecord
+	if vdc.client.IsSysAdmin {
+		diskResults = results.Results.AdminDiskRecord
```
dataclouder: What's the difference between `DiskRecord` and `AdminDiskRecord`? 
Does the caller need to know?

govcd/disk.go
vbauzysvmware: updated

govcd/disk.go
vbauzysvmware: updated

govcd/disk.go
vbauzysvmware: It is how you should call and what type receive to/from Query with different type of users `system` -`none system`. I made it that way that user don't know the details.

govcd/disk.go
```diff
@@ -350,6 +350,31 @@ func (vdc *Vdc) QueryDisk(diskName string) (DiskRecord, error) {
 	return *newDisk, nil
 }
 
+// Find independent disks using disk name. Returns list of VMRecord query type
```
Didainius: go doc format
```suggestion
// QueryDisks ...
```

vbauzysvmware: tx. Updated

PR #256, opened 2019-10-24
Removed parentheses from filtering

vbauzysvmware:
Fix for https://github.com/terraform-providers/terraform-provider-vcd/issues/287

Parentheses isn't needed(accordig documentation used when need to create sophisticated filtering) and if we will need to use them later - we have to figure out how to escape them, as in different env behaves differently.

Building query docs: https://pubs.vmware.com/vcloud-api-1-5/wwhelp/wwhimpl/js/html/wwhelp.htm#href=api_prog/GUID-CDF04296-5EB5-47E1-9BEC-228837C584CE.html#1_22_9_1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 to allow understand if value was returned or not. 
 * Added method VApp.AddNewVMWithStorageProfile that adds a VM with custom storage profile.
 
+BUGS FIXED:
+* Removed parentheses from filtering which in some env wasn't treated correctly [#256] (https://github.com/vmware/go-vcloud-director/pull/256)
+
 ## 2.4.0 (October 28, 2019)
 
 * Deprecated functions `GetOrgByName` and `GetAdminOrgByName`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ to allow understand if value was returned or not.
 * Added method VApp.AddNewVMWithStorageProfile that adds a VM with custom storage profile.
 
 BUGS FIXED:
-* Removed parentheses from filtering which in some env wasn't treated correctly [#256] (https://github.com/vmware/go-vcloud-director/pull/256)
+* Remove parentheses from filtering since they weren't treated correctly in some environment [#256] (https://github.com/vmware/go-vcloud-director/pull/256)
 
 ## 2.4.0 (October 28, 2019)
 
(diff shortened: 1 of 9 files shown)


govcd/api.go
```diff
@@ -122,7 +122,7 @@ func ContainsNotFound(err error) bool {
 	return err != nil && strings.Contains(err.Error(), ErrorEntityNotFound.Error())
 }
 
-// Function allow to pass complex values params which shouldn't be encoded like for queries. e.g. /query?filter=(name=foo)
+// Function allow to pass complex values params which shouldn't be encoded like for queries. e.g. /query?filter=name=foo
```
dataclouder:
```suggestion
// NewRequestWitNotEncodedParams allows passing complex values params that shouldn't be encoded like for queries. e.g. /query?filter=name=foo
```

vbauzysvmware:
> Looks to be dead simple.
> I was just looking at a more recent version of the docs: https://code.vmware.com/docs/8072/vcloud-director-api-programming-guide/GUID-CDF04296-5EB5-47E1-9BEC-228837C584CE.html
> 
> Main concern from me, is the "grouping" (parenthesis) not used so that it matches both conditions (logical **and**) while without parenthesis would be logical **or**?
> 
> In such case a query with multiple fields may have different results or am I wrong?
> 
> This one would look for objects matching both conditions - `name` to be _"name"_ **and** `orgName` to be _"adminOrg.AdminOrg.Name"_:
> 
> ```go
> "filter": fmt.Sprintf("(name==%s;orgName==%s)", url.QueryEscape(name), url.QueryEscape(adminOrg.AdminOrg.Name)),
> ```
> 
> While this one would look for objects whose `name` matches _"name"_ **or** `orgName` matches _"adminOrg.AdminOrg.Name"_
> 
> ```go
> "filter": fmt.Sprintf("name==%s;orgName==%s", url.QueryEscape(name), url.QueryEscape(adminOrg.AdminOrg.Name)),
> ```

We can't 100% rely on latest docs as we working still with older versions.
Regarding question `Logical OR` and `Logical AND` is managed by separator `,` and `;` .

govcd/api.go
vbauzysvmware: update

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 to allow understand if value was returned or not. 
 * Added method VApp.AddNewVMWithStorageProfile that adds a VM with custom storage profile.
 
+BUGS FIXED:
+* Removed parentheses from filtering which in some env wasn't treated correctly [#256] (https://github.com/vmware/go-vcloud-director/pull/256)
```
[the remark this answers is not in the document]
vbauzysvmware: updated

dataclouder:
> Looks to be dead simple.
> I was just looking at a more recent version of the docs: https://code.vmware.com/docs/8072/vcloud-director-api-programming-guide/GUID-CDF04296-5EB5-47E1-9BEC-228837C584CE.html
> 
> Main concern from me, is the "grouping" (parenthesis) not used so that it matches both conditions (logical **and**) while without parenthesis would be logical **or**?
> 
> In such case a query with multiple fields may have different results or am I wrong?
> 
> This one would look for objects matching both conditions - `name` to be _"name"_ **and** `orgName` to be _"adminOrg.AdminOrg.Name"_:
> 
> ```go
> "filter": fmt.Sprintf("(name==%s;orgName==%s)", url.QueryEscape(name), url.QueryEscape(adminOrg.AdminOrg.Name)),
> ```
> 
> While this one would look for objects whose `name` matches _"name"_ **or** `orgName` matches _"adminOrg.AdminOrg.Name"_
> 
> ```go
> "filter": fmt.Sprintf("name==%s;orgName==%s", url.QueryEscape(name), url.QueryEscape(adminOrg.AdminOrg.Name)),
> ```

Parentheses are needed only if we have groups, or sub-expression.
For example, `"(A=1;B=2),(A=2;B=3)"` means (A == 1 AND B == 2) OR (A ==2 AND B == 3)

If we have only one set of parentheses, they should be superfluous, as we are grouping the only sub-expression in the condition.
(Notice that in your example the comma should not be followed by spaces)

Didainius:
> Parentheses are needed only if we have groups, or sub-expression.
> For example, `"(A=1;B=2),(A=2;B=3)"` means (A == 1 AND B == 2) OR (A ==2 AND B == 3)
> 
> If we have only one set of parentheses, they should be superfluous, as we are grouping the only sub-expression in the condition.
> (Notice that in your example the comma should not be followed by spaces)

Alright. I am sold :)

The example in website confused me initially having `and` everywhere:
```
(attribute1==value1;attribute2!=value2);(attribute3==value3;attribute4!=value4)...
```

PR #257, opened 2019-10-28
Fix missing fields in user data

dataclouder:
Add missing field "telephone" to user data.
This change will fix terraform-provider-vcd User resource when that repository will point at the updated tag after we merge.


dataclouder:
> The always funny part - a note in Changelog about `CreateUserSimple` fixing previously ignored field.

Thanks. Done

PR #258, opened 2019-11-04
Change for handling state of vdc

vbauzysvmware:
ref: https://github.com/terraform-providers/terraform-provider-vcd/pull/380

Changes needed to handle system admin and org admin diff set data provided from API.

This allows to know when provisioning isn't returned.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 2.5.0 
 
-* Change fields ResourceGuaranteedCpu, VCpuInMhz, IsThinProvision, NetworkPoolReference,ProviderVdcReference and UsesFastProvisioning in AdminVdc to pointers
+* Change fields ResourceGuaranteedCpu, VCpuInMhz, IsThinProvision, NetworkPoolReference, ProviderVdcReference and UsesFastProvisioning in AdminVdc to pointers
+to allow understand if value was returned or not. 
 
 ## 2.4.0 (October 28, 2019)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.5.0 
+
+* Change fields ResourceGuaranteedCpu, VCpuInMhz, IsThinProvision, NetworkPoolReference,ProviderVdcReference and UsesFastProvisioning in AdminVdc to pointers
+
 ## 2.4.0 (October 28, 2019)
 
 * Deprecated functions `GetOrgByName` and `GetAdminOrgByName`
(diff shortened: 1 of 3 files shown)


vbauzysvmware:
> Thanks for it. The only question is - are these fields enough or could there be more bools needed to convert to pointers.
> Also - a small note in changelog.

Found one more.

types/v56/types.go
```diff
@@ -398,11 +398,11 @@ type AdminVdc struct {
...
 	ResourceGuaranteedMemory *float64   `xml:"ResourceGuaranteedMemory,omitempty"`
 	ResourceGuaranteedCpu    *float64   `xml:"ResourceGuaranteedCpu,omitempty"`
-	VCpuInMhz                int64      `xml:"VCpuInMhz,omitempty"`
-	IsThinProvision          bool       `xml:"IsThinProvision,omitempty"`
+	VCpuInMhz                *int64     `xml:"VCpuInMhz,omitempty"`
+	IsThinProvision          *bool      `xml:"IsThinProvision,omitempty"`
 	NetworkPoolReference     *Reference `xml:"NetworkPoolReference,omitempty"`
 	ProviderVdcReference     *Reference `xml:"ProviderVdcReference"`
-	UsesFastProvisioning     bool       `xml:"UsesFastProvisioning,omitempty"`
+	UsesFastProvisioning     *bool      `xml:"UsesFastProvisioning,omitempty"`
```
dataclouder: Why do we have two pointers to Boolean and two straight Booleans in the same structure?
They are represented in the same way in XML.

vbauzysvmware: This is requeste in ref: terraform-providers/terraform-provider-vcd#380

Issue in terraform that in state file we should know if it is false or wasn't provided

dataclouder: This doesn't answer the question. Why two pointers and two straight ones?

vbauzysvmware: With pointers we know that won't be returned from API when called by org admin

dataclouder: I know what pointers are used for. 
My question is why we we two Booleans and two Boolean pointers in the same structure.
Why not all pointers?

vbauzysvmware: I mean. I made pointers which I know API will return or not according role, on others I don't have info

dataclouder: Thanks.
FYI: the field `OverCommitAllowed` is ignored both in creation and in update.

vbauzysvmware:
> LGTM, although I think that we should be more consistent with the fields in VDC data structure.
> We should treat all Boolean fields in the same way.

Thank you for review. P.s. it's not just booleans. Also I feel that we may come back later to this, that all values should be pointers to know if they were provided by API or not.

vbauzysvmware:
> Request and a question:
> 
>     * Change log missing
> 
>     * Why we are not adding add pointers to OverCommitAllowed and VmDiscoveryEnabled too?

* changelog added
* these two values aren't used and seen xml

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.5.0 
+
+* Change fields ResourceGuaranteedCpu, VCpuInMhz, IsThinProvision, NetworkPoolReference,ProviderVdcReference and UsesFastProvisioning in AdminVdc to pointers
```
lvirbalas: Please add also why this change is needed. Also, there's a space missing after one comma.

vbauzysvmware: done

PR #262, opened 2019-11-13
Add ability to connect using authorization tokens

dataclouder:
* Added method `VCDClient.SetToken`
* Added method `VCDClient.GetAuthResponse`
* Added script `scripts/get_token.sh`

Implementation notes:

* When using a token, we don't know by which user it was created. This may not be a problem for most users, but it is for a team like ours, where we need to test with different users.
* Using a token generated by an org admin for tasks that require an admin user will fail.
* Similarly, a token shows no relationship with the server that generated it. Using in server X a token that was generated for server Y will fail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,5 +134,5 @@ For the token, you use:
 	err := vcdClient.SetToken(Org, govcd.AuthorizationHeader, Token)
 ```
 
-The file `scripts/get_token.sh` provides a handy method of extracting the token for future use.
+The file `scripts/get_token.sh` provides a handy method of extracting the token (`x-vcloud-authorization` value) for future use.
 
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ For the first two methods, you use:
 
 ```go
 	err := vcdClient.Authenticate(User, Password, Org)
-    // or
+	// or
 	resp, err := vcdClient.GetAuthResponse(User, Password, Org)
 ```
 
--- a/README.md
+++ b/README.md
@@ -96,12 +96,12 @@ func main() {
 		fmt.Println(err)
 		os.Exit(1)
 	}
-	org, err := govcd.GetOrgByName(client, config.Org)
+	org, err := client.GetOrgByName(config.Org)
 	if err != nil {
 		fmt.Println(err)
 		os.Exit(1)
 	}
-	vdc, err := org.GetVdcByName(config.VDC)
+	vdc, err := org.GetVDCByName(config.VDC)
 	if err != nil {
 		fmt.Println(err)
 		os.Exit(1)
@@ -116,8 +116,8 @@ func main() {
 
 You can authenticate to the vCD in three ways:
 
-* With a system administration user and password (`administrator@system`)
-* With an Organization user and password (`tenant-admin@org_name)
+* With a System Administration user and password (`administrator@system`)
+* With an Organization user and password (`tenant-admin@org-name`)
 * With an authorization token
 
 For the first two methods, you use:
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -111,3 +111,28 @@ func main() {
 }
 
 ```
+
+## Authentication
+
+You can authenticate to the vCD in three ways:
+
+* With a system administration user and password (`administrator@system`)
```
lvirbalas:
```suggestion
* With a System Administration user and password (`administrator@system`)
```

README.md
```diff
@@ -111,3 +111,28 @@ func main() {
...
+* With an Organization user and password (`tenant-admin@org_name)
```
lvirbalas:
```suggestion
* With an Organization user and password (`tenant-admin@org-name)
```

scripts/get_token.sh
```diff
@@ -0,0 +1,25 @@
...
+# and show the header that cointains an authorization token.
+#
+user=$1
+password=$2
+org=$3
+IP=$4
+
+if [ -z "$IP" ]
+then
+    echo "Syntax $0 user password organization IP_address"
```
lvirbalas: Can one use hostname here? If yes, I suggest we say "hostname" here as there should be more cases where user has hostname as opposed to IP (think of all the tenants renting the vCDs at SPs).

dataclouder: updated

README.md
dataclouder: updated

README.md
dataclouder: updated

scripts/get_token.sh
```diff
@@ -0,0 +1,25 @@
...
+# and show the header that cointains an authorization token.
+#
+user=$1
+password=$2
+org=$3
+IP=$4
+
+if [ -z "$IP" ]
+then
+    echo "Syntax $0 user password organization hostname_or_IP_address)"
```
lvirbalas:
```suggestion
    echo "Syntax $0 user password organization hostname_or_IP_address"
```

dataclouder: Fixed

scripts/get_token.sh
```diff
@@ -0,0 +1,25 @@
...
+
+if [ -z "$IP" ]
+then
+    echo "Syntax $0 user password organization hostname_or_IP_address"
+    exit 1
+fi
+
+auth=$(echo -n "$user@$org:$password" |base64)
+
+curl -I -k --header "Accept: application/*;version=27.0" \
```
Didainius: If we're going away to v29 - we could have it in sync I guess

vbauzysvmware: I recommend to also to change to 
```suggestion
curl -I -k --header "Accept: application/*;version=29.0" \
```

dataclouder: Changed in in both PRs, although the result would be the same.

README.md
```diff
@@ -111,3 +111,28 @@ func main() {
...
+	resp, err := vcdClient.GetAuthResponse(User, Password, Org)
+```
+
+For the token, you use:
+
+```go
+	err := vcdClient.SetToken(Org, govcd.AuthorizationHeader, Token)
+```
+
+The file `scripts/get_token.sh` provides a handy method of extracting the token for future use.
```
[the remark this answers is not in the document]
dataclouder: done

scripts/get_token.sh
vbauzysvmware: That's the very good.

PR #263, opened 2019-11-13
Increment vCD API version to 29.0

vbauzysvmware:
Ref: We dropped support for vCD 8.2 and for new functionality we need access features which are on higher API versions.
Documentaion :https://code.vmware.com/apis/220/vcloud/doc/diff/index.html

* Increase from 27.0 (vCD 8.2) to 29 (vCD 9.0)
* Updated some types to correspond new API

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,9 @@
 * Added method `VCDClient.GetAuthResponse`
 * Added script `scripts/get_token.sh`
 * Increment vCD API version used to 29.0
-* Removed fields `VdcEnabled`, `VAppParentHREF`, `VAppParentName`, `HighestSupportedVersion`, `VmToolsVersion`, `TaskHREF`, `TaskStatusName`, `TaskDetails`, `TaskStatus` from `QueryResultVMRecordType`
-* Added fields `ID, Type, ContainerName, ContainerID, OwnerName, Owner, NetworkHref, IpAddress, CatalogName, VmToolsStatus, GcStatus, AutoUndeployDate, AutoDeleteDate, AutoUndeployNotified, AutoDeleteNotified, Link,	MetaData` to `QueryResultVMRecordType`
+** Removed fields `VdcEnabled`, `VAppParentHREF`, `VAppParentName`, `HighestSupportedVersion`, `VmToolsVersion`, `TaskHREF`, `TaskStatusName`, `TaskDetails`, `TaskStatus` from `QueryResultVMRecordType`
+** Added fields `ID, Type, ContainerName, ContainerID, OwnerName, Owner, NetworkHref, IpAddress, CatalogName, VmToolsStatus, GcStatus, AutoUndeployDate, AutoDeleteDate, AutoUndeployNotified, AutoDeleteNotified, Link, MetaData` to `QueryResultVMRecordType`, `DistributedInterface` to NetworkConfiguration, `RegenerateBiosUuid` to VMGeneralParams
+** Change `DistributedRoutingEnabled` to pointer in GatewayConfiguration 
 
 BUGS FIXED:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 * Increment vCD API version used to 29.0
 ** Removed fields `VdcEnabled`, `VAppParentHREF`, `VAppParentName`, `HighestSupportedVersion`, `VmToolsVersion`, `TaskHREF`, `TaskStatusName`, `TaskDetails`, `TaskStatus` from `QueryResultVMRecordType`
 ** Added fields `ID, Type, ContainerName, ContainerID, OwnerName, Owner, NetworkHref, IpAddress, CatalogName, VmToolsStatus, GcStatus, AutoUndeployDate, AutoDeleteDate, AutoUndeployNotified, AutoDeleteNotified, Link, MetaData` to `QueryResultVMRecordType`, `DistributedInterface` to NetworkConfiguration, `RegenerateBiosUuid` to VMGeneralParams
-** Change `DistributedRoutingEnabled` to pointer in GatewayConfiguration 
+** Change `DistributedRoutingEnabled` to pointer in GatewayConfiguration and `DistributedInterface` in NetworkConfiguration.
 
 BUGS FIXED:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 * Increment vCD API version used to 29.0
     * Removed fields `VdcEnabled`, `VAppParentHREF`, `VAppParentName`, `HighestSupportedVersion`, `VmToolsVersion`, `TaskHREF`, `TaskStatusName`, `TaskDetails`, `TaskStatus` from `QueryResultVMRecordType`
     * Added fields `ID, Type, ContainerName, ContainerID, OwnerName, Owner, NetworkHref, IpAddress, CatalogName, VmToolsStatus, GcStatus, AutoUndeployDate, AutoDeleteDate, AutoUndeployNotified, AutoDeleteNotified, Link, MetaData` to `QueryResultVMRecordType`, `DistributedInterface` to `NetworkConfiguration` and `RegenerateBiosUuid` to `VMGeneralParams`
-    * Change `DistributedRoutingEnabled` to pointer in `GatewayConfiguration` and `DistributedInterface` in `NetworkConfiguration`.
+    * Change to pointers `DistributedRoutingEnabled` in `GatewayConfiguration` and `DistributedInterface` in `NetworkConfiguration`
 
 BUGS FIXED:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 * Added method `VCDClient.SetToken`
 * Added method `VCDClient.GetAuthResponse`
 * Added script `scripts/get_token.sh`
-* Increment vCD API version used to 29.0
+* Increment vCD API version used from 27.0 to 29.0
     * Remove fields `VdcEnabled`, `VAppParentHREF`, `VAppParentName`, `HighestSupportedVersion`, `VmToolsVersion`, `TaskHREF`, `TaskStatusName`, `TaskDetails`, `TaskStatus` from `QueryResultVMRecordType`
     * Add fields `ID, Type, ContainerName, ContainerID, OwnerName, Owner, NetworkHref, IpAddress, CatalogName, VmToolsStatus, GcStatus, AutoUndeployDate, AutoDeleteDate, AutoUndeployNotified, AutoDeleteNotified, Link, MetaData` to `QueryResultVMRecordType`, `DistributedInterface` to `NetworkConfiguration` and `RegenerateBiosUuid` to `VMGeneralParams`
     * Change to pointers `DistributedRoutingEnabled` in `GatewayConfiguration` and `DistributedInterface` in `NetworkConfiguration`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,9 @@
 * Added method `VCDClient.GetAuthResponse`
 * Added script `scripts/get_token.sh`
 * Increment vCD API version used to 29.0
-** Removed fields `VdcEnabled`, `VAppParentHREF`, `VAppParentName`, `HighestSupportedVersion`, `VmToolsVersion`, `TaskHREF`, `TaskStatusName`, `TaskDetails`, `TaskStatus` from `QueryResultVMRecordType`
-** Added fields `ID, Type, ContainerName, ContainerID, OwnerName, Owner, NetworkHref, IpAddress, CatalogName, VmToolsStatus, GcStatus, AutoUndeployDate, AutoDeleteDate, AutoUndeployNotified, AutoDeleteNotified, Link, MetaData` to `QueryResultVMRecordType`, `DistributedInterface` to `NetworkConfiguration` and `RegenerateBiosUuid` to `VMGeneralParams`
-** Change `DistributedRoutingEnabled` to pointer in `GatewayConfiguration` and `DistributedInterface` in `NetworkConfiguration`.
+    * Removed fields `VdcEnabled`, `VAppParentHREF`, `VAppParentName`, `HighestSupportedVersion`, `VmToolsVersion`, `TaskHREF`, `TaskStatusName`, `TaskDetails`, `TaskStatus` from `QueryResultVMRecordType`
+    * Added fields `ID, Type, ContainerName, ContainerID, OwnerName, Owner, NetworkHref, IpAddress, CatalogName, VmToolsStatus, GcStatus, AutoUndeployDate, AutoDeleteDate, AutoUndeployNotified, AutoDeleteNotified, Link, MetaData` to `QueryResultVMRecordType`, `DistributedInterface` to `NetworkConfiguration` and `RegenerateBiosUuid` to `VMGeneralParams`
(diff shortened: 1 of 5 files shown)


govcd/edgegateway_test.go
```diff
@@ -743,31 +743,24 @@ func testCheckFirewallConfig(beforeFw *types.FirewallConfigWithXml, beforeFwXml
...
+	// and ruleTag changed during the test and we must ignore this change while comparing.
+	beforeFw.FirewallRules.Text = replaceAllMatches(beforeFw.FirewallRules.Text, `<id>\d*</id>`, "<id>99999</id>")
+	beforeFw.FirewallRules.Text = replaceAllMatches(beforeFw.FirewallRules.Text, `<ruleTag>\d*</ruleTag>`, "<ruleTag>99999</ruleTag>")
+	afterFw.FirewallRules.Text = replaceAllMatches(afterFw.FirewallRules.Text, `<id>\d*</id>`, "<id>99999</id>")
+	afterFw.FirewallRules.Text = replaceAllMatches(afterFw.FirewallRules.Text, `<ruleTag>\d*</ruleTag>`, "<ruleTag>99999</ruleTag>")
+
+	beforeFwXml = replaceAllMatches(beforeFwXml, `<id>\d*</id>`, "<id>99999</id>")
+	beforeFwXml = replaceAllMatches(beforeFwXml, `<ruleTag>\d*</ruleTag>`, "<ruleTag>99999</ruleTag>")
+	afterFwXml = replaceAllMatches(afterFwXml, `<id>\d*</id>`, "<id>99999</id>")
+	afterFwXml = replaceAllMatches(afterFwXml, `<ruleTag>\d*</ruleTag>`, "<ruleTag>99999</ruleTag>")
```
lvirbalas: Curious, what changed in API that you had to switch from `replaceFirstMatch` to `replaceAllMatches`? Thanks!

types/v56/types.go
```diff
@@ -1778,7 +1780,7 @@ type LbPoolMember struct {
 	Port        int    `xml:"port"`
 	MaxConn     int    `xml:"maxConn,omitempty"`
 	MinConn     int    `xml:"minConn,omitempty"`
-	Condition   string `xml:"condition,omitempty"`
+	Condition   string `xml:"condition"`
```
lvirbalas: Interesting change. Curious, why only this field got it, but others above not?

CC: @Didainius

types/v56/types.go
```diff
@@ -2240,31 +2242,40 @@ type QueryResultEdgeGatewayRecordType struct {
...
-	Deleted                 bool   `xml:"isDeleted,attr,omitempty"`
-	MaintenanceMode         bool   `xml:"isInMaintenanceMode,attr,omitempty"`
-	Published               bool   `xml:"isPublished,attr,omitempty"`
-	VAppTemplate            bool   `xml:"isVAppTemplate,attr,omitempty"`
-	VdcEnabled              bool   `xml:"isVdcEnabled,attr,omitempty"`
-	VdcHREF                 string `xml:"vdc,attr,omitempty"`
-	VAppParentHREF          string `xml:"container,attr,omitempty"`
-	VAppParentName          string `xml:"containerName,attr,omitempty"`
-	HardwareVersion         int    `xml:"hardwareVersion,attr,omitempty"`
-	HighestSupportedVersion int    `xml:"pvdcHighestSupportedHardwareVersion,attr,omitempty"`
```
lvirbalas: Difficult to see in the diff, as formatting changed. Is the only change here that `HighestSupportedVersion` was removed from API in 29.0?

types/v56/types.go
```diff
@@ -2240,31 +2242,40 @@ type QueryResultEdgeGatewayRecordType struct {
...
-	VmToolsVersion          string `xml:"vmToolsVersion,attr,omitempty"`
-	GuestOS                 string `xml:"guestOs,attr,omitempty"`
-	MemoryMB                int    `xml:"memoryMB,attr,omitempty"`
-	Cpus                    int    `xml:"numberOfCpus,attr,omitempty"`
-	StorageProfileName      string `xml:"storageProfileName,attr,omitempty"`
-	NetworkName             string `xml:"networkName,attr,omitempty"`
-	TaskHREF                string `xml:"task,attr,omitempty"`
-	TaskStatusName          string `xml:"taskStatusName,attr,omitempty"`
-	TaskDetails             string `xml:"taskDetails,attr,omitempty"`
-	TaskStatus              string `xml:"TaskStatus,attr,omitempty"`
```
lvirbalas: Difficult to see the diff, as formatting changed, but as I understand these fields got removed in API 29.0:
- VdcEnabled
- VAppParentHREF
- VAppParentName
- HighestSupportedVersion
- VmToolsVersion
- TaskHREF
- TaskStatusName
- TaskDetails
- TaskStatus

Few questions:
1. Do we use them in Terraform?
2. Where did the task go?

CC: @dataclouder @Didainius

govcd/api_vcd.go
```diff
@@ -87,12 +87,12 @@ func (vcdCli *VCDClient) vcdauthorize(user, pass, org string) error {
 }
 
 // NewVCDClient initializes VMware vCloud Director client with reasonable defaults.
-// It accepts functions of type VCDClientOption for adjusting defaults.
+// It accepts functions of type VCDClientOption for adjustedgegateway_test.going defaults.
```
dataclouder: mangled words

govcd/edgegateway_test.go
Didainius: Interested as well. The APIs behind it should be the proxied ones so I would expect they don't depend on versions of vCD API

types/v56/types.go
```diff
@@ -1134,6 +1135,7 @@ type VMGeneralParams struct {
 	Name               string `xml:"Name,omitempty"`               // Name of VM
 	Description        string `xml:"Description,omitempty"`        // VM description
 	NeedsCustomization bool   `xml:"NeedsCustomization,omitempty"` // True if this VM needs guest customization
+	RegenerateBiosUuid bool   `xml:"RegenerateBiosUuid,omitempty"` // True if BIOS UUID of the virtual machine should be regenerated so that it is unique, and not the same as the source virtual machine's BIOS UUID.
```
Didainius: Is this being impacted by version bump or  is it for separate PR?

govcd/edgegateway_test.go
vbauzysvmware: Well you assumptions wasn't right. It not always return first. Sometimes - not. Usually it works, but sometimes failed.

types/v56/types.go
vbauzysvmware: I found that in API diff documentatioin

types/v56/types.go
vbauzysvmware: No, third of values were added and a few removed.

vbauzysvmware: It was interesting - in general a few elements needed to be deleted cause won't be any info in them. Other were added.

types/v56/types.go
vbauzysvmware: Impact from version bump, nothing major - just introduced new field

govcd/api_vcd.go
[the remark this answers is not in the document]
vbauzysvmware: fixed

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * Added method VApp.AddNewVMWithStorageProfile that adds a VM with custom storage profile.
 * Added command `make static` to run staticcheck on all packages
 * Added `make static` to Travis regular checks
+* Increment vCD API version to 29.0
```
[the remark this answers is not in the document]

types/v56/types.go
lvirbalas: Please elaborate/give example. You mean API v29.0 defines a field, but returns nothing in it?

types/v56/types.go
vbauzysvmware: I run test with terraform - no complains on these changes.

vbauzysvmware: Where was not task in version 27.0 -> https://code.vmware.com/apis/72/vcloud-director/doc/doc/types/QueryResultVMRecordType.html

govcd/edgegateway_test.go
Didainius: Is there any way you know  how to replicate this as I've never hit it (at least with v27 )? I can look at it. I am exactly afraid of replacing all `id`s because this may compromise the goal of this test.

types/v56/types.go
vbauzysvmware: Then we move from one version to another - API won't return values even we would have property left. So no values to have them.

types/v56/types.go
Didainius: Interested in a bit more details. This is NSX-V API so v27 -> v29 shouldn't impact it. Any example how to replicate this problem?

govcd/edgegateway_test.go
dataclouder: I'd like more info on this change, as it alters the test behavior substantially. What exactly is failing that warrants such a modification?

types/v56/types.go
vbauzysvmware: VAppParentHREF and VAppParentName change to ContainerName and ContainerId

types/v56/types.go
Didainius: But is it being used anywhere yet (although I don't see in PR)? Just wondering if it is not changing the behavior as currently we are not regenerating it.

types/v56/types.go
vbauzysvmware: Run many times. At least on my env. Seems its just unordered list - so we can't impact behaviour

types/v56/types.go
vbauzysvmware: VmToolsVersion -> VmToolsStatus

vbauzysvmware: could be that HighestSupportedVersion - > HardwareVersion

types/v56/types.go
vbauzysvmware: We don't use it anywhere yet, just added to be up to date as we have such entity and new field arrived.

govcd/api_vcd.go
vbauzysvmware: fixed

govcd/edgegateway_test.go
vbauzysvmware: Test lack explicit value checking and expects that vCD always returns XML list elements to same order. So just check savedXML against newXML - to verify that everything is same. In general is the same, but sometimes vCD returns list in different order and XML don't match, though values of xml is the same.

Didainius: If XML rule order changes - then it means that real rule processing order is changed during global firewall setting changed, which is bad. I need to catch it.

Is it the `TestEdgeGateway_UpdateFwGeneralParams` test run? I will try to leave it running overnight on v29 and see if it fails.

vbauzysvmware: TestEdgeGateway_UpdateFwGeneralParams is enough.

Didainius: I have ran it 3812 times over night on v29 and it didn't fail. I suspect this is because your edge gateway firewall rule list is not empty for testing (looks like it has more than 1 rule of type `internal_high`). That isn't perfect, but I'd prefer to split this issue to a separate one (I can pick it on me).

types/v56/types.go
Didainius: Did it cause any issues? Because this is a proxied API I think it should depend on NSX version rather than vCD. For example the recent NSX version 6.4 states that this field is optional.

image

https://docs.vmware.com/en/VMware-NSX-Data-Center-for-vSphere/6.4/nsx_64_api.pdf (page 415)

Our provider codebase usually sends this parameter anyway, but having it without `omitempty` may send an empty value for someone who is not setting it and get an error so I'd say we could leave as it was unless the behavior of testing has changed with v29.

types/v56/types.go
Didainius: I guess we just need a list of it in changelog for those users who may rely on it

types/v56/types.go
Didainius: Again a list about removed fields in changelog may make sense for someone

govcd/edgegateway_test.go
Didainius: We have agreed to split this issue out of version change PR to minimize its footprint and I have raised a separate issue for investigation of this problem #264

vbauzysvmware: As we agree - revert back

types/v56/types.go
vbauzysvmware: Revert back. This is NSX-V type not vCD API.

types/v56/types.go

types/v56/types.go

lvirbalas: 1.) I suggest making changelog even clearer, and instead of only listing what has been removed and added, list what has been renamed too. I mean:
```
* `QueryResultVMRecordType` fields adjusted to accommodate API v27.0 to v29.0 upgrade as follows:
** Renamed fields: `VmToolsVersion` -> `VmToolsStatus`, ...
** Removed fields: ...
** Added fields: ...
```

This way users (and we) will have know-how **how** this type transformed from older to newer API version.

2.) Do I understand correctly that other than QueryResultVMRecordType, GatewayConfiguration, VMGeneralParams and NetworkConfiguration, other types' definitions haven't changed?

3.) If that's it, then I suggest adding those small pointer changes made in the other three types to changelog as well. The goal would be to have full write-up of the type changes caused by this API upgrade.

vbauzysvmware: Updated

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
 * Added method `VCDClient.SetToken`
 * Added method `VCDClient.GetAuthResponse`
 * Added script `scripts/get_token.sh`
+* Increment vCD API version used to 29.0
+    * Removed fields `VdcEnabled`, `VAppParentHREF`, `VAppParentName`, `HighestSupportedVersion`, `VmToolsVersion`, `TaskHREF`, `TaskStatusName`, `TaskDetails`, `TaskStatus` from `QueryResultVMRecordType`
```
lvirbalas:
```suggestion
    * Remove fields `VdcEnabled`, `VAppParentHREF`, `VAppParentName`, `HighestSupportedVersion`, `VmToolsVersion`, `TaskHREF`, `TaskStatusName`, `TaskDetails`, `TaskStatus` from `QueryResultVMRecordType`
```

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
...
+    * Added fields `ID, Type, ContainerName, ContainerID, OwnerName, Owner, NetworkHref, IpAddress, CatalogName, VmToolsStatus, GcStatus, AutoUndeployDate, AutoDeleteDate, AutoUndeployNotified, AutoDeleteNotified, Link, MetaData` to `QueryResultVMRecordType`, `DistributedInterface` to `NetworkConfiguration` and `RegenerateBiosUuid` to `VMGeneralParams`
```
lvirbalas:
```suggestion
    * Add fields `ID, Type, ContainerName, ContainerID, OwnerName, Owner, NetworkHref, IpAddress, CatalogName, VmToolsStatus, GcStatus, AutoUndeployDate, AutoDeleteDate, AutoUndeployNotified, AutoDeleteNotified, Link, MetaData` to `QueryResultVMRecordType`, `DistributedInterface` to `NetworkConfiguration` and `RegenerateBiosUuid` to `VMGeneralParams`
```

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
...
+    * Change `DistributedRoutingEnabled` to pointer in `GatewayConfiguration` and `DistributedInterface` in `NetworkConfiguration`.
```
lvirbalas:
```suggestion
    * Change to pointers `DistributedRoutingEnabled` in `GatewayConfiguration` and `DistributedInterface` in `NetworkConfiguration`
```

Issue #265, opened 2019-11-18
Add ability to get all virtual servers/pools from NSX


voyvodov:
It will be good to have method to get all virtual servers configured on the NSX load balancer instead of only looking by id/name.
This could help us to get all used IPs for an example or make checks if IP:Port combination is already in used (in case we want to create 2 VS on the same IP, but different ports).

This should be pretty easy to implement since all VS are already get and then filtered out by name/id.

voyvodov:
Already implemented

Didainius:
Just adding a message that this was for NSX-V. NSX-T Pools are in progress.

PR #266, opened 2019-11-18
Add ability to get all virtualserver/pools/service monitors

voyvodov:
Added method `EdgeGateway.GetLbVirtualServers`
Added method `EdgeGateway.GetLbServiceMonitors`
Added method `EdgeGateway.GetLbServerPools`

Methods `EdgeGateway.getLbVirtualServer`, `EdgeGateway.getLbServerPool` and `EdgeGateway.getLbServiceMonitor` are now using public methods as base to get all resources

## Description

Related issue: ([Issue 265](https://github.com/vmware/go-vcloud-director/issues/265))

- (Required) Add methods to return all configured Virtual Servers, Server Pools or Service Monitors on NSX Load Balancer

- (Required) 
Added method `EdgeGateway.GetLbVirtualServers`
Added method `EdgeGateway.GetLbServiceMonitors`
Added method `EdgeGateway.GetLbServerPools`

Private methods `getLb*` are now utilizing the public one to get all resources and then make the filtering as before.
Added new tests which checks that those public methods will return at least 1 resource (which was created before that).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,11 @@
     * Remove fields `VdcEnabled`, `VAppParentHREF`, `VAppParentName`, `HighestSupportedVersion`, `VmToolsVersion`, `TaskHREF`, `TaskStatusName`, `TaskDetails`, `TaskStatus` from `QueryResultVMRecordType`
     * Add fields `ID, Type, ContainerName, ContainerID, OwnerName, Owner, NetworkHref, IpAddress, CatalogName, VmToolsStatus, GcStatus, AutoUndeployDate, AutoDeleteDate, AutoUndeployNotified, AutoDeleteNotified, Link, MetaData` to `QueryResultVMRecordType`, `DistributedInterface` to `NetworkConfiguration` and `RegenerateBiosUuid` to `VMGeneralParams`
     * Change to pointers `DistributedRoutingEnabled` in `GatewayConfiguration` and `DistributedInterface` in `NetworkConfiguration`
+* Added method `EdgeGateway.GetLbVirtualServers` that gets all virtual servers configured on NSX load balancer. [#266](https://github.com/vmware/go-vcloud-director/pull/266)
+* Added method `EdgeGateway.GetLbServerPools` that gets all pools configured on NSX load balancer. [#266](https://github.com/vmware/go-vcloud-director/pull/266)
+* Added method `EdgeGateway.GetLbServiceMonitors` that gets all service monitors configured on NSX load balancer. [#266](https://github.com/vmware/go-vcloud-director/pull/266)
 
 BUGS FIXED:
-
 * Remove parentheses from filtering since they weren't treated correctly in some environment [#256]
   (https://github.com/vmware/go-vcloud-director/pull/256)
 * Take into account all subnets (SubnetParticipation) on edge gateway interface instead of the first
--- a/govcd/lbserverpool_test.go
+++ b/govcd/lbserverpool_test.go
@@ -110,6 +110,11 @@ func (vcd *TestVCD) Test_LBServerPool(check *C) {
 
 	check.Assert(createdLbPool.Algorithm, Equals, lbPoolConfig.Algorithm)
 
+	// GetLbServerPools should return at least one pool which is ours.
+	pools, err := edge.GetLbServerPools()
+	check.Assert(err, IsNil)
+	check.Assert(pools, Not(HasLen), 0)
+
 	// Test updating fields
 	// Update algorithm
 	lbPoolByID.Algorithm = "ip-hash"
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -14,8 +14,10 @@
     * Remove fields `VdcEnabled`, `VAppParentHREF`, `VAppParentName`, `HighestSupportedVersion`, `VmToolsVersion`, `TaskHREF`, `TaskStatusName`, `TaskDetails`, `TaskStatus` from `QueryResultVMRecordType`
     * Add fields `ID, Type, ContainerName, ContainerID, OwnerName, Owner, NetworkHref, IpAddress, CatalogName, VmToolsStatus, GcStatus, AutoUndeployDate, AutoDeleteDate, AutoUndeployNotified, AutoDeleteNotified, Link, MetaData` to `QueryResultVMRecordType`, `DistributedInterface` to `NetworkConfiguration` and `RegenerateBiosUuid` to `VMGeneralParams`
     * Change to pointers `DistributedRoutingEnabled` in `GatewayConfiguration` and `DistributedInterface` in `NetworkConfiguration`
+* Added method `EdgeGateway.GetLbVirtualServers` that gets all virtual servers configured on NSX load balancer.
```
Didainius: We somehow yet missed in this release, but could you link this PR as we have done in 2.4 release. Thanks
```suggestion
* Added method `EdgeGateway.GetLbVirtualServers` that gets all virtual servers configured on NSX load balancer. [#266](https://github.com/vmware/go-vcloud-director/pull/266)
```

govcd/lbserverpool_test.go
```diff
@@ -110,6 +110,11 @@ func (vcd *TestVCD) Test_LBServerPool(check *C) {
 
 	check.Assert(createdLbPool.Algorithm, Equals, lbPoolConfig.Algorithm)
 
+	// GetLbServerPools should return at least one pool which is our.
```
dataclouder:
```suggestion
	// GetLbServerPools should return at least one pool which is ours.
```

govcd/lbserverpool_test.go
```diff
@@ -110,6 +110,11 @@ func (vcd *TestVCD) Test_LBServerPool(check *C) {
...
+	pools, err := edge.GetLbServerPools()
+	check.Assert(err, IsNil)
+	check.Assert(pools, Not(HasLen), 0)
```
dataclouder: We should check that at least some of the server fields are as expected.

govcd/lbservicemonitor.go
```diff
@@ -92,6 +81,27 @@ func (egw *EdgeGateway) getLbServiceMonitor(lbMonitorConfig *types.LbMonitor) (*
 	return nil, ErrorEntityNotFound
 }
 
+// GetLbServiceMonitors return all service monitors without filtering
```
dataclouder: I don't see a test for this function

govcd/lbvirtualserver_test.go
```diff
@@ -100,6 +99,11 @@ func (vcd *TestVCD) Test_LBVirtualServer(check *C) {
 	check.Assert(lbVirtualServerById.ID, Equals, lbVirtualServerByName.ID)
 	check.Assert(lbVirtualServerById.Name, Equals, lbVirtualServerByName.Name)
 
+	// GetLbVirtualServers should return at least one vs which is our.
```
dataclouder:
```suggestion
	// GetLbVirtualServers should return at least one vs which is ours.
```

govcd/lbvirtualserver_test.go
```diff
@@ -100,6 +99,11 @@ func (vcd *TestVCD) Test_LBVirtualServer(check *C) {
...
+	servers, err := edge.GetLbVirtualServers()
+	check.Assert(err, IsNil)
+	check.Assert(servers, Not(HasLen), 0)
```
dataclouder: We should check at least some of the server fields, to make sure they are as expected

govcd/lbservicemonitor.go
voyvodov: Forgot to push it. Now should be there.

govcd/lbvirtualserver_test.go
voyvodov: Changed.

govcd/lbserverpool_test.go
voyvodov: Changed

govcd/lbvirtualserver_test.go
voyvodov: The problem is that if the testing environment is not clean (e.g. there is more than 1VS/Pool/etc.) we cannot be sure what are correct values. Generally we should not expect correct values, since we're not filtering by anything - we want to get everything without expecting something specific (for which there is get by id, get by name funcs)

govcd/lbserverpool_test.go
voyvodov: The problem is that if the testing environment is not clean (e.g. there is more than 1VS/Pool/etc.) we cannot be sure what are correct values. Generally we should not expect correct values, since we're not filtering by anything - we want to get everything without expecting something specific (for which there is get by id, get by name funcs)

PR #268, opened 2019-11-25
Support for routed network subinterface

dataclouder:
* Add support for routed network subinterface
* Fix network description management. The field was not retrieved due to its misplacement in the data structure.

### DIFF
--- a/govcd/orgvdcnetwork.go
+++ b/govcd/orgvdcnetwork.go
@@ -193,3 +193,60 @@ func (vdc *Vdc) CreateOrgVDCNetwork(networkConfig *types.OrgVDCNetwork) (Task, e
 	}
 	return Task{}, fmt.Errorf("network creation failed: no operational link found")
 }
+
+// FindEdgeGatewayNameByNetwork searches the VDC for a connection between an edge gateway and a given network.
+// On success, returns the name of the edge gateway
+//
+// CAVEAT: this function should only be called on routed networks.
+// There is a slim chance that this function returns a misleading result when BOTH the following
+// conditions are met:
+// (a) we query an isolated or direct network instead of a routed one
+// AND
+// (b) the network has the same name of an existing edge gateway
+// In normal operations (like retrieving the edge gateway name for a routed network data source) this will work
+// as expected.
+// If this function is used to detect the network type (i.e. if it is connected to an edge gateway it's routed)
+// it may fail its purpose if the above mentioned circumstance apply.
+func (vdc *Vdc) FindEdgeGatewayNameByNetwork(networkName string) (string, error) {
+
+	// Find the list of networks with the wanted name
+	result, err := vdc.client.QueryWithNotEncodedParams(nil, map[string]string{
+		"type":   "orgVdcNetwork",
+		"filter": fmt.Sprintf("name==%s;vdc==%s", url.QueryEscape(networkName), url.QueryEscape(vdc.Vdc.ID)),
+	})
+	if err != nil {
+		return "", fmt.Errorf("[findEdgeGatewayConnection] error returning the list of networks for VDC: %s", err)
+	}
+	netList := result.Results.OrgVdcNetworkRecord
+
+	// Find the list of edge gateways
+	edgeGatewayResult := new(types.QueryResultEdgeGatewayRecordsType)
+	for _, av := range vdc.Vdc.Link {
+		if av.Rel == "edgeGateways" && av.Type == "application/vnd.vmware.vcloud.query.records+xml" {
+
+			_, err := vdc.client.ExecuteRequest(av.HREF, http.MethodGet,
+				"", "error querying edge gateways: %s", nil, edgeGatewayResult)
+			if err != nil {
+				return "", err
+			}
+		}
+	}
+
+	var isAnEdgeGateway = func(gwName string) bool {
+		for _, gw := range edgeGatewayResult.EdgeGatewayRecord {
+			if gw.Name == gwName {
+				return true
+			}
+		}
+		return false
+	}
+
+	for _, net := range netList {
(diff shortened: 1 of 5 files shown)


types/v56/types.go
```diff
@@ -211,6 +211,7 @@ type NetworkConfiguration struct {
 	RetainNetInfoAcrossDeployments bool             `xml:"RetainNetInfoAcrossDeployments,omitempty"`
 	Features                       *NetworkFeatures `xml:"Features,omitempty"`
 	GuestVlanAllowed               *bool            `xml:"GuestVlanAllowed,omitempty"`
+	SubInterface                   *bool            `xml:"SubInterface,omitempty"`
 	DistributedInterface           *bool            `xml:"DistributedInterface,omitempty"`
```
lvirbalas: Is it possible to have both `SubInterface` and `DistributedInterface` set to true? If not, I'd add a comment for future reference.

dataclouder: Looking at how the GUI treats these values, they should be mutually exclusive.
See the picture in [TF Issue 342](https://github.com/terraform-providers/terraform-provider-vcd/issues/342).

lvirbalas: Yes, I'm aware of that. However, I don't know what happens if we send both true value to API.

dataclouder: A quick test with the API shows that we can't set both.
The API treats the interface types like this:
1. if both SubInterface and DistributedInterface are `nil`, it's "internal"
2. if either SubInterface or DistributedInterface are set, the corresponding interface is used.
3. If both variables are set, it's an error.

Notice that choosing the "internal" interface is the result of lack of other interfaces being set.

dataclouder:
> Don't we miss test for DistributedInterface ?

Thanks. Added

govcd/orgvdcnetwork.go
```diff
@@ -193,3 +193,60 @@ func (vdc *Vdc) CreateOrgVDCNetwork(networkConfig *types.OrgVDCNetwork) (Task, e
...
+// There is a slim chance that this function returns a misleading result when BOTH the following
+// conditions are met:
+// (a) we query an isolated or direct network instead of a routed one
+// AND
+// (b) the network has the same name of an existing edge gateway
+// In normal operations (like retrieving the edge gateway name for a routed network data source) this will work
+// as expected.
+// If this function is used to detect the network type (i.e. if it is connected to an edge gateway it's routed)
+// it may fail its purpose if the above mentioned circumstance apply.
+func (vdc *Vdc) FindEdgeGatewayNameByNetwork(networkName string) (string, error) {
```
vbauzysvmware: If comment are right, then function name should be named `FindEdgeGatewayNameByRoutedNetwork`

govcd/orgvdcnetwork.go
```diff
@@ -193,3 +193,60 @@ func (vdc *Vdc) CreateOrgVDCNetwork(networkConfig *types.OrgVDCNetwork) (Task, e
...
+// CAVEAT: this function should only be called on routed networks.
+// There is a slim chance that this function returns a misleading result when BOTH the following
+// conditions are met:
+// (a) we query an isolated or direct network instead of a routed one
+// AND
+// (b) the network has the same name of an existing edge gateway
+// In normal operations (like retrieving the edge gateway name for a routed network data source) this will work
+// as expected.
+// If this function is used to detect the network type (i.e. if it is connected to an edge gateway it's routed)
+// it may fail its purpose if the above mentioned circumstance apply.
```
lvirbalas: Could we somehow advance the function to **not** search the non-routed networks, which would fix all of the above caveat?

dataclouder: The data structure doesn't offer much in this respect. We don't know if a network is isolated or routed until we find (or not) a connection.

govcd/orgvdcnetwork.go
dataclouder: I think the name is long enough as it is.
It was @lvirbalas suggestion.

govcd/orgvdcnetwork.go
```diff
@@ -193,3 +193,60 @@ func (vdc *Vdc) CreateOrgVDCNetwork(networkConfig *types.OrgVDCNetwork) (Task, e
...
+		"type":   "orgVdcNetwork",
+		"filter": fmt.Sprintf("name==%s;vdc==%s", url.QueryEscape(networkName), url.QueryEscape(vdc.Vdc.ID)),
+	})
+	if err != nil {
+		return "", fmt.Errorf("[findEdgeGatewayConnection] error returning the list of networks for VDC: %s", err)
+	}
+	netList := result.Results.OrgVdcNetworkRecord
+
+	// Find the list of edge gateways
+	edgeGatewayResult := new(types.QueryResultEdgeGatewayRecordsType)
```
vbauzysvmware: why we checking in edge gtw list? we don't trust `ConnectedTo` ?

dataclouder: No. We don't.
Because for an isolated networks, `ConnectedTo` gives the name of the network itself.

govcd/orgvdcnetwork.go
vbauzysvmware: @lvirbalas This suggestions implies that I can use any network.

lvirbalas: As only routed has EG, I'm OK with leaving it as "short" as it is right now :)

govcd/orgvdcnetwork.go
vbauzysvmware: Very not nice from API. Interesting then, how in UI they different them ^^

dataclouder: FYI, I am using the same query that the UI is using.

PR #272, opened 2019-12-05
Add internal disk functions

vbauzysvmware:
Changes needed for https://github.com/terraform-providers/terraform-provider-vcd/pull/412
Implement functions which helps work with internal disk of VM.

* Add VM.AddInternalDisk
* Add VM.GetInternalDisk
* VM.DeleteInternalDisk
* VM.UpdateInternalDisks
* VM.UpdateInternalDisksAsync


govcd/vm.go
```diff
@@ -850,3 +851,186 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
...
+
+	return "", fmt.Errorf("created disk wasn't in list of returned VM internal disks")
+}
+
+func (vm *VM) validateInternalDiskInput(diskData *types.DiskSettings) error {
+	if diskData.AdapterType == "" {
+		return fmt.Errorf("disk settings missing required field: adapter type")
+	}
+
+	if diskData.BusNumber < 0 {
```
Didainius: I think this validation cannot be hit because the type is `int` and an empty value will be 0. Same for UnitNumber and SizeMb below. Go itself will "default" it to 0.

govcd/vm.go
```diff
@@ -850,3 +851,186 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
...
+		return "", fmt.Errorf("cannot add internal disks - VM HREF is unset")
+	}
+
+	err := vm.validateInternalDiskInput(diskData)
+	if err != nil {
+		return "", err
+	}
+
+	diskSettings := vm.VM.VmSpecSection.DiskSection.DiskSettings
+	if diskSettings == nil {
```
Didainius: Should this function have the "refresh" option or maybe even an implicit  refresh? I suspect that if one didn't refresh the VM before triggering this function - he would wipe all the disk configuration (if he has set the HREF only).

I wonder if it would be safe to assume that if `diskSettings == nil` the VM is actually not refreshed. Wondering if the `diskSettings` structure has any difference when a VM has no disks (but is refreshed) vs being `nil` (not refreshed).

govcd/vm.go
```diff
@@ -850,3 +851,186 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
...
+	if diskData.StorageProfile != nil {
+		return fmt.Errorf("disk settings missing required field: storage profile")
+	}
+
+	return nil
+}
+
+// GetInternalDisk returns a valid *types.DiskSettings if it exists.
+// If it doesn't, returns nil and ErrorEntityNotFound or other err.
+func (vm *VM) GetInternalDisk(diskId string) (*types.DiskSettings, error) {
```
Didainius: Maybe the below name to follow our standard? 

```suggestion
func (vm *VM) GetInternalDiskById(diskId string) (*types.DiskSettings, error) {
```

govcd/vm.go
```diff
@@ -850,3 +851,186 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
...
+	if vm.VM.HREF == "" {
```
Didainius: HREF is validated, but it is not used anywhere. Also what about the "refresh" parameter?

govcd/vm.go
```diff
@@ -850,3 +851,186 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
...
+			return diskSetting, nil
+		}
+	}
+
+	return nil, ErrorEntityNotFound
+}
+
+// DeleteInternalDisk delete disk using provided disk ID.
+// Runs synchronously, VM is ready for another operation after this function returns.
+func (vm *VM) DeleteInternalDisk(diskId string) error {
```
Didainius:
```suggestion
func (vm *VM) DeleteInternalDiskById(diskId string) error {
```

govcd/vm.go
```diff
@@ -850,3 +851,186 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
...
+		return nil, fmt.Errorf("cannot get internal disk - VM HREF is unset")
+	}
+
+	if diskId == "" {
+		return nil, fmt.Errorf("cannot get internal disk - provided diskId is empty")
+	}
+
+	if vm.VM.VmSpecSection.DiskSection == nil || vm.VM.VmSpecSection.DiskSection.DiskSettings == nil ||
+		len(vm.VM.VmSpecSection.DiskSection.DiskSettings) == 0 {
+		return nil, fmt.Errorf("cannot get internal disk - VM don't have internal disks")
```
Didainius:
```suggestion
		return nil, fmt.Errorf("cannot get internal disk - VM doesn't have internal disks")
```

govcd/vm.go
```diff
@@ -850,3 +851,186 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
...
+
+	diskPlacement := -1
+	for i, diskSetting := range vm.VM.VmSpecSection.DiskSection.DiskSettings {
+		if diskSetting.DiskId == diskId {
+			diskPlacement = i
+		}
+	}
+
+	if diskPlacement == -1 {
+		return fmt.Errorf("cannot find VM internal disk with Id: %s", diskId)
```
Didainius: What about returning `ErrorEntityNotFound` here?

types/v56/types.go
```diff
@@ -1360,11 +1360,106 @@ type VM struct {
...
+// MemoryResourceMb from VM/VmSpecSection struct
+type MemoryResourceMb struct {
+	Configured  int64  `xml:"Configured`             // The amount of resource configured on the virtual machine.
+	Reservation *int64 `xml:"Reservation,omitempty"` // The amount of reservation of this resource on the underlying virtualization infrastructure.
+	Limit       *int64 `xml:"Limit,omitempty"`       // The limit for how much of this resource can be consumed on the underlying virtualization infrastructure. This is only valid when the resource allocation is not unlimited.
+	SharesLevel string `xml:"SharesLevel,omitempty"` //	Pre-determined relative priorities according to which the non-reserved portion of this resource is made available to the virtualized workload.
+	Shares      *int   `xml:"Shares,omitempty"`      // Custom priority for the resource. This is a read-only, unless the share level is CUSTOM.
+}
+
+type HardwareVersion struct {
```
Didainius: Missing doc

govcd/vm.go
```diff
@@ -850,3 +851,186 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
...
+
+	if diskData.SizeMb < int64(0) {
+		return fmt.Errorf("disk settings size MB has to be 0 or higher")
+	}
+
+	if diskData.Iops != nil && *diskData.Iops < int64(0) {
+		return fmt.Errorf("disk settings iops has to be 0 or higher")
+	}
+
+	if diskData.ThinProvisioned != nil {
```
Didainius: Shouldn't it be the opposite? 
```suggestion
	if diskData.ThinProvisioned == nil {
```

Same applies to below `StorageProfile` field


```shell
vm_test.go:1065:
    check.Assert(err, IsNil)
... value *errors.errorString = &errors.errorString{s:"disk settings missing required field: thin provisioned"} ("disk settings missing required field: thin provisioned")
```

types/v56/types.go
```diff
@@ -1360,11 +1360,106 @@ type VM struct {
...
+// VM represents a virtual machine only with Disk setting update part
+type VMDiskChange struct {
+	// Attributes
+	XMLName xml.Name `xml:"Vm"`
+	Ovf     string   `xml:"xmlns:ovf,attr,omitempty"`
+	Xsi     string   `xml:"xmlns:xsi,attr,omitempty"`
+	Xmlns   string   `xml:"xmlns,attr,omitempty"`
+
+	HREF string `xml:"href,attr,omitempty"`
+	Type string `xml:"type,attr,omitempty"`
```
lvirbalas: Is there some documentation about `Type` which we could copy/paste here as a comment? I'd like to see here available Type values.

govcd/vm.go
vbauzysvmware: This is govcd, not only terraform can call it. So user could provide for e.g. `-1`

govcd/vm.go
vbauzysvmware: Getting me to thinking, not to have such option at all. User won't know details and get's to risk to messup no refreshing when needed and deleting disks which he wasn't expecting

govcd/vm.go
lvirbalas:
```suggestion
		return nil, fmt.Errorf("cannot get internal disk - VM doesn't have internal disks")
```

govcd/vm.go
vbauzysvmware: ty. Changed

types/v56/types.go

govcd/vm.go
vbauzysvmware: updated

govcd/vm.go
vbauzysvmware: fixed

govcd/vm.go
vbauzysvmware: changed

govcd/vm.go
vbauzysvmware: changed

govcd/vm.go
vbauzysvmware: I will need to check, forget todo it, though thought about it. A lof things here to lookup after

types/v56/types.go
```diff
@@ -1360,11 +1360,106 @@ type VM struct {
...
+	Name string `xml:"name,attr"`
```
lvirbalas: Thinking more about it, I'd like to see all fields of this structure commented (like of structure `VmSpecSection` below). It's a good practice and avoids uncertainties like what does this `Name` mean. Apparently, it's VM's name not disk's, but that is not clear without a comment.

govcd/vm.go
Didainius: True, but my main concern is probably the values being `int` instead of `*int` as if user didn't set anything - we will get 0. Up to you if it looks to be risky.

vbauzysvmware: it wouldn't be correct to make *int from vCD API side

types/v56/types.go
```diff
@@ -1360,11 +1360,106 @@ type VM struct {
...
 	// Environment OVF_Environment `xml:"Environment,omitempty"
 
+	VmSpecSection *VmSpecSection `xml:"VmSpecSection,omitempty"`
+
 	VMCapabilities *VMCapabilities `xml:"VmCapabilities,omitempty"` // Allows you to specify certain capabilities of this virtual machine.
 	StorageProfile *Reference      `xml:"StorageProfile,omitempty"` // A reference to a storage profile to be used for this object. The specified storage profile must exist in the organization vDC that contains the object. If not specified, the default storage profile for the vDC is used.
 	ProductSection *ProductSection `xml:"ProductSection,omitempty"`
 }
 
+// VM represents a virtual machine only with Disk setting update part
```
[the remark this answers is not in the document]
vbauzysvmware: updated

govcd/vm.go
vbauzysvmware: so when where isn't any internal disk, then we get `<DiskSection/>` - code or more less correct. I am just thinking to look for possibility enforce refresh.

govcd/vm.go
vbauzysvmware: updated

types/v56/types.go

types/v56/types.go

govcd/vm.go
vbauzysvmware: I added refresh handling. I made it implicit due big risk leave for user and he would forget to pass correct value.

govcd/vm.go
vbauzysvmware: I added refresh handling. I made it implicit due big risk leave for user and he would forget to pass correct value.

govcd/vm.go
Didainius: Agreed.

govcd/vm.go
```diff
@@ -850,3 +851,186 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
...
 	return vm.GetGuestCustomizationSection()
 }
+
+// AddInternalDisk creates disk type *types.DiskSettings to the VM.
+// Returns new disk ID and error.
+// Runs synchronously, VM is ready for another operation after this function returns.
+func (vm *VM) AddInternalDisk(diskData *types.DiskSettings) (string, error) {
+	err := vm.Refresh()
+	if err != nil {
+		return "", fmt.Errorf("error refresing vm: %s", err)
```
lvirbalas: I noticed you removed VM name from this and two more identical error messages:
https://github.com/vmware/go-vcloud-director/commit/f0aee2fde6614f85633967023aaf8c3e3714158e

How does the message look like now?

BTW, there's a typo in a all three.

govcd/vm.go
```diff
@@ -850,3 +851,186 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
...
+			return diskSetting, nil
+		}
+	}
+
+	return nil, ErrorEntityNotFound
+}
+
+// DeleteInternalDiskById delete disk using provided disk ID.
+// Runs synchronously, VM is ready for another operation after this function returns.
+func (vm *VM) DeleteInternalDiskById(diskId string) error {
```
dataclouder: Why is the `ById` in the name? We don't have `DeleteInternalDiskByName`, and BTW, the changelog has the name without the `ById` suffix

govcd/vm.go
```diff
@@ -850,3 +851,186 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
...
+		if diskSetting.DiskId == diskId {
+			diskPlacement = i
+		}
+	}
+
+	if diskPlacement == -1 {
+		return ErrorEntityNotFound
+	}
+
+	// remove disk in slice
```
dataclouder:
```suggestion
	// remove disk from slice
```

govcd/vm.go
```diff
@@ -850,3 +851,186 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
...
+// If it doesn't, returns nil and ErrorEntityNotFound or other err.
+func (vm *VM) GetInternalDiskById(diskId string, refresh bool) (*types.DiskSettings, error) {
+	if refresh {
+		err := vm.Refresh()
+		if err != nil {
+			return nil, fmt.Errorf("error refreshing VM: %s", err)
+		}
+	}
+
+	if diskId == "" {
```
dataclouder: We should check for valid ID before refreshing.

govcd/vm_test.go
```diff
@@ -985,23 +987,92 @@ func (vcd *TestVCD) Test_GetVirtualHardwareSection(check *C) {
...
+		OverrideVmDefault: true,
+		Iops:              &iops,
+	}
+	diskId, err := vm.AddInternalDisk(diskSettings)
+
+	check.Assert(err, IsNil)
+	check.Assert(diskId, NotNil)
+
+	//verify
+	err = vm.Refresh()
```
dataclouder: This refresh is not needed, given that `GetInternalDiskById` is called with refresh=true

govcd/vm_test.go
```diff
@@ -985,23 +987,92 @@ func (vcd *TestVCD) Test_GetVirtualHardwareSection(check *C) {
...
+	check.Assert(disk.BusNumber, Equals, diskSettings.BusNumber)
+	check.Assert(disk.AdapterType, Equals, diskSettings.AdapterType)
+}
 
-	// Loop over the Items to ensure
-	for _, item := range section.Item {
-		check.Assert(item.ResourceType, NotNil)
-		check.Assert(item.ResourceSubType, NotNil)
+// Test delete internal disk For VM
+func (vcd *TestVCD) Test_DeleteInternalDisk(check *C) {
```
dataclouder: There is a lot in common between this function and `Test_AddInternalDisk`. Can you combine them?
If not, could you add a creation function that can be called by both?

types/v56/types.go
```diff
@@ -1360,11 +1360,107 @@ type VM struct {
...
+	VmSpecSection *VmSpecSection `xml:"VmSpecSection,omitempty"`
+
 	VMCapabilities *VMCapabilities `xml:"VmCapabilities,omitempty"` // Allows you to specify certain capabilities of this virtual machine.
 	StorageProfile *Reference      `xml:"StorageProfile,omitempty"` // A reference to a storage profile to be used for this object. The specified storage profile must exist in the organization vDC that contains the object. If not specified, the default storage profile for the vDC is used.
 	ProductSection *ProductSection `xml:"ProductSection,omitempty"`
 }
 
+// VMDiskChange represents a virtual machine only with Disk setting update part
+type VMDiskChange struct {
+	// Attributes
```
dataclouder: redundant comment

types/v56/types.go
```diff
@@ -1360,11 +1360,107 @@ type VM struct {
...
+	Ovf     string   `xml:"xmlns:ovf,attr,omitempty"`
+	Xsi     string   `xml:"xmlns:xsi,attr,omitempty"`
+	Xmlns   string   `xml:"xmlns,attr,omitempty"`
+
+	HREF string `xml:"href,attr,omitempty"` // The URI of the VM entity.
+	Type string `xml:"type,attr,omitempty"` // The MIME type of the entity - application/vnd.vmware.vcloud.vm+xml
+	Name string `xml:"name,attr"`           // VM name
+	ID   string `xml:"id,attr,omitempty"`   // VM ID. The entity identifier, expressed in URN format. The value of this attribute uniquely identifies the entity, persists for the life of the entity, and is never reused.
+
+	VmSpecSection *VmSpecSection `xml:"VmSpecSection,omitempty"` // Container for the specification of this virtual machine. This is an alternate to using ovf:VirtualHardwareSection + ovf:OperatingSystemSection
```
dataclouder:
```suggestion
	VmSpecSection *VmSpecSection `xml:"VmSpecSection,omitempty"` // Container for the specification of this virtual machine. This is an alternative to using ovf:VirtualHardwareSection + ovf:OperatingSystemSection
```

types/v56/types.go
```diff
@@ -1360,11 +1360,107 @@ type VM struct {
...
+}
+
+// VmSpecSection from VM struct
+type VmSpecSection struct {
+	Modified          *bool             `xml:"Modified,attr,omitempty"`
+	Info              string            `xml:"ovf:Info"`
+	OsType            string            `xml:"OsType,omitempty"`            // The type of the OS. This parameter may be omitted when using the VmSpec to update the contents of an existing VM.
+	NumCpus           *int              `xml:"NumCpus,omitempty"`           // Number of CPUs. This parameter may be omitted when using the VmSpec to update the contents of an existing VM.
+	NumCoresPerSocket *int              `xml:"NumCoresPerSocket,omitempty"` // Number of cores among which to distribute CPUs in this virtual machine.. This parameter may be omitted when using the VmSpec to update the contents of an existing VM.
```
dataclouder: Double dot in the text

types/v56/types.go
```diff
@@ -1360,11 +1360,107 @@ type VM struct {
...
+	CpuResourceMhz    *CpuResourceMhz   `xml:"CpuResourceMhz,omitempty"`    // CPU compute resources. This parameter may be omitted when using the VmSpec to update the contents of an existing VM.
+	MemoryResourceMb  *MemoryResourceMb `xml:"MemoryResourceMb"`            // Memory compute resources. This parameter may be omitted when using the VmSpec to update the contents of an existing VM.
+	MediaSection      *MediaSection     `xml:"MediaSection,omitempty"`      // The media devices of this VM.
+	DiskSection       *DiskSection      `xml:"DiskSection,omitempty"`       // virtual disks of this VM.
+	HardwareVersion   *HardwareVersion  `xml:"HardwareVersion"`             // vSphere name of Virtual Hardware Version of this VM. Example: vmx-13 This parameter may be omitted when using the VmSpec to update the contents of an existing VM.
```
dataclouder:
```suggestion
	HardwareVersion   *HardwareVersion  `xml:"HardwareVersion"`             // vSphere name of Virtual Hardware Version of this VM. Example: vmx-13 - This parameter may be omitted when using the VmSpec to update the contents of an existing VM.
```

types/v56/types.go
```diff
@@ -1360,11 +1360,107 @@ type VM struct {
...
+}
+
+// DiskSection from VM/VmSpecSection struct
+type DiskSection struct {
+	DiskSettings []*DiskSettings `xml:"DiskSettings"`
+}
+
+// DiskSettings from VM/VmSpecSection/DiskSection struct
+type DiskSettings struct {
+	DiskId              string     `xml:"DiskId,omitempty"`              // Specifies a unique identifier for this disk in the scope of the corresponding VM. This element is optional when creating a VM, but if it is provided it should be unique. This element is mandatory when updating an existing disk.
```
dataclouder: Is it unique among all VMs or unique within the VM? Please specify in the comment

types/v56/types.go
```diff
@@ -1360,11 +1360,107 @@ type VM struct {
...
+	SizeMb              int64      `xml:"SizeMb"`                        // The size of the disk in MB.
+	UnitNumber          int        `xml:"UnitNumber"`                    // The device number on the SCSI or IDE controller of the disk.
+	BusNumber           int        `xml:"BusNumber"`                     //	The number of the SCSI or IDE controller itself.
+	AdapterType         string     `xml:"AdapterType"`                   // The type of disk controller, e.g. IDE vs SCSI and if SCSI bus-logic vs LSI logic.
```
dataclouder: Why using `vs` here? Can we use `or` instead?

types/v56/types.go
```diff
@@ -1360,11 +1360,107 @@ type VM struct {
...
+	AdapterType string     `xml:"AdapterType,omitempty"` // The type of controller, e.g. IDE vs SCSI and if SCSI bus-logic vs LSI logic
+}
+
+// CpuResourceMhz from VM/VmSpecSection struct
+type CpuResourceMhz struct {
+	Configured  int64  `xml:"Configured`             // The amount of resource configured on the virtual machine.
+	Reservation *int64 `xml:"Reservation,omitempty"` // The amount of reservation of this resource on the underlying virtualization infrastructure.
+	Limit       *int64 `xml:"Limit,omitempty"`       // The limit for how much of this resource can be consumed on the underlying virtualization infrastructure. This is only valid when the resource allocation is not unlimited.
+	SharesLevel string `xml:"SharesLevel,omitempty"` //	Pre-determined relative priorities according to which the non-reserved portion of this resource is made available to the virtualized workload.
+	Shares      *int   `xml:"Shares,omitempty"`      // Custom priority for the resource. This is a read-only, unless the share level is CUSTOM.
```
dataclouder:
```suggestion
	Shares      *int   `xml:"Shares,omitempty"`      // Custom priority for the resource. This field is read-only, unless the shares level is CUSTOM.
```

types/v56/types.go
```diff
@@ -1360,11 +1360,107 @@ type VM struct {
...
+	BusNumber   int        `xml:"BusNumber"`             //	The bus number of the media device controller.
+	AdapterType string     `xml:"AdapterType,omitempty"` // The type of controller, e.g. IDE vs SCSI and if SCSI bus-logic vs LSI logic
+}
+
+// CpuResourceMhz from VM/VmSpecSection struct
+type CpuResourceMhz struct {
+	Configured  int64  `xml:"Configured`             // The amount of resource configured on the virtual machine.
+	Reservation *int64 `xml:"Reservation,omitempty"` // The amount of reservation of this resource on the underlying virtualization infrastructure.
+	Limit       *int64 `xml:"Limit,omitempty"`       // The limit for how much of this resource can be consumed on the underlying virtualization infrastructure. This is only valid when the resource allocation is not unlimited.
+	SharesLevel string `xml:"SharesLevel,omitempty"` //	Pre-determined relative priorities according to which the non-reserved portion of this resource is made available to the virtualized workload.
```
dataclouder: I don't understand this comment. Could it be rephrased differently?

types/v56/types.go
```diff
@@ -1360,11 +1360,107 @@ type VM struct {
...
+	MediaImage  *Reference `xml:"MediaImage,omitempty"`  // The media image that is mounted onto the device. This property can be 'null' which represents that no media is mounted on the device.
+	UnitNumber  int        `xml:"UnitNumber"`            // Specified the type of media that is mounted onto the device.
+	BusNumber   int        `xml:"BusNumber"`             //	The bus number of the media device controller.
+	AdapterType string     `xml:"AdapterType,omitempty"` // The type of controller, e.g. IDE vs SCSI and if SCSI bus-logic vs LSI logic
+}
+
+// CpuResourceMhz from VM/VmSpecSection struct
+type CpuResourceMhz struct {
+	Configured  int64  `xml:"Configured`             // The amount of resource configured on the virtual machine.
+	Reservation *int64 `xml:"Reservation,omitempty"` // The amount of reservation of this resource on the underlying virtualization infrastructure.
```
dataclouder: 1. Do you mean `the reserved amount`? (if such, can this field be named `Reserved`?
2. Why Configured is a integer and Reservation is a pointer to an integer?

types/v56/types.go
```diff
@@ -1360,11 +1360,107 @@ type VM struct {
...
+	Limit       *int64 `xml:"Limit,omitempty"`       // The limit for how much of this resource can be consumed on the underlying virtualization infrastructure. This is only valid when the resource allocation is not unlimited.
+	SharesLevel string `xml:"SharesLevel,omitempty"` //	Pre-determined relative priorities according to which the non-reserved portion of this resource is made available to the virtualized workload.
+	Shares      *int   `xml:"Shares,omitempty"`      // Custom priority for the resource. This is a read-only, unless the share level is CUSTOM.
+}
+
+// MemoryResourceMb from VM/VmSpecSection struct
+type MemoryResourceMb struct {
+	Configured  int64  `xml:"Configured`             // The amount of resource configured on the virtual machine.
+	Reservation *int64 `xml:"Reservation,omitempty"` // The amount of reservation of this resource on the underlying virtualization infrastructure.
```
dataclouder: Same remarks as for `CpuResourceMhz`

govcd/vm.go
vbauzysvmware: updated mistype. We can't have `vm.VM.Name,` as we can get nil errors  if VM is empty

govcd/vm.go
```diff
@@ -850,3 +851,186 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
 
 	return vm.GetGuestCustomizationSection()
 }
+
+// AddInternalDisk creates disk type *types.DiskSettings to the VM.
+// Returns new disk ID and error.
+// Runs synchronously, VM is ready for another operation after this function returns.
+func (vm *VM) AddInternalDisk(diskData *types.DiskSettings) (string, error) {
+	err := vm.Refresh()
```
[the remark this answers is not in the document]
vbauzysvmware: We agreed within Dainius comment that's it best scenario to have enforced(implicit) refresh as we handle very important thing - internal disk. As any disk change involves passing all set of internal disk we need to work with latest info.

govcd/vm.go
vbauzysvmware: This was request by Dainius. I am also fine with it. Leave more abstract name for feature in case we may need it. FYI - update changelog. Thank you for noticing

govcd/vm.go
vbauzysvmware: updated

govcd/vm.go
vbauzysvmware: Moved.

govcd/vm_test.go
vbauzysvmware: removed

types/v56/types.go
vbauzysvmware: removed

types/v56/types.go
vbauzysvmware: updated

types/v56/types.go
vbauzysvmware: updated

types/v56/types.go
vbauzysvmware: updated

types/v56/types.go
vbauzysvmware: Just copy from documentation. We can change if we don't like

types/v56/types.go
vbauzysvmware: updated

types/v56/types.go
vbauzysvmware: If you know how - please advice. It's from documentation ^^

types/v56/types.go
vbauzysvmware: 1. Comment is from documentation. 
2. As it is optional https://vdc-download.vmware.com/vmwb-repository/dcr-public/037ccaee-649a-417e-b365-1331034fb28d/1f0fd9eb-0238-4af6-89b5-7e6636f29c65/doc/doc/types/ComputeResourceType.html

types/v56/types.go
vbauzysvmware: Answer as for `CpuResourceMhz`

govcd/vm.go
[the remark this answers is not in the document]
dataclouder: Let me rephrase: I see that we don't check whether the disk we want to add exists already.
Given that, why do we need to refresh? If the disk addition operation is idempotent, all is fine and we don't need a refresh. If it isn't, then after refresh we need to check whether the disk exists. Otherwise, we are just wasting time.

@Didainius, @lvirbalas , your take?

govcd/vm.go
Didainius: The var `diskId` confused me. If it is ID then I thought it is ById

govcd/vm.go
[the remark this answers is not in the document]
Didainius: Just a few lines below there is a check if `diskSettings == nil` and also appending to empty diskSettings slice. If someone initialized the VM with just HREF - he could actually send an empty disk payload and wipe all the disks. At least that was how I understood.

govcd/vm.go
dataclouder: Probably the confusion comes from a different approach we're taking here. 
In other entities, we do `entity.Delete`, while here we do `parent.DeleteEntityById`.
In both case, I think the `ById` suffix is not needed, as it suggests that there are other `DeleteBySomeField` which we don't have.

govcd/vm.go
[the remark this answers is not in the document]
vbauzysvmware: Normally when you add Disk, you don't provide diskId, as vCD generates it. So you want to check if disk exist by adapter type,bus number, unit number and all other details? I think we will get to may strange situations and I recommend to leave that to API. `Refresh` is needed as we send all existing disk with new one together as new request to `reconfigureVM`

govcd/vm.go
vbauzysvmware: @Didainius is it fine I revert back?

Didainius: Go ahead - it is not a critical thing although to my mind this name is closer to reality :)

govcd/vm.go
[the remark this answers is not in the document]
dataclouder: What would happen if the combination of type, bus number, and unit number is already taken? Would the current disk go to overwrite the existing one?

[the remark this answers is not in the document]
vbauzysvmware: If it can't create disk due bus type taken or something else, API sends back error with message. You can't overwrite disk.

govcd/vm.go
```diff
@@ -850,3 +851,186 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
...
+			diskSetting.BusNumber == diskData.BusNumber &&
+			diskSetting.UnitNumber == diskData.UnitNumber {
+			return diskSetting.DiskId, nil
+		}
+	}
+
+	return "", fmt.Errorf("created disk wasn't in list of returned VM internal disks")
+}
+
+func (vm *VM) validateInternalDiskInput(diskData *types.DiskSettings) error {
```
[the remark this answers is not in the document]
vbauzysvmware: Added

govcd/vm.go
vbauzysvmware: done

govcd/vm_test.go
vbauzysvmware: done

types/v56/types.go
vbauzysvmware: don't have info for that

govcd/api_vcd_versions.go
```diff
@@ -178,3 +178,12 @@ func (cli *Client) validateAPIVersion() error {
 
 	return nil
 }
+
+// GetDefaultAPIVersionOrLaterThan32 returns default version or 32 if it is connected to version vCD 9.7 or later
```
dataclouder: I don't understand the reason for this function, with an API version hard coded in the name.
* What's the problem we need to solve?
* Can this function become a more general one, so that it doesn't become obsolete when 32 is not so important?

govcd/vm.go
```diff
@@ -850,3 +851,187 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
...
+	}
+
+	if diskData.StorageProfile == nil {
+		return fmt.Errorf("[VM %s Id %s]disk settings missing required field: storage profile", vmName, vmId)
+	}
+
+	return nil
+}
+
+// GetInternalDiskById returns a valid *types.DiskSettings if it exists.
```
dataclouder:
```suggestion
// GetInternalDiskById returns a *types.DiskSettings if one exists.
```

govcd/vm.go
```diff
@@ -850,3 +851,187 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
...
+	if err != nil {
+		return fmt.Errorf("error deleting VM %s internal disk %s: %s", vm.VM.Name, diskId, err)
+	}
+
+	return nil
+}
+
+// UpdateInternalDisks applies disks configuration for the VM.
+// types.VmSpecSection requires consist of all disk entities which exist and not updated,
+// as also new ones or changed ones. Returns new disk ID and error.
```
dataclouder: Please rephrase. I don't understand this comment.

govcd/vm.go
```diff
@@ -850,3 +851,187 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
...
+	err = vm.Refresh()
+	if err != nil {
+		return nil, fmt.Errorf("error refreshing VM %s: %s", vm.VM.Name, err)
+	}
+	return vm.VM.VmSpecSection, nil
+}
+
+// UpdateInternalDisksAsync applies disks configuration and return task or err
+// types.VmSpecSection requires consist of all disk entities which exist and not updated,
+// as also new ones or changed ones.
```
dataclouder: Same as above

govcd/vm.go
```diff
@@ -850,3 +851,187 @@ func (vm *VM) SetGuestCustomizationSection(guestCustomizationSection *types.Gues
...
+func (vm *VM) UpdateInternalDisksAsync(disksSettingToUpdate *types.VmSpecSection) (Task, error) {
+	if vm.VM.HREF == "" {
+		return Task{}, fmt.Errorf("cannot update disks, VM HREF is unset")
+	}
+
+	disksSettingToUpdate.Info = "Virtual hardware requirements (simplified)"
```
dataclouder: 1. Where does this hard coded value come from?
2. And where will it go?
3. Does it have to be exactly like that or can it be customized?

govcd/vm_test.go
```diff
@@ -967,15 +967,34 @@ func (vcd *TestVCD) Test_VMSetGetGuestCustomizationSection(check *C) {
...
+	fmt.Printf("Running: %s\n", check.TestName())
+
+	vm, storageProfile, diskSettings, diskId, err := vcd.createInternalDisk(check, 1)
+	check.Assert(err, IsNil)
+
+	//verify
+	disk, err := vm.GetInternalDiskById(diskId, true)
+	check.Assert(err, IsNil)
 
+	check.Assert(disk.StorageProfile.HREF, Equals, storageProfile.HREF)
```
dataclouder: Please add a check for ID correspondence

govcd/vm_test.go
```diff
@@ -967,15 +967,34 @@ func (vcd *TestVCD) Test_VMSetGetGuestCustomizationSection(check *C) {
...
+	check.Assert(disk.AdapterType, Equals, diskSettings.AdapterType)
+	check.Assert(*disk.ThinProvisioned, Equals, *diskSettings.ThinProvisioned)
+	check.Assert(*disk.Iops, Equals, *diskSettings.Iops)
+	check.Assert(disk.SizeMb, Equals, diskSettings.SizeMb)
+	check.Assert(disk.UnitNumber, Equals, diskSettings.UnitNumber)
+	check.Assert(disk.BusNumber, Equals, diskSettings.BusNumber)
+	check.Assert(disk.AdapterType, Equals, diskSettings.AdapterType)
+}
+
+func (vcd *TestVCD) createInternalDisk(check *C, busNumber int) (*VM, types.Reference, *types.DiskSettings, string, error) {
```
dataclouder: Please add a comment with the function purpose and the description of the returned values

types/v56/types.go
dataclouder: It doesn't sound right to me.

types/v56/types.go
dataclouder: To suggest a change I'd have to understand it first

govcd/api_vcd_versions.go
vbauzysvmware: Reason:
If vCD 9.7 or later we need to use 32.0 version in some requests to work with internal disk in VM.
I could copy paste these 4 lines, but created function to be nice. My aim to create specific function for dedicated version override. Not generic one.

types/v56/types.go
vbauzysvmware: According info I get, ID are unique within VM.

govcd/vm.go
vbauzysvmware: updated

govcd/vm.go
vbauzysvmware: updated

govcd/vm.go
vbauzysvmware: updated

govcd/vm.go
vbauzysvmware: 1. It's from: 
```
 <VmSpecSection>
            <ovf:Info>Virtual hardware requirements (simplified)</ovf:Info>
            <OsType>centos7_64Guest</OsType>
            <NumCpus>4</NumCpus>
            <NumCoresPerSocket>1</NumCoresPerSocket>
            <CpuResourceMhz>
```
2. Goes with POST (updating VM internal disk)
3. When it's not there, API complains

govcd/vm_test.go
vbauzysvmware: added

govcd/api_vcd_versions.go
Didainius: I guess @dataclouder  meant something like `func (cli *Client) GetDefaultAPIVersionOrLaterThan(32) string`

vbauzysvmware: Well, if you both want that I could change function. My intention was to have explicit traceability by version in which function we override. Also hides need to know details how to define exact API version. As for example it's not `32` it's `32.0`

govcd/vm.go
vbauzysvmware: I rechecked with a few version and removed it. Seems previously missing this field show errors due other things.

govcd/vm_test.go
vbauzysvmware: added

CHANGELOG.md
```diff
@@ -1,4 +1,14 @@
-## 2.5.0 (Unreleased)
+## 2.6.0 (Unreleased)
+
+* Moved `VCDClient.supportedVersions` to `VCDClient.Client.supportedVersions` [#274](https://github.com/vmware/go-vcloud-director/pull/274)    
+* Added methods `VM.AddInternalDisk`, `VM.GetInternalDiskById`, `VM.DeleteInternalDisk`, `VM.UpdateInternalDisks` and `VM.UpdateInternalDisksAsync`
+
+BUGS FIXED:
```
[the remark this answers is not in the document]
vbauzysvmware: update according request

govcd/api_vcd_versions.go
lvirbalas: We really need a generalized function with the purpose of having a **single code path** for dealing with version manipulations. Possible name suggestion:
```
func (cli *Client) GetSpecificApiVersionOnCondition(vcdApiVersionCondition, wantedApiVersion)
```
Call example:
```
client.GetSpecificApiVersionOnCondition(">= 32.0", "32.0")
```
The reason I propose to expose the ">=" in addition to the number, because it's actually clearer of what the function does with this approach than to interpret a very long function name.

For example, current name "GetDefaultAPIVersionOrLaterThan32" is very hard to grasp and is also not precise, because in reality it is "GetDefaultApiVersionOr32IfVersionIsLaterThan32". And at this point this conundrum reminds me of "42" and the true meaning of life :)

BTW, "API" should be "Api" ;)

Of course, in this case, the comment needs to state that the default version is returned if condition is not met.

govcd/vm_test.go
```diff
@@ -991,11 +991,14 @@ func (vcd *TestVCD) Test_AddInternalDisk(check *C) {
 	//cleanup
 	err = vm.DeleteInternalDisk(diskId)
 	check.Assert(err, IsNil)
+
+	// disable fast provisioning if needed
+	updateVdcFastProvisioning(vcd, check, previousProvisioningValue)
```
Didainius: Could there be a cleanup task (using AddToCleanupList) so that if the test crashes we still have the ability to restore the environment to what it was.

govcd/vm_test.go
```diff
@@ -986,23 +1015,80 @@ func (vcd *TestVCD) Test_GetVirtualHardwareSection(check *C) {
...
+	check.Assert(diskId, NotNil)
+	return vm, storageProfile, diskSettings, diskId, previousVdcFastProvisioningValue, err
+}
 
-	// Loop over the Items to ensure
-	for _, item := range section.Item {
-		check.Assert(item.ResourceType, NotNil)
-		check.Assert(item.ResourceSubType, NotNil)
+// updateVdcFastProvisioning Enables or Disables fast provisioning if needed
+func updateVdcFastProvisioning(vcd *TestVCD, check *C, enable string) string {
```
dataclouder: We are returning only two possible values.
Seems a task that is suited for a `bool` instead of a string.

vbauzysvmware: Yes, it's true - only two possible values. I made string that it would be more readable. updateVdcFastProvisioning(.., `enable`) vs updateVdcFastProvisioning(.., true)

govcd/vm_test.go
vbauzysvmware: added

govcd/api_vcd_versions.go
vbauzysvmware: done

PR #273, opened 2019-12-06
vCD 9.0.0.3 - test error fixes

Didainius:
This PR has specific fixes for vCD 9.0.0.3 (older 9.0 versions are buggy and would still fail) related testing:
* `Test_CreateExternalNetwork` - IPRanges are sorted differently everytime.
* `Test_CreateDeleteEdgeGatewayAdvanced` - 
  * FIPS mode is not supported by vCD 9.0 therefore the field must not be sent
  * Edge Gateway interface ID is not reported after API call `GatewayInterface[0].Network.ID`
* `Test_PowerOnAndForceCustomization` - VM reports `GuestCustStatusComplete` instead of `GuestCustStatusPending` after `PowerOnAndForceCustomization()` in vCD 9.0
* `TestEdgeGateway_GetVnics` (renamed to `TestEdgeGateway_getVdcNetworks` because there is a different endpoint to get network mappings correctly in vCD 9.0 and newer versions)
* `Test_NsxvDnatRule` and `Test_NsxvSnatRule`

**Note**. Legacy NAT related functions (which are not using NSXV-API) do not work in vCD 9.0.0: `Test_AddDNATRule`, `Test_NATMapping`, `Test_NATMapping`, `Test_UpdateNATRule`. They fail with error similar to:
```
Edge Gateway: API Error: 400: [ b19b360a-1b43-4eb1-b065-6a83eb661846 ] A NAT rule has an invalid vnic value 1. The value should be a vnic index number
```

This is a vCD bug which was fixed in 9.0.0.2 (https://docs.vmware.com/en/VMware-vCloud-Director-for-Service-Providers/9.0.0.2/rn/rel_notes_vcloud_director_9-0-0-2.html) 

Testing status with vCD 9.0.0.9840881 (includes 9.0.03 patches):
```
OK: 177 passed, 1 skipped
PASS
ok      github.com/vmware/go-vcloud-director/v2/govcd   2792.658s
```
vCD 9.5.0.11038232
```
OK: 178 passed
PASS
ok      github.com/vmware/go-vcloud-director/v2/govcd   3122.972s
```

10.0.0.14636284
```
OK: 176 passed, 2 skipped
PASS
ok      github.com/vmware/go-vcloud-director/v2/govcd   4549.654s
```


govcd/edgegateway.go
```diff
@@ -1378,31 +1368,31 @@ func getVnicIndexByNetworkNameAndType(networkName, networkType string, vnics *ty
...
-		if vnic.Type == types.EdgeGatewayVnicTypeTrunk && len(vnic.SubInterfaces.SubInterface) > 0 {
-			for _, subInterface := range vnic.SubInterfaces.SubInterface {
-				if subInterface.Index != nil && *subInterface.Index == vNicIndex {
-					foundCount++
-					networkName = subInterface.LogicalSwitchName
-					networkType = types.EdgeGatewayVnicTypeSubinterface
-				}
-			}
-		}
+		// // Search inside "subinterface tree"
```
lvirbalas: Please either remove the commented block or add a comment explaining why we need it.

Didainius: Oh, sorry - leftover. Removed.

PR #274, opened 2019-12-06
Allow to pass and override default hard coded api version

vbauzysvmware:
This change is driven from VM internal disk PR(https://github.com/vmware/go-vcloud-director/pull/272) and investigation. Also we saw need that in near feature in other cases, as we face more such cases when we need specific version due limitation of using lowest possible API version.

* Moved  field supportedVersions from vCDclient to client
* Updated all references with cange of supportedVersions
* Added request handling methods `WithApiVersion` which allows to pass API version which will be written in request header(override default hard coded one)

We need override only per request base API version. Changing connection API version would create compatibility issues and breakage.


govcd/api.go
```diff
@@ -125,8 +127,13 @@ func ContainsNotFound(err error) bool {
 	return err != nil && strings.Contains(err.Error(), ErrorEntityNotFound.Error())
 }
 
-// NewRequestWitNotEncodedParams allows passing complex values params that shouldn't be encoded like for queries. e.g. /query?filter=name=foo
+// NewRequestWitNotEncodedParamsWithApi allows passing complex values params that shouldn't be encoded like for queries. e.g. /query?filter=name=foo
```
lvirbalas: `NewRequestWitNotEncodedParamsWithApi` doesn't seem to match real function name.

govcd/api.go
```diff
@@ -125,8 +127,13 @@ func ContainsNotFound(err error) bool {
...
 func (cli *Client) NewRequestWitNotEncodedParams(params map[string]string, notEncodedParams map[string]string, method string, reqUrl url.URL, body io.Reader) *http.Request {
+	return cli.NewRequestWitNotEncodedParamsWithApiVersion(params, notEncodedParams, method, reqUrl, body, cli.APIVersion)
+}
+
+// NewRequestWitNotEncodedParamsWithApiVersion allows passing complex values params that shouldn't be encoded like for queries. e.g. /query?filter=name=foo
+func (cli *Client) NewRequestWitNotEncodedParamsWithApiVersion(params map[string]string, notEncodedParams map[string]string, method string, reqUrl url.URL, body io.Reader, apiVersion string) *http.Request {
```
lvirbalas: I would like to see additional comment about how to use `apiVersion` field.

govcd/api.go
vbauzysvmware: updated

govcd/api.go
vbauzysvmware: added

govcd/api.go
```diff
@@ -188,6 +196,12 @@ func (cli *Client) NewRequest(params map[string]string, method string, reqUrl ur
 	return cli.NewRequestWitNotEncodedParams(params, nil, method, reqUrl, body)
 }
 
+// NewRequest creates a new HTTP request and applies necessary auth headers if set.
```
dataclouder:
```suggestion
// NewRequestWithApiVersion creates a new HTTP request and applies necessary auth headers if set.
```

govcd/api.go
```diff
@@ -127,6 +129,12 @@ func ContainsNotFound(err error) bool {
 
 // NewRequestWitNotEncodedParams allows passing complex values params that shouldn't be encoded like for queries. e.g. /query?filter=name=foo
 func (cli *Client) NewRequestWitNotEncodedParams(params map[string]string, notEncodedParams map[string]string, method string, reqUrl url.URL, body io.Reader) *http.Request {
+	return cli.NewRequestWitNotEncodedParamsWithApiVersion(params, notEncodedParams, method, reqUrl, body, cli.APIVersion)
```
dataclouder: I'd like to see a few tests for the new functions, to make sure that:
1. They are really using the intended API version.
2. Subsequent requests using the regular `Execute*` methods are not affected.

vbauzysvmware: Could you share your idea how to test that?

govcd/api.go
vbauzysvmware: udpated

govcd/api.go
dataclouder: Here's a possible way:
1. Add fields to a type that was enhanced in API versions later than the current one.
2. Run a GET that retrieves the type with the enhanced fields
3. Make sure the fields are non-empty
4. Run the same query without the new API version
5. Make sure that the enhanced fields are empty

Didainius: Two more ideas:
* https://golang.org/pkg/net/http/httptest/
or 
* IIRC when you send a request on a specific version, vCD responds with API version (in headers). You could probably check that.

vbauzysvmware: first option could be valid with on specific vCD version or I have to find valid for all versions, which is lottery
I'll check possibility with response headers

vbauzysvmware: Test added

govcd/api_test.go
```diff
@@ -95,3 +100,23 @@ func TestTags(t *testing.T) {
 		showTags()
 	}
 }
+
+// Test_NewRequestWitNotEncodedParamsWithApiVersion verifies that api version override works
+func (vcd *TestVCD) Test_NewRequestWitNotEncodedParamsWithApiVersion(check *C) {
+	fmt.Printf("Running: %s\n", check.TestName())
```
Didainius: Double check the outputs as they are missing new lines. Heres' how it looks.

```
START: api_test.go:105: TestVCD.Test_NewRequestWitNotEncodedParamsWithApiVersion
Running: TestVCD.Test_NewRequestWitNotEncodedParamsWithApiVersion
Test: TestVCD.Test_NewRequestWitNotEncodedParamsWithApiVersion run with api Version: 33.0
PASS: api_test.go:105: TestVCD.Test_NewRequestWitNotEncodedParamsWithApiVersion 0.191s
```

Didainius: Ok. This is fine as it is as planned.

PR #278, opened 2020-01-12
List functions

dataclouder:
Add a few functions that are needed for list manipulation in other libraries or applications. (`GetVappTemplateByHref` and `GetEdgeGatewayReferenceList`)

Remove unused fields in `VCDClient` and `Vdc`.
Export function `GetEdgeGatewayByHref`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,8 @@ module github.com/vmware/go-vcloud-director/v2
 
 require (
 	github.com/hashicorp/go-version v1.2.0
-	github.com/kr/pretty v0.1.0 // indirect
+	github.com/kr/pretty v0.2.0 // indirect
+	// github.com/kr/pretty v0.1.0 // indirect
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127
 	gopkg.in/yaml.v2 v2.2.2
 )
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,6 @@ module github.com/vmware/go-vcloud-director/v2
 require (
 	github.com/hashicorp/go-version v1.2.0
 	github.com/kr/pretty v0.2.0 // indirect
-	// github.com/kr/pretty v0.1.0 // indirect
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127
 	gopkg.in/yaml.v2 v2.2.2
 )
--- a/govcd/catalog_test.go
+++ b/govcd/catalog_test.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
 	"github.com/vmware/go-vcloud-director/v2/util"
 	. "gopkg.in/check.v1"
 )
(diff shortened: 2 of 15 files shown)


go.mod
```diff
@@ -2,7 +2,8 @@ module github.com/vmware/go-vcloud-director/v2
 
 require (
 	github.com/hashicorp/go-version v1.2.0
-	github.com/kr/pretty v0.1.0 // indirect
+	github.com/kr/pretty v0.2.0 // indirect
+	// github.com/kr/pretty v0.1.0 // indirect
```
Didainius: No need for commented line.

govcd/vdc.go
```diff
@@ -353,10 +352,10 @@ func (vdc *Vdc) FindEdgeGateway(edgegateway string) (EdgeGateway, error) {
 
 }
 
-// getEdgeGatewayByHref retrieves an edge gateway from VDC
+// GetEdgeGatewayByHref retrieves an edge gateway from VDC
 // by querying directly its HREF.
 // The name passed as parameter is only used for error reporting
-func (vdc *Vdc) getEdgeGatewayByHref(name, href string) (*EdgeGateway, error) {
+func (vdc *Vdc) GetEdgeGatewayByHref(name, href string) (*EdgeGateway, error) {
```
Didainius: IMHO name is extraneous here. There may be cases when one doesn't have name and it would be some random string

govcd/vdc.go
```diff
@@ -837,3 +836,41 @@ func (vdc *Vdc) buildNsxvNetworkServiceEndpointURL(optionalSuffix string) (strin
...
+func (vdc *Vdc) GetEdgeGatewayReferenceList(refresh bool) ([]*types.Reference, error) {
+
+	if refresh {
+		err := vdc.Refresh()
+		if err != nil {
+			return nil, fmt.Errorf("error refreshing vdc: %s", err)
+		}
+	}
+	for _, av := range vdc.Vdc.Link {
+		if av.Rel == "edgeGateways" && av.Type == "application/vnd.vmware.vcloud.query.records+xml" {
```
Didainius: Could potentially be a constant as we have it for some other types. Not a necessity.

govcd/vdc.go
```diff
@@ -837,3 +836,41 @@ func (vdc *Vdc) buildNsxvNetworkServiceEndpointURL(optionalSuffix string) (strin
...
+					Type: egw.Type,
+					ID:   "",
+					HREF: egw.HREF,
+				}
+				refList[i] = &ref
+			}
+			return refList, nil
+		}
+	}
+	return nil, fmt.Errorf("no edge gateway query link found in VDC %s", vdc.Vdc.Name)
```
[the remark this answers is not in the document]
dataclouder: Not in this case. We are returning a list, not a single entity.

go.mod
dataclouder: removed

govcd/vdc.go
dataclouder: Thanks. Removed

govcd/vdc.go
dataclouder: In fact, we had a constant, although it contained a different text and was unused.
Now I put it to good use with the right contents.

govcd/vdc.go
```diff
@@ -837,3 +836,41 @@ func (vdc *Vdc) buildNsxvNetworkServiceEndpointURL(optionalSuffix string) (strin
 
 	return hostname, nil
 }
+
+// GetEdgeGatewayReferenceList retrieves a list of edge gateway references.
+// Unlike other entities under Vdc, edge gateway are not listed in the parent entities
```
lvirbalas: Nit: "Vdc" should be "VDC" in comments (sentences for humans), while "Vdc" should be used in function names.

govcd/catalog_test.go
```diff
@@ -530,3 +531,41 @@ func (vcd *TestVCD) Test_CatalogGetItem(check *C) {
...
+	catalogItem, err := catalog.GetCatalogItemByName(vcd.config.VCD.Catalog.CatalogItem, false)
+	check.Assert(err, IsNil)
+	check.Assert(catalogItem, NotNil)
+	check.Assert(catalogItem.CatalogItem.Entity, NotNil)
+
+	vappTemplate, err := catalog.GetVappTemplateByHref(catalogItem.CatalogItem.Entity.HREF)
+	check.Assert(err, IsNil)
+	check.Assert(vappTemplate, NotNil)
+	check.Assert(vappTemplate.VAppTemplate.ID, Not(Equals), catalogItem.CatalogItem.ID)
+	check.Assert(vappTemplate.VAppTemplate.Type, Equals, types.MimeVAppTemplate)
```
vbauzysvmware: Is it possible to add assert by item name, description? I think vapp template and catalog item name should match. At least in that way we confirm re received what we expected.

govcd/edgegateway_test.go
```diff
@@ -771,3 +771,30 @@ func replaceFirstMatch(text, regex, replacement string) string {
...
+	check.Assert(err, IsNil)
+	check.Assert(edge, NotNil)
+	edgeList, err := vcd.vdc.GetEdgeGatewayReferenceList(false)
+	check.Assert(err, IsNil)
+	check.Assert(edgeList, NotNil)
+	check.Assert(len(edgeList) > 0, Equals, true)
+	foundName := false
+	foundHref := false
+	for _, ref := range edgeList {
+		if ref.Name == vcd.config.VCD.EdgeGateway {
```
vbauzysvmware:
```suggestion
		if ref.Name == edge.EdgeGateway.name {
```

govcd/vdc.go
```diff
@@ -837,3 +836,41 @@ func (vdc *Vdc) buildNsxvNetworkServiceEndpointURL(optionalSuffix string) (strin
...
+// and therefore we need to get the list using a query.
+func (vdc *Vdc) GetEdgeGatewayReferenceList(refresh bool) ([]*types.Reference, error) {
```
vbauzysvmware: what is difference between `getEdgeGatewayRecordsType` and this function?

dataclouder: The difference is only the return type.
I think I can achieve the same result by making `getEdgeGatewayRecordsType` public

vbauzysvmware: Or making more like wrapper function which converts to your needed type, but it's up to you.

govcd/vdc.go
dataclouder: "Fixed". On Vaidotas advice, I remove this function completely, and made a similar private function public.

govcd/catalog_test.go
dataclouder: done

govcd/edgegateway_test.go
dataclouder: done

Issue #279, opened 2020-01-15
External network doesn't able to be createad


slavaavr:
Hi there. Recently I came across that current version of library is not able to create external network. The thing is the line associated with network configuration -  https://github.com/vmware/go-vcloud-director/blob/master/govcd/system.go#L623 is not fully correct. Regarding to the vCloud Director API documentation (p.360), internal fields of network configuration must have xml prefix - `vcloud`, but they haven't. Thus, generated xml bodies are not valid.

vbauzysvmware:
Hi @slavaavr,

Could you provide us HCL example you are runing. Also terraform runing/crash log and to better understand issue we need more logs from go-vcloud-director which should be enabled.
How to enable is mentioned in documentation: https://www.terraform.io/docs/providers/vcd/index.html - parameters logging, logging_file.

slavaavr:
@vbauzysvmware is it ok if I provide you golang config I am using with? (I don't use terraform)

lvirbalas:
Hi @slavaavr , curious, are you creating some Go tool based on go-vcloud-director?

CC: @dataclouder @Didainius

vbauzysvmware:
Yes, @slavaavr, such example is enough. Thank you

dataclouder:
Please have a look at the sample code linked below, and see if you can create an external network using this snippet as guideline:

https://github.com/vmware/go-vcloud-director/blob/master/govcd/externalnetwork_test.go#L77

slavaavr:
@lvirbalas, yep, this is for work.

slavaavr:
@vbauzysvmware, glad to hear it

slavaavr:
@dataclouder thank you very much.  I did confused a bit, but now I've realised where the problem was: it's because of the lack 'Xmlns: vcdtypes.XMLNamespaceVCloud' field in the 'NetworkConfiguration' struct. 
__
So, thanks everyone!

vbauzysvmware:
@slavaavr Glad to here :)

I wonder, is #280 still needed?

slavaavr:
@vbauzysvmware, I closed the PR :)

vbauzysvmware:
> @vbauzysvmware, I closed the PR :)

@slavaavr  thank you :)

PR #280, opened 2020-01-15
Fix bug connected with vcloud prefix propagation

slavaavr:
https://github.com/vmware/go-vcloud-director/issues/279

To solve the problem described in the issue, 'vcloud' prefixes were added to the network configuration internal fields. And one more thing, the field - 'VimPortGroupRef' was deleted from 'externalNetworkCreate' struct since this field doesn't work at all.


lvirbalas:
@slavaavr Which vCD versions have you tried these changes with?

slavaavr:
@lvirbalas I tried only vCD with version 9.7.0.3. I added some details to the issue https://github.com/vmware/go-vcloud-director/issues/279

PR #284, opened 2020-01-20
Add handling name spaces improvement

vbauzysvmware:
To avoid such issues in future https://github.com/vmware/go-vcloud-director/issues/279 we need to stop expect users will provide needed XML name spaces.

Change handles name spaces inside functions as we did in other places - this makes our code more consistent and user friendly for users. Also errors are very nasty when name space isn't provided, so let's take care of them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Moved `VCDClient.supportedVersions` to `VCDClient.Client.supportedVersions` [#274](https://github.com/vmware/go-vcloud-director/pull/274)    
 * Added methods `VM.AddInternalDisk`, `VM.GetInternalDiskById`, `VM.DeleteInternalDisk`, `VM.UpdateInternalDisks` and `VM.UpdateInternalDisksAsync`
 * Added methods `vdc.GetEdgeGatewayReferenceList` and `catalog.GetVappTemplateByHref`
+* Improved functions to not expect XML namespaces provided in argument structure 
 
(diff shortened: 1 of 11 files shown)


vbauzysvmware:
> A note in changelog :)
> 
> ```
> OK: 184 passed
> PASS
> ok      github.com/vmware/go-vcloud-director/v2/govcd   3528.119s
> ```
> 
> Do you know if terraform provider tests pass fine with this PR?

I am not sure is worth it to mentioned anything in changelog regarding this..

I am not expecting any impact in terraform provider from this. If even use namespace it will be overwritten.

vbauzysvmware:
> Does this PR cover all occurrences in the code base of setting the namespace? To rephrase, are there any calls left which would still need to set it explicitly? Thanks!

What I found - I changed.

lvirbalas:
Let's start adding PR URLs/links to the changelog, like we do in Terraform. BTW, please add them to the other two current version changelog items that are missing them. Right now only one has it.

vbauzysvmware:
> Let's start adding PR URLs/links to the changelog, like we do in Terraform. BTW, please add them to the other two current version changelog items that are missing them. Right now only one has it.

added

lvirbalas:
The new links are missing parenthesis and because of that they are not parsed:

Screenshot 2020-01-21 at 14 57 03

vbauzysvmware:
> The new links are missing parenthesis and because of that they are not parsed:
> 
> Screenshot 2020-01-21 at 14 57 03

updated

PR #285, opened 2020-01-22
VDC supports Flex

vbauzysvmware:
Ref:https://github.com/terraform-providers/terraform-provider-vcd/issues/423, https://github.com/terraform-providers/terraform-provider-vcd/pull/443
Enhances existing implementation to allow create VDC allocation model Flex:
* refactored and introduced structure to work with different versions of API using same public function.
* added Flex and new fields
* added unit tests


govcd/adminorg.go
```diff
@@ -90,6 +90,7 @@ func (adminOrg *AdminOrg) GetVdcByName(vdcname string) (Vdc, error) {
 // CreateVdc creates a VDC with the given params under the given organization.
 // Returns an AdminVdc.
 // API Documentation: https://code.vmware.com/apis/220/vcloud#/doc/doc/operations/POST-VdcConfiguration.html
+// Deprecated in favor adminOrg.CreateOrgVdcAsync
```
Didainius:
```suggestion
// Deprecated in favor of adminOrg.CreateOrgVdcAsync
```

govcd/adminorg.go
```diff
@@ -117,6 +118,7 @@ func (adminOrg *AdminOrg) CreateVdc(vdcConfiguration *types.VdcConfiguration) (T
 }
 
 // Creates the vdc and waits for the asynchronous task to complete.
+// Deprecated in favor adminOrg.CreateOrgVdc
```
Didainius:
```suggestion
// Deprecated in favor of adminOrg.CreateOrgVdc
```

govcd/adminvdc.go
```diff
@@ -52,33 +54,273 @@ func (adminVdc *AdminVdc) Refresh() error {
...
-		types.MimeAdminVDC, "error updating VDC: %s", adminVdc.AdminVdc)
 }
 
 // Update function updates an Admin VDC from current VDC struct contents.
 // Any differences that may be legally applied will be updated.
 // Returns an empty AdminVdc struct and error if the call to vCD fails.
 // API Documentation: https://vdc-repo.vmware.com/vmwb-repository/dcr-public/7a028e78-bd37-4a6a-8298-9c26c7eeb9aa/09142237-dd46-4dee-8326-e07212fb63a8/doc/doc/operations/PUT-Vdc.html
 func (adminVdc *AdminVdc) Update() (AdminVdc, error) {
-	task, err := adminVdc.UpdateAsync()
+	apiVersion, err := adminVdc.client.maxSupportedVersion()
```
Didainius: Wondering if we could reuse `UpdateAsync` here instead of function lookup, etc.

govcd/adminvdc.go
```diff
@@ -52,33 +54,273 @@ func (adminVdc *AdminVdc) Refresh() error {
...
 	if err != nil {
-		return AdminVdc{}, err
+		return nil, err
+	}
+
+	return adminVdc, nil
+}
+
+// updateVdcAsyncV97 updates a VDC with the given params. Supports Flex type allocation.
+// Needs vCD 9.7 to work. Returns an Task.
```
Didainius: Looking at the code below I think this is the correct version notation:
```suggestion
// Needs vCD 9.7+ to work. Returns an Task.
```

govcd/api_vcd_versions.go
```diff
@@ -26,6 +26,24 @@ type SupportedVersions struct {
...
+var vcdVersionToApiVersion = map[string]string{
+	"29.0": "9.0",
+	"30.0": "9.1",
+	"31.0": "9.5",
+	"32.0": "9.7",
+	"33.0": "10.0",
+}
+
+// map allows you get max supported API version from vCD version
+var vpiVersionToVcdVersion = map[string]string{
```
Didainius: Typo
```suggestion
var apiVersionToVcdVersion = map[string]string{
```

govcd/adminvdc_test.go
```diff
@@ -0,0 +1,237 @@
...
+		vdc, _ := adminOrg.GetVDCByName(vdcConfiguration.Name, false)
+		if vdc != nil {
+			err = vdc.DeleteWait(true, true)
+			check.Assert(err, IsNil)
+		}
+
+		task, err := adminOrg.CreateOrgVdcAsync(vdcConfiguration)
+		check.Assert(err, Not(IsNil))
+		check.Assert(task, Equals, Task{})
+		if vcd.client.Client.APIVCDMaxVersionIs("> 32.0") {
```
Didainius: Could you put a comment with reason for this `if`? It is hard to grasp not knowing the background.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 * Added methods `VM.AddInternalDisk`, `VM.GetInternalDiskById`, `VM.DeleteInternalDisk`, `VM.UpdateInternalDisks` and `VM.UpdateInternalDisksAsync`
 * Change `int` and `bool` fields from types.VAppTemplateLeaseSettings and VAppLeaseSettings into pointers
 * Added method `catalog.GetVappTemplateByHref`, and expose methods `vdc.GetEdgeGatewayByHref` and `vdc.GetEdgeGatewayRecordsType`
+* Added methods `adminOrg.CreateOrgVdc`, `adminOrg.CreateOrgVdcAsync` and improved existing to support Flex VDC model
```
Didainius: I would say we need a sentence that there is a dynamic method and version change invocation behind the scenes based on vCD version

govcd/adminvdc.go
```diff
@@ -52,33 +54,273 @@ func (adminVdc *AdminVdc) Refresh() error {
...
+
+	// Return the task
+	return adminVdc.client.ExecuteTaskRequestWithApiVersion(adminVdc.AdminVdc.HREF, http.MethodPut,
+		types.MimeAdminVDC, "error updating VDC: %s", adminVdc.AdminVdc,
+		adminVdc.client.GetSpecificApiVersionOnCondition(">= 32.0", "32.0"))
+}
+
+// updateVdcV97 updates a VDC with the given params
+// and waits for the asynchronous task to complete. Supports Flex type allocation.
+// Needs vCD 9.7 to work. Returns an AdminVdc.
```
Didainius:
```suggestion
// Needs vCD 9.7+ to work. Returns an AdminVdc.
```

govcd/adminvdc.go
```diff
@@ -52,33 +54,273 @@ func (adminVdc *AdminVdc) Refresh() error {
...
+// Returns an Task.
+func createVdcAsync(adminOrg *AdminOrg, vdcConfiguration *types.VdcConfiguration) (Task, error) {
+	util.Logger.Printf("[TRACE] createVdcAsync called %#v", *vdcConfiguration)
+	return adminOrg.CreateVdc(vdcConfiguration)
+}
 
-	return *adminVdc, nil
+// createVdcAsyncV97 creates a VDC with the given params under the given organization
+// and waits for the asynchronous task to complete. Supports Flex type allocation.
+// Needs vCD 9.7 to work. Returns an Vdc.
```
Didainius:
```suggestion
// Needs vCD 9.7+ to work. Returns an Vdc.
```

govcd/adminorg.go
vbauzysvmware: fixed

govcd/adminorg.go
vbauzysvmware: fixed

govcd/adminvdc.go
vbauzysvmware: fixed

govcd/api_vcd_versions.go
vbauzysvmware: fixed

govcd/adminvdc_test.go
vbauzysvmware: added

CHANGELOG.md
vbauzysvmware: added

govcd/adminvdc.go
vbauzysvmware: fixed

govcd/adminvdc.go
vbauzysvmware: fixed

vbauzysvmware:
> Had a look over. There is still a conflict for merge and some inline comments.

Conflicts solved

govcd/adminvdc.go
vbauzysvmware: Internally it's using Async, but here we resolving which function to use.

govcd/adminvdc.go
```diff
@@ -37,8 +39,8 @@ func (adminVdc *AdminVdc) Refresh() error {
 	// elements in slices.
 	unmarshalledAdminVdc := &types.AdminVdc{}
 
-	_, err := adminVdc.client.ExecuteRequest(adminVdc.AdminVdc.HREF, http.MethodGet,
-		"", "error refreshing VDC: %s", nil, unmarshalledAdminVdc)
+	_, err := adminVdc.client.ExecuteRequestWithApiVersion(adminVdc.AdminVdc.HREF, http.MethodGet,
```
lvirbalas: Please add a comment why we're capping the version to be no higher than 32 on this refresh operation.

govcd/api_vcd_versions.go
```diff
@@ -26,6 +26,27 @@ type SupportedVersions struct {
...
+var vcdVersionToApiVersion = map[string]string{
+	"29.0": "9.0",
+	"30.0": "9.1",
+	"31.0": "9.5",
+	"32.0": "9.7",
+	"33.0": "10.0",
+}
+
+// map allows you get max supported API version from vCD version
+var apiVersionToVcdVersion = map[string]string{
```
lvirbalas: The names of these two maps are inverted. The names should be switched places, because now the names do opposite of what they say. E.g.:

```
var apiVersionToVcdVersion = map[string]string{
	"9.0":  "29.0",
	"9.1":  "30.0",
	"9.5":  "31.0",
	"9.7":  "32.0",
	"10.0": "33.0",
}

func main() {
	fmt.Println(apiVersionToVcdVersion["29.0"])
}
```

This will not print anything, although one would expect "9.0" to be returned.

CC: @dataclouder @Didainius

govcd/adminvdc.go
vbauzysvmware: added

govcd/api_vcd_versions.go
vbauzysvmware: corrected naming

govcd/adminvdc.go
lvirbalas: Sorry, but I don't understand the logic behind the code or comment. Comment says "new version API additionally returns ...", but the code **limits** the version to 32, if the version is higher than 32! Please rephrase/elaborate.

vbauzysvmware: @lvirbalas it doesn't. Function returns default api version 29 or 32 if conditions is met.
Function name is according your suggestions form previous PR.

govcd/adminvdc.go
```diff
@@ -52,33 +55,273 @@ func (adminVdc *AdminVdc) Refresh() error {
...
 // API Documentation: https://vdc-repo.vmware.com/vmwb-repository/dcr-public/7a028e78-bd37-4a6a-8298-9c26c7eeb9aa/09142237-dd46-4dee-8326-e07212fb63a8/doc/doc/operations/PUT-Vdc.html
 func (adminVdc *AdminVdc) Update() (AdminVdc, error) {
-	task, err := adminVdc.UpdateAsync()
+	apiVersion, err := adminVdc.client.maxSupportedVersion()
 	if err != nil {
 		return AdminVdc{}, err
 	}
 
-	err = task.WaitTaskCompletion()
+	realFunction, ok := vdcProducerByVersion["vdc"+apiVersionToVcdVersion[apiVersion]]
```
dataclouder: Rather than `realFunction`, a better name here would be `producer`

govcd/adminvdc.go
```diff
@@ -52,33 +55,273 @@ func (adminVdc *AdminVdc) Refresh() error {
 // Returns an error if the call to vCD fails.
 // API Documentation: https://vdc-repo.vmware.com/vmwb-repository/dcr-public/7a028e78-bd37-4a6a-8298-9c26c7eeb9aa/09142237-dd46-4dee-8326-e07212fb63a8/doc/doc/operations/PUT-Vdc.html
 func (adminVdc *AdminVdc) UpdateAsync() (Task, error) {
+	apiVersion, err := adminVdc.client.maxSupportedVersion()
+	if err != nil {
+		return Task{}, err
+	}
+	realFunction, ok := vdcProducerByVersion["vdc"+apiVersionToVcdVersion[apiVersion]]
```
dataclouder: Rather than `realFunction`, a better name here would be `producer`

govcd/adminvdc.go
```diff
@@ -52,33 +55,273 @@ func (adminVdc *AdminVdc) Refresh() error {
...
+	"vdc9.0":  vdcCrudV90,
+	"vdc9.1":  vdcCrudV90,
+	"vdc9.5":  vdcCrudV90,
+	"vdc9.7":  vdcCrudV97,
+	"vdc10.0": vdcCrudV97,
+}
+
+// CreateOrgVdc creates a VDC with the given params under the given organization
+// and waits for the asynchronous task to complete.
+// Returns an AdminVdc.
```
dataclouder:
```suggestion
// Returns an AdminVdc pointer and an error.
```

govcd/adminvdc.go
```diff
@@ -52,33 +55,273 @@ func (adminVdc *AdminVdc) Refresh() error {
...
+	if realFunction.CreateVdc == nil {
+		return nil, fmt.Errorf("function CreateVdc is not defined for %s", "vdc"+apiVersion)
+	}
+
+	util.Logger.Printf("[DEBUG] CreateOrgVdc call function for version %s", realFunction.SupportedVersion)
+	return realFunction.CreateVdc(adminOrg, vdcConfiguration)
+}
+
+// CreateOrgVdcAsync creates a VDC with the given params under the given organization.
+// Returns an Task.
```
dataclouder:
```suggestion
// Returns a Task and an error
```

govcd/adminvdc.go
```diff
@@ -52,33 +55,273 @@ func (adminVdc *AdminVdc) Refresh() error {
...
+		return Task{}, fmt.Errorf("function CreateVdcAsync is not defined for %s", "vdc"+apiVersion)
+	}
+
+	util.Logger.Printf("[DEBUG] CreateOrgVdcAsync call function for version %s", realFunction.SupportedVersion)
+
+	return realFunction.CreateVdcAsync(adminOrg, vdcConfiguration)
+}
+
+// createVdc creates a VDC with the given params under the given organization.
+// Returns an Vdc.
```
dataclouder:
```suggestion
// Returns a Vdc pointer and an error
```

govcd/adminvdc.go
```diff
@@ -52,33 +55,273 @@ func (adminVdc *AdminVdc) Refresh() error {
...
+}
+
+// CreateOrgVdcAsync creates a VDC with the given params under the given organization.
+// Returns an Task.
+func (adminOrg *AdminOrg) CreateOrgVdcAsync(vdcConfiguration *types.VdcConfiguration) (Task, error) {
+	apiVersion, err := adminOrg.client.maxSupportedVersion()
+	if err != nil {
+		return Task{}, err
+	}
+	realFunction, ok := vdcProducerByVersion["vdc"+apiVersionToVcdVersion[apiVersion]]
```
dataclouder: Rather than realFunction, a better name here would be producer

govcd/adminvdc.go
```diff
@@ -52,33 +55,273 @@ func (adminVdc *AdminVdc) Refresh() error {
...
 	if err != nil {
-		return AdminVdc{}, err
+		return nil, err
+	}
+
+	return adminVdc, nil
+}
+
+// updateVdcAsyncV97 updates a VDC with the given params. Supports Flex type allocation.
+// Needs vCD 9.7+ to work. Returns an Task.
```
dataclouder:
```suggestion
// Needs vCD 9.7+ to work. Returns a Task and an error
```

govcd/adminvdc.go
```diff
@@ -52,33 +55,273 @@ func (adminVdc *AdminVdc) Refresh() error {
...
+	}
+
+	vdc, err := adminOrg.GetVDCByName(vdcConfiguration.Name, true)
+	if err != nil {
+		return nil, err
+	}
+	return vdc, nil
+}
+
+// updateVdcAsync updates a VDC with the given params. Returns an Task.
```
dataclouder:
```suggestion
// updateVdcAsync updates a VDC with the given params. Returns a Task and an error
```

govcd/adminvdc.go
```diff
@@ -52,33 +55,273 @@ func (adminVdc *AdminVdc) Refresh() error {
...
+	}
+	err = adminVdc.Refresh()
+	if err != nil {
+		return nil, err
 	}
+	return adminVdc, nil
+}
+
+// createVdcAsync creates a VDC with the given params under the given organization.
+// Returns an Task.
```
dataclouder:
```suggestion
// Returns a Task and an error
```

govcd/adminvdc.go
```diff
@@ -52,33 +55,273 @@ func (adminVdc *AdminVdc) Refresh() error {
...
+func createVdcAsync(adminOrg *AdminOrg, vdcConfiguration *types.VdcConfiguration) (Task, error) {
+	util.Logger.Printf("[TRACE] createVdcAsync called %#v", *vdcConfiguration)
+	return adminOrg.CreateVdc(vdcConfiguration)
+}
 
-	return *adminVdc, nil
+// createVdcAsyncV97 creates a VDC with the given params under the given organization
+// and waits for the asynchronous task to complete. Supports Flex type allocation.
+// Needs vCD 9.7+ to work. Returns an Vdc.
```
dataclouder:
```suggestion
// Needs vCD 9.7+ to work. Returns a Vdc pointer and an error
```

govcd/adminvdc.go
```diff
@@ -52,33 +55,273 @@ func (adminVdc *AdminVdc) Refresh() error {
...
+
+	vdc, err := adminOrg.GetVDCByName(vdcConfiguration.Name, true)
+	if err != nil {
+		return nil, err
+	}
+	return vdc, nil
+}
+
+// createVdcAsyncV97 creates a VDC with the given params under the given organization. Supports Flex type allocation.
+// Needs vCD 9.7+ to work. Returns an Task.
```
dataclouder:
```suggestion
// Needs vCD 9.7+ to work. Returns a Task and an error
```

govcd/adminvdc_test.go
```diff
@@ -0,0 +1,239 @@
...
+	"fmt"
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+	"math"
+
+	. "gopkg.in/check.v1"
+)
+
+// Tests org function GetVDCByName with the vdc specified
+// in the config file. Then tests with a vdc that doesn't exist.
+// Fails if the config file name doesn't match with the found vdc, or
```
dataclouder:
```suggestion
// Fails if the config file name doesn't match with the found VDC, or
```

govcd/adminvdc_test.go
```diff
@@ -0,0 +1,239 @@
...
+// if the invalid vdc is found by the function.  Also tests an vdc
```
dataclouder:
```suggestion
// if the invalid VDC is found by the function.  Also tests a VDC
```

govcd/adminvdc_test.go
```diff
@@ -0,0 +1,239 @@
...
+			vdcConfiguration.IncludeMemoryOverhead = &trueValue
+		}
+
+		vdc, _ := adminOrg.GetVDCByName(vdcConfiguration.Name, false)
+		if vdc != nil {
+			err = vdc.DeleteWait(true, true)
+			check.Assert(err, IsNil)
+		}
+
+		// expecting fail due missing value
```
dataclouder:
```suggestion
		// expected to fail due to missing value
```

govcd/adminvdc_test.go
```diff
@@ -0,0 +1,239 @@
...
+		task, err := adminOrg.CreateOrgVdcAsync(vdcConfiguration)
+		check.Assert(err, Not(IsNil))
```
dataclouder:
```suggestion
		check.Assert(err, NotNil)
```

govcd/adminvdc_test.go
```diff
@@ -0,0 +1,239 @@
...
+		check.Assert(task, Equals, Task{})
+		// From vCD 10+ error changed
+		if vcd.client.Client.APIVCDMaxVersionIs("> 32.0") {
+			check.Assert(err.Error(), Equals, "error retrieving vdc: API Error: 400:  is not a valid unit. Please use MB")
```
dataclouder: This comparison seems too strict. Can we just check that there is an API error 400? The message may change in future API versions

govcd/adminvdc_test.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	adminOrg, vdcConfiguration, err := setupVDc(vcd, check, "Flex")
+	check.Assert(err, IsNil)
+
+	// Refresh so the new VDC shows up in the org's list
+	err = adminOrg.Refresh()
+	check.Assert(err, IsNil)
+
+	adminVdc, err := adminOrg.GetAdminVDCByName(vdcConfiguration.Name, false)
```
dataclouder:
```suggestion
	adminVdc, err := adminOrg.GetAdminVDCByName(vdcConfiguration.Name, true)
```
Using this construct will make the explicit refresh above unnecessary

govcd/api_vcd_versions.go
```diff
@@ -26,6 +26,27 @@ type SupportedVersions struct {
 	VersionInfos `xml:"VersionInfo"`
 }
 
+// map allows you get vCD version from max supported API version
```
dataclouder:
```suggestion
// apiVersionToVcdVersion gets the vCD version from max supported API version
```

govcd/api_vcd_versions.go
```diff
@@ -26,6 +26,27 @@ type SupportedVersions struct {
...
+var apiVersionToVcdVersion = map[string]string{
+	"29.0": "9.0",
+	"30.0": "9.1",
+	"31.0": "9.5",
+	"32.0": "9.7",
+	"33.0": "10.0",
+}
+
+// map allows you get max supported API version from vCD version
```
dataclouder:
```suggestion
// vcdVersionToApiVersion gets the max supported API version from vCD version
```

govcd/adminvdc.go
vbauzysvmware: changed

govcd/adminvdc.go
vbauzysvmware: changed

govcd/adminvdc.go
vbauzysvmware: changed

govcd/adminvdc.go
vbauzysvmware: changed

govcd/adminvdc.go
vbauzysvmware: changed

govcd/adminvdc.go
vbauzysvmware: changed

govcd/adminvdc.go
vbauzysvmware: changed

govcd/adminvdc.go
vbauzysvmware: changed

govcd/adminvdc.go
vbauzysvmware: changed

govcd/adminvdc.go
vbauzysvmware: changed

govcd/adminvdc_test.go
vbauzysvmware: changed

govcd/adminvdc_test.go
vbauzysvmware: changed

govcd/adminvdc_test.go
vbauzysvmware: changed

govcd/adminvdc_test.go
vbauzysvmware: changed

govcd/adminvdc_test.go
vbauzysvmware: changed

govcd/adminvdc_test.go
vbauzysvmware: changed

govcd/api_vcd_versions.go
vbauzysvmware: changed

govcd/api_vcd_versions.go
vbauzysvmware: changed

govcd/api_vcd_versions.go
```diff
@@ -26,6 +26,27 @@ type SupportedVersions struct {
...
+var vcdVersionToApiVersion = map[string]string{
+	"9.0":  "29.0",
+	"9.1":  "30.0",
+	"9.5":  "31.0",
+	"9.7":  "32.0",
+	"10.0": "33.0",
+}
+
+// to make apiVersionToVcdVersion used
+var _ = vcdVersionToApiVersion
```
lvirbalas: The above comment uses the old name. BTW, would you extended the comment to say _why_ does it need to be used? Also, why only one of these maps gets this treatment, but not both?

govcd/adminvdc.go
lvirbalas: Yes, I know, my question was why in this particular case we are capping to 32. I think we can address all this uncertainty by polishing the comment:

> //new version API additionally returns IncludeMemoryOverhead and IsElastic

Let's try to make it specific ("new" will become misleading after one year) and elaborate on the **why** we are using the `GetSpecificApiVersionOnCondition` here. Nit: there's a space missing at the beginning ;) What about something like this?

> We are executing below request with a specific API version in the header, because we want to retrieve the most available fields in AdminVdc which vCD provides, but also which our code understands. As we can't blindly use the latest version, we're limiting the highest used version to the one we support with the GetSpecificApiVersionOnCondition(...) function. Specifically, the API version 32 returns two additional fields: IncludeMemoryOverhead and IsElastic

vbauzysvmware: added.

govcd/api_vcd_versions.go
vbauzysvmware: Comment fixed. This map also was added as it was requested to have ready maps to convert API to vCD version and versa. This variable initialization is needed as for a moment that map isn't used.

govcd/adminvdc.go
lvirbalas: Thanks. One more thought, let's also add that this is needed for Flex:

> two additional fields for Flex allocation model: IncludeMemoryOverhead and IsElastic

vbauzysvmware: done

CODING_GUIDELINES.md
```diff
@@ -151,6 +151,55 @@ prior to attempting a search.
 Note: We are in the process of replacing methods that don't adhere to the above principles (for example, return a
 structure instead of a pointer, return a nil error on not-found, etc).
 
+## Implementing functions to support different API versions
+
+Functions dealing with different versions should use a matrix structure to identify which calls to run according to the 
+highest API version supported by vCD. An example can be found in adminvdc.go.
+
+```
+type vdcProducer struct {
```
lvirbalas: In the light of possible design patterns, I think we are misnaming this. Maybe I'm missing something here and if that's the case please share _a link_ to the design pattern that matches current implementation, but my reasoning is:

a.) Yes, there is a **producer**-consumer pattern, but it is used for **multi-threading** use cases where producer is generating data and consumer - consuming it. E.g.:
https://tpaschalis.github.io/golang-producer-consumer/

b.) Instead, what we are doing here is returning specific function implementations depending on the vCD version, which in my view much better maps to the **factory** pattern. E.g. (though not specifically our use case):
https://www.sohamkamani.com/blog/golang/2018-06-20-golang-factory-patterns/
https://travix.io/factory-method-pattern-in-go-2e09b233453e

Hence, what about naming all these a `factory` as opposed to `producer`? Or any better pattern name that we are actually using here?

CC: @dataclouder @Didainius

CODING_GUIDELINES.md
```diff
@@ -151,6 +151,55 @@ prior to attempting a search.
 Note: We are in the process of replacing methods that don't adhere to the above principles (for example, return a
 structure instead of a pointer, return a nil error on not-found, etc).
 
+## Implementing functions to support different API versions
+
+Functions dealing with different versions should use a matrix structure to identify which calls to run according to the 
+highest API version supported by vCD. An example can be found in adminvdc.go.
```
lvirbalas: Knowing how crucial this versioning system is, I'd like to see more explanation how it works & how to use it. We can address other comments and once the details stabilize elaborate this write-up.

govcd/adminorg.go
```diff
@@ -726,8 +729,8 @@ func (adminOrg *AdminOrg) GetAdminVDCByHref(vdcHref string) (*AdminVdc, error) {
 
 	adminVdc := NewAdminVdc(adminOrg.client)
 
-	_, err := adminOrg.client.ExecuteRequest(vdcHref, http.MethodGet,
-		"", "error getting vdc: %s", nil, adminVdc.AdminVdc)
+	_, err := adminOrg.client.ExecuteRequestWithApiVersion(vdcHref, http.MethodGet,
+		"", "error getting vdc: %s", nil, adminVdc.AdminVdc, adminVdc.client.GetSpecificApiVersionOnCondition(">= 32.0", "32.0"))
```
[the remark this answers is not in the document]
vbauzysvmware: added

govcd/adminvdc.go
```diff
@@ -52,33 +59,273 @@ func (adminVdc *AdminVdc) Refresh() error {
...
+
+var vdcCrudV97 = vdcProducer{
+	SupportedVersion: "32.0",
+	CreateVdc:        createVdcV97,
+	CreateVdcAsync:   createVdcAsyncV97,
+	UpdateVdc:        updateVdcV97,
+	UpdateVdcAsync:   updateVdcAsyncV97,
+}
+
+var vdcProducerByVersion = map[string]vdcProducer{
```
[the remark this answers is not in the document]
vbauzysvmware: changed

govcd/adminvdc.go
```diff
@@ -52,33 +59,273 @@ func (adminVdc *AdminVdc) Refresh() error {
...
+	util.Logger.Printf("[DEBUG] Update call function for version %s", producer.SupportedVersion)
+
+	updatedAdminVdc, err := producer.UpdateVdc(adminVdc)
 	if err != nil {
 		return AdminVdc{}, err
 	}
+	return *updatedAdminVdc, err
+}
+
+type vdcProducer struct {
```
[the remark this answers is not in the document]
vbauzysvmware: moved, added comments

govcd/adminvdc.go
```diff
@@ -52,33 +59,273 @@ func (adminVdc *AdminVdc) Refresh() error {
...
+	}
+
+	// Return the task
+	task := NewTask(adminOrg.client)
+	task.Task = adminVdc.AdminVdc.Tasks.Task[0]
+	return *task, nil
+}
+
+// validateVdcConfigurationV97 uses validateVdcConfiguration and additionally checks Flex dependent values
+func validateVdcConfigurationV97(vdcDefinition types.VdcConfiguration) error {
```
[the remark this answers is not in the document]
vbauzysvmware: IMHO,
* no - it calls internal functions if needed
* no - it calls internal functions if needed

govcd/adminvdc_test.go
```diff
@@ -0,0 +1,232 @@
...
+// if the invalid vdc is found by the function.  Also tests a VDC
+// that doesn't exist. Asserts an error if the function finds it or
+// if the error is not nil.
+func (vcd *TestVCD) Test_CreateOrgVdcWithFlex(check *C) {
+	if vcd.skipAdminTests {
+		check.Skip(fmt.Sprintf(TestRequiresSysAdminPrivileges, check.TestName()))
+	}
+
+	// Flex vCD supported from 9.7 vCD
+	if vcd.client.Client.APIVCDMaxVersionIs("< 32.0") {
```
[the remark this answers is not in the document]
vbauzysvmware: something like `allocation mode Flex isn't supported`

govcd/org_test.go
```diff
@@ -494,7 +494,7 @@ func (vcd *TestVCD) Test_GetAdminCatalog(check *C) {
 // variable is updated.
 func (vcd *TestVCD) Test_RefreshVdc(check *C) {
 
-	adminOrg, vdcConfiguration, err := setupVDc(vcd, check)
+	adminOrg, vdcConfiguration, err := setupVDc(vcd, check, "AllocationPool")
```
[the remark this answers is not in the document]
vbauzysvmware: done

govcd/adminvdc_test.go
```diff
@@ -0,0 +1,232 @@
...
+// Tests VDC by updating it and then asserting if the
+// variable is updated.
+func (vcd *TestVCD) Test_UpdateVdcFlex(check *C) {
+	if vcd.skipAdminTests {
+		check.Skip(fmt.Sprintf(TestRequiresSysAdminPrivileges, check.TestName()))
+	}
+
+	// Flex vCD supported from 9.7 vCD
+	if vcd.client.Client.APIVCDMaxVersionIs("< 32.0") {
+		check.Skip(fmt.Sprintf("Test %s requires vCD 9.7 or higher", check.TestName()))
```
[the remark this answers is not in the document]
vbauzysvmware: added

types/v56/types.go
```diff
@@ -416,6 +416,8 @@ type AdminVdc struct {
 	UsesFastProvisioning     *bool          `xml:"UsesFastProvisioning,omitempty"`
 	OverCommitAllowed        bool           `xml:"OverCommitAllowed,omitempty"`
 	VmDiscoveryEnabled       bool           `xml:"VmDiscoveryEnabled,omitempty"`
+	IsElastic                *bool          `xml:"IsElastic,omitempty"`             // supported from 32.0
+	IncludeMemoryOverhead    *bool          `xml:"IncludeMemoryOverhead,omitempty"` // supported from 32.0
```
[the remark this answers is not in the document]

types/v56/types.go
```diff
@@ -416,6 +416,8 @@ type AdminVdc struct {
 	UsesFastProvisioning     *bool          `xml:"UsesFastProvisioning,omitempty"`
 	OverCommitAllowed        bool           `xml:"OverCommitAllowed,omitempty"`
 	VmDiscoveryEnabled       bool           `xml:"VmDiscoveryEnabled,omitempty"`
+	IsElastic                *bool          `xml:"IsElastic,omitempty"`             // supported from 32.0
```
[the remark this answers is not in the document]

types/v56/types.go
```diff
@@ -459,6 +461,8 @@ type VdcConfiguration struct {
 	UsesFastProvisioning     bool                 `xml:"UsesFastProvisioning,omitempty"`
 	OverCommitAllowed        bool                 `xml:"OverCommitAllowed,omitempty"`
 	VmDiscoveryEnabled       bool                 `xml:"VmDiscoveryEnabled,omitempty"`
+	IsElastic                *bool                `xml:"IsElastic,omitempty"`             // supported from 32.0
```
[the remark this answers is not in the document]

types/v56/types.go
```diff
@@ -459,6 +461,8 @@ type VdcConfiguration struct {
...
+	IncludeMemoryOverhead    *bool                `xml:"IncludeMemoryOverhead,omitempty"` // supported from 32.0
```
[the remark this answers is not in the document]

CODING_GUIDELINES.md
```diff
@@ -151,6 +151,55 @@ prior to attempting a search.
...
+	UpdateVdc:        updateVdcV97,
+	UpdateVdcAsync:   updateVdcAsyncV97,
+}
+
+func (adminOrg *AdminOrg) CreateOrgVdc(vdcConfiguration *types.VdcConfiguration) (*Vdc, error) {
+	apiVersion, err := adminOrg.client.maxSupportedVersion()
+	if err != nil {
+		return nil, err
+	}
+	producer, ok := vdcProducerByVersion["vdc"+vcdVersionToApiVersion[apiVersion]]
```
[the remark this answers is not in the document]
vbauzysvmware: added

CODING_GUIDELINES.md
vbauzysvmware: Please tell more what is missing to make it more elaborated.

CODING_GUIDELINES.md
Didainius: If we are to be picky with names I would say this is a lot simpler than some factory because it is tied to `vdc` handling only. This is just struct with function definitions for specific version and the `var vdcProducerByVcdVersion = map[string]vdcProducer` is just a map holding it.


I could go with something like 
`type vdcVersionCrud struct`
`type vdcVersionedFunctions struct`
`type vdcFunctionsByVersion struct`

and name of the map like
`var vdcVersionMapCrud = map[string]vdcVersionedCrud`

govcd/adminvdc.go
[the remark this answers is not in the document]
lvirbalas: Thanks and comments do help.

CODING_GUIDELINES.md
lvirbalas: > because it is tied to vdc handling only

Well, this scheme will incrementally grow into all other files as well wherever we'll need to add a feature available in the newer versions of vCD. So, let's pick the best matching name for this producer/factory/map :)

To add to your suggestions:
`vdcCrudByVcdVersion`
But I don't really like it, because it is likely that such map may contain not only CRUD functions. The fact that we already have a function `validateVdcConfigurationV97` is a good indication of that.

vbauzysvmware: I don't agree that `validateVdcConfigurationV97` is good example. Mapping is for main points main functions mapping. Validation it's inside info. I don't expect them to go to mapping as we may create spaghetti in that way very soon.

lvirbalas: @vbauzysvmware Also one more question about the approach. Right now you're "overloading" only Create and Update functions. How would it look like in code if you'd need to "overload" some Get functions as well? Imagine you need to add multi-version support for `adminorg.go`.

govcd/adminvdc.go
```diff
@@ -52,33 +95,241 @@ func (adminVdc *AdminVdc) Refresh() error {
...
+
+// createVdcAsync creates a VDC with the given params under the given organization.
+// Returns a Task and an error
+func createVdcAsync(adminOrg *AdminOrg, vdcConfiguration *types.VdcConfiguration) (Task, error) {
+	util.Logger.Printf("[TRACE] createVdcAsync called %#v", *vdcConfiguration)
+	return adminOrg.CreateVdc(vdcConfiguration)
+}
 
-	return *adminVdc, nil
+// createVdcAsyncV97 creates a VDC with the given params under the given organization
```
lvirbalas: Function name in the comment doesn't match actual function name.

CODING_GUIDELINES.md
vbauzysvmware: It would like as it is now. It's easy to add get, delete or any function. You just enhance interface, add mappings and change main function get/delete to call private functions by version which are defined in already mentioned maps. It's 2 cents todo that.

govcd/adminvdc.go
vbauzysvmware: fixed

govcd/adminvdc.go
```diff
@@ -52,33 +95,241 @@ func (adminVdc *AdminVdc) Refresh() error {
...
+		return nil, fmt.Errorf("function CreateVdc is not defined for %s", "vdc"+apiVersion)
+	}
+
+	util.Logger.Printf("[DEBUG] CreateOrgVdc call function for version %s", producer.SupportedVersion)
+	return producer.CreateVdc(adminOrg, vdcConfiguration)
+}
+
+// CreateOrgVdcAsync creates a VDC with the given params under the given organization.
+// Returns a Task and an error.
+func (adminOrg *AdminOrg) CreateOrgVdcAsync(vdcConfiguration *types.VdcConfiguration) (Task, error) {
```
lvirbalas: Correct me if I'm wrong, but previously CRUD functions for VDC were in `adminorg.go`. With these changes we have functions for older vCD versions in `adminorg.go`, but functions for vCD 9.7+ in `adminvdc.go`. Can we have the same type of functions in a single file, or is there a good reason behind such move?

CC: @dataclouder @Didainius

CODING_GUIDELINES.md
vbauzysvmware: > > because it is tied to vdc handling only
> 
> Well, this scheme will incrementally grow into all other files as well wherever we'll need to add a feature available in the newer versions of vCD. So, let's pick the best matching name for this producer/factory/map :)
> 
> To add to your suggestions:
> `vdcCrudByVcdVersion`
> But I don't really like it, because it is likely that such map may contain not only CRUD functions. The fact that we already have a function `validateVdcConfigurationV97` is a good indication of that.

Also keep in mind, that aim is to have separate maps for different types. So these maps are only for vDC functions. If we have specific functions to need mapping, then we will now how to name it.

lvirbalas: @vbauzysvmware Will it always be exclusive to CRUD functions or any function may land into this map?

govcd/adminvdc.go
vbauzysvmware: In early days we try to keep with hierarchy type of storing functions in parent if parent is `receiver`, but later we chose directly hold functions in files according type. So everything we do new goes to their own file, not in parents. Version 3.0 is good option to cleanup and move functions accordingly.

CODING_GUIDELINES.md
vbauzysvmware: For this moment I don't see any other function which needs such thing unless CRUD

govcd/adminvdc.go
lvirbalas: Is there a reason not to move the touched functions from `adminorg.go` to `adminvdc.go` those functions now, so we have them in a single place and have a tidiness from the perspective of CRUD functions? Now it's pretty hard to navigate the code trying to understand where are 9.7+ functions and where are the older ones.

vbauzysvmware: I can move

vbauzysvmware: done

lvirbalas: > In early days we try to keep with hierarchy type of storing functions in parent if parent is receiver, but later we chose directly hold functions in files according type. So everything we do new goes to their own file, not in parents.

> I can move

> done

I was actually referring only to the `CreateVdc` and `CreateVdc` functions, so they'd all be in a single file with other `Create...` ones, but I'm OK with this move if everyone else is. Summary:

131 lines moved from `adminvdc.go` to `adminorg.go`, functions that moved:
- GetAdminVdcByName
- GetAdminVDCByHref
- GetAdminVDCByName
- GetAdminVDCById
- GetAdminVDCByNameOrId
- CreateVdc
- CreateVdc

Commit: https://github.com/vmware/go-vcloud-director/commit/65965d28eb114e83df53d1fd9fa9dfb2ca8c9922

CC: @Didainius @dataclouder

CODING_GUIDELINES.md
lvirbalas: Then a few suggestions based on all of the above:

a.)
Screenshot 2020-02-10 at 16 45 05

b.)
Screenshot 2020-02-10 at 16 40 42

Thoughts?

Didainius: a). It gives a hint that this is related to versions.

vbauzysvmware: For me `a)` better. P.s. as matter of fact, current `producer` was inclined and agreed with Giuseppe.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 * Improved functions to not expect XML namespaces provided in argument structure [#284](https://github.com/vmware/go-vcloud-director/pull/284)
 * Change `int` and `bool` fields from types.VAppTemplateLeaseSettings and VAppLeaseSettings into pointers
 * Added method `catalog.GetVappTemplateByHref`, and expose methods `vdc.GetEdgeGatewayByHref` and `vdc.GetEdgeGatewayRecordsType`
+* Added methods `adminOrg.CreateOrgVdc`, `adminOrg.CreateOrgVdcAsync` and improved existing to support Flex VDC model. These new methods are dynamic as they change invocation behind the scenes based on vCD version
```
Didainius: github reference missing in both listed points

govcd/adminorg.go
```diff
@@ -721,67 +633,31 @@ func (adminOrg *AdminOrg) GetVDCByNameOrId(identifier string, refresh bool) (*Vd
 	return entity.(*Vdc), err
 }
 
-// GetVDCByHref retrieves a VDC using a direct call with the HREF
-func (adminOrg *AdminOrg) GetAdminVDCByHref(vdcHref string) (*AdminVdc, error) {
-
-	adminVdc := NewAdminVdc(adminOrg.client)
+// If user specifies valid vdc name then this returns a vdc object.
```
Didainius: // GetVdcByName....

CHANGELOG.md
vbauzysvmware: done

CODING_GUIDELINES.md
lvirbalas: I believe this last choice, though longer, is clearer, because:
- Just by looking at the var name `vdcVersionedFuncs...` you know that it is about versioning.
- The names of all the four vars are consistent among each other.
- The fact that we have "version" in the names make them handily inline with related var names of `api_vcd_versions.go` (`VersionInfo`, `SupportedVersions`, `apiVersionToVcdVersion`, etc.)
- It doesn't imply the producer-consumer multi-threading pattern.

vbauzysvmware: done

CODING_GUIDELINES.md
lvirbalas: Actually, with the new var naming it is becoming quite self-explanatory, I'll suggest some improvements in-line.

CODING_GUIDELINES.md
```diff
@@ -151,6 +151,63 @@ prior to attempting a search.
 Note: We are in the process of replacing methods that don't adhere to the above principles (for example, return a
 structure instead of a pointer, return a nil error on not-found, etc).
 
+## Implementing functions to support different API versions
```
lvirbalas:
```suggestion
## Implementing functions to support different vCD API versions
```

CODING_GUIDELINES.md
```diff
@@ -151,6 +151,63 @@ prior to attempting a search.
...
+
+Functions dealing with different versions should use a matrix structure to identify which calls to run according to the 
+highest API version supported by vCD. An example can be found in adminvdc.go.
```
lvirbalas:
```suggestion
highest API version supported by vCD. An example can be found in adminvdc.go.

Note: use this pattern for adding new vCD functionality, which is not available in the earliest API version supported by the code base (as indicated by the `Client.APIVersion`).
```

CODING_GUIDELINES.md
lvirbalas: Thanks, two comments:
1.) Proposal was plural `Funcs` but you made singular `Func` - was this on purpose?
2.) `producer` is still used as var across the files, it should be changed to `vdcFuncs` or something like that.

vbauzysvmware: done

CODING_GUIDELINES.md
vbauzysvmware: done

CODING_GUIDELINES.md
vbauzysvmware: done

CODING_GUIDELINES.md
```diff
@@ -151,6 +151,65 @@ prior to attempting a search.
 Note: We are in the process of replacing methods that don't adhere to the above principles (for example, return a
 structure instead of a pointer, return a nil error on not-found, etc).
 
+## Implementing functions to support different vCD API versions
+
+Functions dealing with different versions should use a matrix structure to identify which calls to run according to the 
+highest API version supported by vCD. An example can be found in adminvdc.go.
+Note: use this pattern for adding new vCD functionality, which is not available in the earliest API version supported
```
lvirbalas: Nit: please add a new line above this note.

CODING_GUIDELINES.md
```diff
@@ -151,6 +151,66 @@ prior to attempting a search.
...
+var vdcVersionedFuncsV97 = vdcVersionedFuncs{
+	SupportedVersion: "32.0",
+	CreateVdc:        createVdcV97,
+	CreateVdcAsync:   createVdcAsyncV97,
+	UpdateVdc:        updateVdcV97,
+	UpdateVdcAsync:   updateVdcAsyncV97,
+}
+
+var vdcVersionedFuncsByVcdVersion = map[string]vdcVersionedFuncs{
+	"vdc9.0":  vdcCrudV90,
```
dataclouder: Please note that `vcdCrudV90` is not defined.

govcd/adminvdc.go
```diff
@@ -23,6 +25,174 @@ func NewAdminVdc(cli *Client) *AdminVdc {
 	}
 }
 
+// vdcVersionedFuncs holds interface of functions
```
dataclouder:
```suggestion
// vdcVersionedFuncs is a generic representation of a VDC creation across multiple versions
```

govcd/adminvdc.go
```diff
@@ -23,6 +25,174 @@ func NewAdminVdc(cli *Client) *AdminVdc {
...
+type vdcVersionedFuncs struct {
```
dataclouder: `vdcVersionedFuncs` describes what the type is at a grammatical level, rather than at functional level.
The reject3ed term `vdcProducer` was showing the purpose of the structure, while the current name only shows that it contains functions.

govcd/adminvdc.go
```diff
@@ -23,6 +25,174 @@ func NewAdminVdc(cli *Client) *AdminVdc {
...
+func (adminOrg *AdminOrg) GetAdminVDCByHref(vdcHref string) (*AdminVdc, error) {
+
+	adminVdc := NewAdminVdc(adminOrg.client)
+
+	// We are executing below request with a specific API version in the header, because we want to retrieve the most
+	// available fields in AdminVdc which vCD provides, but also which our code understands. As we can't blindly use
+	// the latest version, we're limiting the highest used version to the one we support with
+	// the GetSpecificApiVersionOnCondition(...) function. Specifically, the API version 32 returns
+	// two additional fields: IncludeMemoryOverhead and IsElastic for Flex allocation
+	_, err := adminOrg.client.ExecuteRequestWithApiVersion(vdcHref, http.MethodGet,
```
dataclouder: If possible, I'd like this function to be implemented similarly to the others defined in `vdcVersionedFuncs`, i.e. one function per version.

govcd/adminvdc.go
```diff
@@ -23,6 +25,174 @@ func NewAdminVdc(cli *Client) *AdminVdc {
...
+		return Task{}, err
+	}
+
+	// Return the task
+	task := NewTask(adminOrg.client)
+	task.Task = adminVdc.AdminVdc.Tasks.Task[0]
+	return *task, nil
+}
+
+// Creates the vdc and waits for the asynchronous task to complete.
```
dataclouder:
```suggestion
// Creates the VDC and waits for the asynchronous task to complete.
```

govcd/adminvdc.go
```diff
@@ -37,8 +207,13 @@ func (adminVdc *AdminVdc) Refresh() error {
...
 
-	_, err := adminVdc.client.ExecuteRequest(adminVdc.AdminVdc.HREF, http.MethodGet,
-		"", "error refreshing VDC: %s", nil, unmarshalledAdminVdc)
+	// We are executing below request with a specific API version in the header, because we want to retrieve the most
+	// available fields in AdminVdc which vCD provides, but also which our code understands. As we can't blindly use
+	// the latest version, we're limiting the highest used version to the one we support with
+	// the GetSpecificApiVersionOnCondition(...) function. Specifically, the API version 32 returns
+	// two additional fields: IncludeMemoryOverhead and IsElastic for Flex allocation
+	_, err := adminVdc.client.ExecuteRequestWithApiVersion(adminVdc.AdminVdc.HREF, http.MethodGet,
+		"", "error refreshing VDC: %s", nil, unmarshalledAdminVdc, adminVdc.client.GetSpecificApiVersionOnCondition(">= 32.0", "32.0"))
```
dataclouder: Can't we use `GetVDCByHref`  here?

CODING_GUIDELINES.md
vbauzysvmware: fixed

govcd/adminvdc.go
vbauzysvmware: fixed

govcd/adminvdc.go
vbauzysvmware: and what I can do about it?

govcd/adminvdc.go
vbauzysvmware: Possible, yet it won't bring any value just code repetition

govcd/adminvdc.go
vbauzysvmware: fixed

govcd/adminvdc.go
vbauzysvmware: recepients are different. Refresh works with adminVDC and `GetVDCByHref` is for `adminOrg`

govcd/adminvdc.go
dataclouder: Code repetition is when we repeat the same code in various places. I am saying we should have a GET for 9.0 to 9.5 and another for 9.7+, same as what we did for Create and Update

govcd/adminvdc.go
lvirbalas: @dataclouder "VDC creation" is too specific in this case, because tries to cover all CRUD operations:
Screenshot 2020-02-13 at 14 07 24
 What about changing "a VDC creation" to "VDC CRUD operations"? That would read:
`// vdcVersionedFuncs is a generic representation of VDC CRUD operations across multiple versions`

dataclouder: Yes, it's better. Although we only support creation and update, not all CRUD, since `D` is missing and the `R` was explicitly implemented with hidden version choice (see my note under `GetAdminVDCByHref` ) instead of one function in this structure

govcd/adminvdc.go
dataclouder: I still think that a purposeful name such as `provider` conveys the meaning better than `Funcs`.  How about changing it to `vdcCRUD` ? At least it suggests some purpose.
cc: @lvirbalas, @Didainius

lvirbalas: There was a long discussion about this here:
https://github.com/vmware/go-vcloud-director/pull/285#discussion_r376947094

There `vdcCrud` was proposed too, but at the end we agreed on `vdcVersionedFuncs`, because:
- It has a word "versioned", which makes it instantly clear what the pattern's purpose is
- With such name not only CRUD functions could land here if needed

govcd/adminvdc.go
vbauzysvmware: I don't see any value here until more changes needed. Now I will create same functions for sake of creating and have to make tests, and so on.

govcd/adminvdc.go
vbauzysvmware: updated

Issue #202, opened 2019-06-12
Add update functionality to edge gateway


dataclouder:
As defined in PR #195, we support edge gateway creation and deletion.

To complete the set of functionalities, we should also support updates.

While the [API for updates](https://code.vmware.com/apis/553/vcloud-director/doc/doc/operations/PUT-EdgeGateway.html) is conceptually simple, as it requires passing the same structure used for creation, determining what can be updated and how is more complicated.

For example, when we change the dependency from an external network as default gateway, we need to change the elements that we know (gateway IP and netmask) but also elements that we don't know directly (the external IP allocated or whether there is an available IP). Also, changing the external network will influence how the depending routed networks will respond.

Moreover, we need to understand which parts of the configuration can be altered without side effects (or with manageable side effects) and which should be considered non-alterable, meaning that we will need to re-deploy the edge gateway to change them.

Given the complexity of the matter, we need to decide between two approaches:

*. create an `Update` function, with a structure similar to the one used for `CreateEdgeGateway` and handle all the possibilities in there.
* create a separate function for each element that can be changed (such as `UnlinkNetwork`, `RemoveDefaultGateway`, `ChangeDefaultGateway`, and so on)

lvirbalas:
I agree about the complexity and as such like the second suggestion more (with separate functions foreach updatable element). This way we can roll out updates incrementally and manage risks better.

lvirbalas:
@slavaavr in the light of your recent PR #287 I hope you'll be interested in trying out this PR when it's ready for merge?

bogi0704:
I would like to use the update functionality. Does anyone have an update about this?

PR #287, opened 2020-02-06
Support update of Edge configuration (except Edge Services)

slavaavr:
Hi there. This PR kind of related with this issue https://github.com/vmware/go-vcloud-director/issues/202
I got it how you decided to implement Edge update, however, I think this method will not be needless. What do you think of it?

### DIFF
--- a/govcd/edgegateway.go
+++ b/govcd/edgegateway.go
@@ -1398,3 +1398,12 @@ func getNetworkNameAndTypeByVnicIndex(vNicIndex int, vnics *types.EdgeGatewayInt
 
 	return networkName, networkType, nil
 }
+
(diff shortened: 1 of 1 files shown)


dataclouder:
Thanks for your contribution.
We are already in the process of implementing a full update of Edge Gateway.
Although the update function is conceptually simple, it requires extensive additional testing to make sure we are covering all the cases.

slavaavr:
@dataclouder oh, I see. So, when do you think update function will be resealed (approximately)? And what I have to do with this PR, close it for now?

datacharmer:
I think you can close for now. We don't have an official release date yet, although our release cycles are usually 3 to 4 months

slavaavr:
@datacharmer, i got it. Thanks!

PR #288, opened 2020-02-11
Fix data race in catalog/media upload reporting 

Didainius:
As part of https://github.com/terraform-providers/terraform-provider-vcd/issues/445 there is a small issue in upload reporting which may have data races. This is not a dangerous situation because all it does is prints out the progress, but still this has to be cleaned up.


govcd/upload.go
```diff
@@ -14,11 +14,32 @@ import (
...
 	"strconv"
+	"sync"
 
 	"github.com/vmware/go-vcloud-director/v2/types/v56"
 	"github.com/vmware/go-vcloud-director/v2/util"
 )
 
+// mutemutexedProgress is a thread-safe structure to update and report progress during an UploadTask.
+//
+// Value must be read/written using Get/Set values insted of directly accessing the
```
lvirbalas: Incomplete comment.

Didainius: This was a mess. Fixed it now.

PR #290, opened 2020-02-14
Add vApp network functions

vbauzysvmware:
Add new function need for handling vApp networks:

    Add AddNetwork, AddNetworkAsync
    Add AddOrgNetwork, AddOrgNetworkAsync
    Add UpdateNetwork, UpdateNetworkAsync
    Add UpdateOrgNetwork, UpdateOrgNetworkAsync
    Add RemoveNetwork
    Add GetUuidFromHref

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 * Added methods `adminOrg.CreateOrgVdc`, `adminOrg.CreateOrgVdcAsync` and improved existing to support Flex VDC model. These new methods are dynamic as they change invocation behind the scenes based on vCD version [#285](https://github.com/vmware/go-vcloud-director/pull/285) 
 * Deprecated functions `adminOrg.CreateVdc` and `adminOrg.CreateVdcWait` [#285](https://github.com/vmware/go-vcloud-director/pull/285)
 * Added methods `vapp.AddNetwork`, `vapp.AddNetworkAsync`, `vapp.AddOrgNetwork`, `vapp.AddOrgNetworkAsync`, `vapp.UpdateNetwork`, `vapp.UpdateNetworkAsync`, `vapp.UpdateOrgNetwork`, `vapp.UpdateOrgNetworkAsync`, `vapp.RemoveNetwork`, `vapp.RemoveNetworkAsync` and `GetUuidFromHref` [#289](https://github.com/vmware/go-vcloud-director/pull/290)
-* Deprecated function `vapp.RemoveIsolatedNetwork` [#289](https://github.com/vmware/go-vcloud-director/pull/290)
+* Deprecated function `vapp.RemoveIsolatedNetwork`, `vapp.AddRAWNetworkConfig` and `vapp.AddIsolatedNetwork`  [#289](https://github.com/vmware/go-vcloud-director/pull/290)
 
 ## 2.5.1 (December 12, 2019)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
   `VM.WaitForDhcpIpByNicIndexes()`, `VM.GetParentVApp()`, `VM.GetParentVdc()`
   [#283](https://github.com/vmware/go-vcloud-director/pull/283)
 * Added methods `vapp.AddNetwork`, `vapp.AddNetworkAsync`, `vapp.AddOrgNetwork`, `vapp.AddOrgNetworkAsync`, `vapp.UpdateNetwork`, `vapp.UpdateNetworkAsync`, `vapp.UpdateOrgNetwork`, `vapp.UpdateOrgNetworkAsync`, `vapp.RemoveNetwork`, `vapp.RemoveNetworkAsync` and `GetUuidFromHref` [#289](https://github.com/vmware/go-vcloud-director/pull/290)
-* Deprecated function `vapp.RemoveIsolatedNetwork`, `vapp.AddRAWNetworkConfig` and `vapp.AddIsolatedNetwork`  [#289](https://github.com/vmware/go-vcloud-director/pull/290)
+* Deprecated functions `vapp.RemoveIsolatedNetwork`, `vapp.AddRAWNetworkConfig` and `vapp.AddIsolatedNetwork`  [#289](https://github.com/vmware/go-vcloud-director/pull/290)
 
 BUGS FIXED:
 * A data race in catalog/media item upload status reporting [#288](https://github.com/vmware/go-vcloud-director/pull/288)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 * Improved functions to not expect XML namespaces provided in argument structure [#284](https://github.com/vmware/go-vcloud-director/pull/284)
 * Change `int` and `bool` fields from types.VAppTemplateLeaseSettings and VAppLeaseSettings into pointers
 * Added method `catalog.GetVappTemplateByHref`, and expose methods `vdc.GetEdgeGatewayByHref` and `vdc.GetEdgeGatewayRecordsType`
+* Added methods `vapp.AddNetwork`, `vapp.AddNetworkAsync`, `vapp.AddOrgNetwork`, `vapp.AddOrgNetworkAsync`, `vapp.UpdateNetwork`, `vapp.UpdateNetworkAsync`, `vapp.UpdateOrgNetwork`, `vapp.UpdateOrgNetworkAsync`, `vapp.RemoveNetwork`, `vapp.RemoveNetworkAsync` and `GetUuidFromHref` [#289](https://github.com/vmware/go-vcloud-director/pull/289)
+* Deprecated function `vapp.RemoveIsolatedNetwork` [#289](https://github.com/vmware/go-vcloud-director/pull/289)
 
 ## 2.5.1 (December 12, 2019)
 
--- a/govcd/system.go
+++ b/govcd/system.go
@@ -93,24 +93,6 @@ func getBareEntityUuid(entityId string) (string, error) {
 	return matchList[0][1], nil
 }
 
-// Returns the UUID part of an HREF
-// Similar to getBareEntityUuid, but tailored to HREF
-func getUuidFromHref(href string) (string, error) {
-	// Regular expression to match an ID:
-	//     1 string starting by 'https://' and ending with a '/',
-	//     followed by
-	//        1 group of 8 hexadecimal digits
-	//        3 groups of 4 hexadecimal digits
-	//        1 group of 12 hexadecimal digits
-	reGetID := regexp.MustCompile(`^https://.+/([a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})$`)
-	matchList := reGetID.FindAllStringSubmatch(href, -1)
-
-	if len(matchList) == 0 || len(matchList[0]) < 2 {
-		return "", fmt.Errorf("error extracting UUID from '%s'", href)
-	}
-	return matchList[0][1], nil
-}
-
 // CreateEdgeGatewayAsync creates an edge gateway using a simplified configuration structure
 // https://code.vmware.com/apis/442/vcloud-director/doc/doc/operations/POST-CreateEdgeGateway.html
 //
@@ -392,7 +374,7 @@ func getOrgHREFById(vcdClient *VCDClient, orgId string) (string, error) {
 	// Look for org UUID within OrgList
 	for _, org := range orgList.Org {
 		// ID in orgList is usually empty. We extract the UUID from HREF to make the comparison
-		uuidFromHref, err := getUuidFromHref(org.HREF)
+		uuidFromHref, err := GetUuidFromHref(org.HREF, true)
 		if err != nil {
 			return "", err
 		}
@@ -924,8 +906,8 @@ func (vcdClient *VCDClient) GetAdminOrgByNameOrId(identifier string) (*AdminOrg,
 
 // Returns the UUID part of an HREF
 // Similar to getBareEntityUuid, but tailored to HREF
-func GetUuidFromHref(href string) (string, error) {
-	util.Logger.Printf("[TRACE] GetUuidFromHref got href: %s", href)
+func GetUuidFromHref(href string, idAtEnd bool) (string, error) {
+	util.Logger.Printf("[TRACE] GetUuidFromHref got href: %s with idAtEnd: %t", href, idAtEnd)
 	// Regular expression to match an ID:
 	//     1 string starting by 'https://' and ending with a '/',
 	//     followed by
@@ -933,7 +915,13 @@ func GetUuidFromHref(href string) (string, error) {
 	//        3 groups of 4 hexadecimal digits
 	//        1 group of 12 hexadecimal digits
 
-	reGetID := regexp.MustCompile(`^https://.+/([a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}).*$`)
+	searchExpression := `^https://.+/([a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})`
+	if idAtEnd {
+		searchExpression += `$`
+	} else {
+		searchExpression += `.*$`
+	}
+	reGetID := regexp.MustCompile(searchExpression)
 	matchList := reGetID.FindAllStringSubmatch(href, -1)
 
 	if len(matchList) == 0 || len(matchList[0]) < 2 {
--- a/govcd/system.go
+++ b/govcd/system.go
@@ -921,3 +921,24 @@ func (vcdClient *VCDClient) GetAdminOrgByNameOrId(identifier string) (*AdminOrg,
 	}
 	return entity.(*AdminOrg), err
 }
+
+// Returns the UUID part of an HREF
+// Similar to getBareEntityUuid, but tailored to HREF
+func GetUuidFromHref(href string) (string, error) {
+	util.Logger.Printf("[TRACE] GetUuidFromHref got href: %s", href)
+	// Regular expression to match an ID:
+	//     1 string starting by 'https://' and ending with a '/',
+	//     followed by
+	//        1 group of 8 hexadecimal digits
+	//        3 groups of 4 hexadecimal digits
+	//        1 group of 12 hexadecimal digits
+
+	reGetID := regexp.MustCompile(`^https://.+/([a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}).*$`)
+	matchList := reGetID.FindAllStringSubmatch(href, -1)
+
+	if len(matchList) == 0 || len(matchList[0]) < 2 {
+		return "", fmt.Errorf("error extracting UUID from '%s'", href)
+	}
+	util.Logger.Printf("[TRACE] GetUuidFromHref returns UUID : %s", matchList[0][1])
+	return matchList[0][1], nil
+}
--- a/govcd/vapp.go
+++ b/govcd/vapp.go
@@ -996,6 +996,13 @@ func (vapp *VApp) UpdateNetworkAsync(networkSettingsToUpdate *VappNetworkSetting
 		networkToUpdate.Configuration.Features.NatService = &types.NatService{IsEnabled: *networkSettingsToUpdate.NatEnabled, NatType: "ipTranslation", Policy: "allowTrafficIn"}
 	}
 
+	if networkToUpdate.Configuration.Features.FirewallService != nil && networkSettingsToUpdate.FirewallEnabled != nil {
+		networkToUpdate.Configuration.Features.FirewallService.IsEnabled = *networkSettingsToUpdate.FirewallEnabled
+	}
+	if networkToUpdate.Configuration.Features.FirewallService == nil && networkSettingsToUpdate.FirewallEnabled != nil {
+		networkToUpdate.Configuration.Features.FirewallService = &types.FirewallService{IsEnabled: *networkSettingsToUpdate.FirewallEnabled}
+	}
+
 	networkToUpdate.Configuration.Features.FirewallService.IsEnabled = *networkSettingsToUpdate.FirewallEnabled
 	networkToUpdate.Configuration.RetainNetInfoAcrossDeployments = networkSettingsToUpdate.RetainIpMacEnabled
 	if networkToUpdate.Configuration.ParentNetwork != nil && orgNetwork != nil && networkToUpdate.Configuration.ParentNetwork.HREF != orgNetwork.HREF {
--- a/govcd/vapp.go
+++ b/govcd/vapp.go
@@ -978,15 +978,17 @@ func (vapp *VApp) UpdateNetworkAsync(networkSettingsToUpdate *VappNetworkSetting
 	var networkToUpdate types.VAppNetworkConfiguration
 	var networkToUpdateIndex int
 	for index, networkConfig := range currentNetworkConfiguration.NetworkConfig {
-		uuid, err := GetUuidFromHref(networkConfig.Link.HREF, false)
-		if err != nil {
-			return Task{}, err
-		}
+		if networkConfig.Link != nil {
+			uuid, err := GetUuidFromHref(networkConfig.Link.HREF, false)
+			if err != nil {
+				return Task{}, err
+			}
 
-		if uuid == networkSettingsToUpdate.ID {
-			networkToUpdate = networkConfig
-			networkToUpdateIndex = index
-			break
+			if uuid == networkSettingsToUpdate.ID {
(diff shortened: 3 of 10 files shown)


govcd/vapp.go
```diff
@@ -786,6 +795,328 @@ func (vapp *VApp) AddIsolatedNetwork(newIsolatedNetworkSettings *VappNetworkSett
...
+		IsDeployed: false,
+	}
+	networkConfigurations = append(networkConfigurations,
+		vappConfiguration)
+
+	return updateNetworkConfigurations(vapp, networkConfigurations)
+
+}
+
+// UpdateNetwork updates vApp networks(isolated or connected to org VDC network)
```
lvirbalas:
```suggestion
// UpdateNetwork updates vApp networks (isolated or connected to Org VDC network)
```

govcd/vapp.go
```diff
@@ -786,6 +795,328 @@ func (vapp *VApp) AddIsolatedNetwork(newIsolatedNetworkSettings *VappNetworkSett
...
+
+	vAppNetworkConfig, err := vapp.GetNetworkConfig()
+	if err != nil {
+		return nil, fmt.Errorf("error getting vApp networks: %#v", err)
+	}
+
+	return vAppNetworkConfig, nil
+}
+
+// UpdateNetworkAsync asyncronously updates vApp networks(isolated or connected to org VDC network).
```
lvirbalas:
```suggestion
// UpdateNetworkAsync asyncronously updates vApp networks (isolated or connected to Org VDC network).
```

govcd/vapp.go
```diff
@@ -786,6 +795,328 @@ func (vapp *VApp) AddIsolatedNetwork(newIsolatedNetworkSettings *VappNetworkSett
...
+		networkToUpdate.Configuration.Features.DhcpService.MaxLeaseTime = networkSettingsToUpdate.DhcpSettings.MaxLeaseTime
+		networkToUpdate.Configuration.Features.DhcpService.IPRange = networkSettingsToUpdate.DhcpSettings.IPRange
+	}
+
+	currentNetworkConfiguration.NetworkConfig[networkToUpdateIndex] = networkToUpdate
+
+	return updateNetworkConfigurations(vapp, currentNetworkConfiguration.NetworkConfig)
+}
+
+// UpdateOrgNetwork updates to vApp added org VDC network
```
lvirbalas: Maybe this would be better:
```suggestion
// UpdateOrgNetwork updates Org VDC network which is part of a vApp
```

govcd/vapp.go
```diff
@@ -786,6 +795,328 @@ func (vapp *VApp) AddIsolatedNetwork(newIsolatedNetworkSettings *VappNetworkSett
...
+
+	vAppNetworkConfig, err := vapp.GetNetworkConfig()
+	if err != nil {
+		return nil, fmt.Errorf("error getting vApp networks: %#v", err)
+	}
+
+	return vAppNetworkConfig, nil
+}
+
+// UpdateOrgNetworkAsync asyncronously updates to vApp added org VDC network
```
lvirbalas: Maybe this:
```suggestion
// UpdateOrgNetworkAsync asyncronously updates Org VDC network which is part of a vApp
```

govcd/vapp.go
```diff
@@ -814,7 +1145,57 @@ func validateNetworkConfigSettings(networkSettings *VappNetworkSettings) error {
 	return nil
 }
 
+// RemoveNetwork removes any network from vApp
```
lvirbalas: Did you mean this:
```suggestion
// RemoveNetwork removes any network (be it isolated or connected to an Org Network) from vApp
```

govcd/vapp.go
```diff
@@ -814,7 +1145,57 @@ func validateNetworkConfigSettings(networkSettings *VappNetworkSettings) error {
...
+
+	vAppNetworkConfig, err := vapp.GetNetworkConfig()
+	if err != nil {
+		return nil, fmt.Errorf("error getting vApp networks: %#v", err)
+	}
+
+	return vAppNetworkConfig, nil
+}
+
+// RemoveNetworkAsync asyncronously removes vApp any network
```
lvirbalas: Sentence doesn't make sense.

govcd/vapp.go
```diff
@@ -814,7 +1145,57 @@ func validateNetworkConfigSettings(networkSettings *VappNetworkSettings) error {
...
+// Accepts network Id or name
```
lvirbalas:
```suggestion
// Accepts network ID or name
```

govcd/vapp.go
```diff
@@ -814,7 +1145,57 @@ func validateNetworkConfigSettings(networkSettings *VappNetworkSettings) error {
...
+func (vapp *VApp) RemoveNetworkAsync(id string) (Task, error) {
+
+	if id == "" {
+		return Task{}, fmt.Errorf("network id/name can't be empty")
```
lvirbalas:
```suggestion
		return Task{}, fmt.Errorf("network ID/name can't be empty")
```

govcd/vapp_test.go
```diff
@@ -769,3 +769,356 @@ func (vcd *TestVCD) Test_GetVM(check *C) {
...
+	if vcd.skipVappTests {
+		check.Skip("Skipping test because vapp was not successfully created at setup")
+	}
+
+	if vcd.config.VCD.Network.Net1 == "" {
+		check.Skip("Skipping test because no network was given")
+	}
+
+	// Add Metadata
+	networkName := "AddAndRemoveVappNetworkTest"
```
lvirbalas: What about changing the `networkName` so they are unique for each of the new tests? They could have a word to depict the type of the network too.

govcd/vapp.go
vbauzysvmware: updated

govcd/vapp.go
vbauzysvmware: updated

govcd/vapp.go
vbauzysvmware: updated

govcd/vapp.go
vbauzysvmware: updated

govcd/vapp.go
vbauzysvmware: updated

govcd/vapp.go
vbauzysvmware: updated

govcd/vapp.go
vbauzysvmware: updated

govcd/vapp.go
vbauzysvmware: updated

govcd/vapp_test.go
vbauzysvmware: changed

govcd/system.go
```diff
@@ -921,3 +921,24 @@ func (vcdClient *VCDClient) GetAdminOrgByNameOrId(identifier string) (*AdminOrg,
 	}
 	return entity.(*AdminOrg), err
 }
+
+// Returns the UUID part of an HREF
+// Similar to getBareEntityUuid, but tailored to HREF
+func GetUuidFromHref(href string) (string, error) {
```
Didainius: A few questions - do we really need this to be exported? Maybe it could work internally only?
Also - is there any different between our current private function `getUuidFromHref`? Can they be wrapped or reused?

govcd/vapp.go
```diff
@@ -786,6 +798,328 @@ func (vapp *VApp) AddIsolatedNetwork(newIsolatedNetworkSettings *VappNetworkSett
...
+	if err != nil {
+		return Task{}, err
+	}
+
+	// for case when range is one ip address
+	if newNetworkSettings.DhcpSettings != nil && newNetworkSettings.DhcpSettings.IPRange != nil && newNetworkSettings.DhcpSettings.IPRange.EndAddress == "" {
+		newNetworkSettings.DhcpSettings.IPRange.EndAddress = newNetworkSettings.DhcpSettings.IPRange.StartAddress
+	}
+
+	// explicitly check if to add data, to not send any values
```
Didainius: Could you rephrase, this one is hard to follow.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 * Added method `catalog.GetVappTemplateByHref`, and expose methods `vdc.GetEdgeGatewayByHref` and `vdc.GetEdgeGatewayRecordsType`
 * Added methods `adminOrg.CreateOrgVdc`, `adminOrg.CreateOrgVdcAsync` and improved existing to support Flex VDC model. These new methods are dynamic as they change invocation behind the scenes based on vCD version [#285](https://github.com/vmware/go-vcloud-director/pull/285) 
 * Deprecated functions `adminOrg.CreateVdc` and `adminOrg.CreateVdcWait` [#285](https://github.com/vmware/go-vcloud-director/pull/285)
+* Added methods `vapp.AddNetwork`, `vapp.AddNetworkAsync`, `vapp.AddOrgNetwork`, `vapp.AddOrgNetworkAsync`, `vapp.UpdateNetwork`, `vapp.UpdateNetworkAsync`, `vapp.UpdateOrgNetwork`, `vapp.UpdateOrgNetworkAsync`, `vapp.RemoveNetwork`, `vapp.RemoveNetworkAsync` and `GetUuidFromHref` [#289](https://github.com/vmware/go-vcloud-director/pull/290)
+* Deprecated function `vapp.RemoveIsolatedNetwork` [#289](https://github.com/vmware/go-vcloud-director/pull/290)
```
Didainius: As far as I see this is not the only deprecated function. At least `vapp.AddIsolatedNetwork`, `AddRAWNetworkConfig` are also deprecated.

govcd/vapp.go
```diff
@@ -814,7 +1148,57 @@ func validateNetworkConfigSettings(networkSettings *VappNetworkSettings) error {
...
+	if err != nil {
+		return nil, fmt.Errorf("error getting vApp networks: %#v", err)
+	}
+
+	return vAppNetworkConfig, nil
+}
+
+// RemoveNetworkAsync asyncronously removes any network (be it isolated or connected to an Org Network) from vApp
+// Accepts network ID or name
+func (vapp *VApp) RemoveNetworkAsync(id string) (Task, error) {
```
Didainius: I believe we use such variable name when name and id are accepted.

```suggestion
func (vapp *VApp) RemoveNetworkAsync(identifier string) (Task, error) {
```

govcd/vapp.go
```diff
@@ -35,15 +35,21 @@ func (vcdCli *VCDClient) NewVApp(client *Client) VApp {
...
-	Name             string
-	Gateway          string
-	NetMask          string
-	DNS1             string
-	DNS2             string
-	DNSSuffix        string
-	GuestVLANAllowed *bool
-	StaticIPRanges   []*types.IPRange
-	DhcpSettings     *DhcpSettings
+	Id                 string
```
Didainius: I think we are using capital cased ID strings in types.

```suggestion
	ID                 string
```

govcd/vapp.go
```diff
@@ -786,6 +798,328 @@ func (vapp *VApp) AddIsolatedNetwork(newIsolatedNetworkSettings *VappNetworkSett
...
+// If orgNetwork is nil, then isolated network created.
+func (vapp *VApp) AddNetworkAsync(newNetworkSettings *VappNetworkSettings, orgNetwork *types.OrgVDCNetwork) (Task, error) {
+
+	err := validateNetworkConfigSettings(newNetworkSettings)
+	if err != nil {
+		return Task{}, err
+	}
+
+	// for case when range is one ip address
+	if newNetworkSettings.DhcpSettings != nil && newNetworkSettings.DhcpSettings.IPRange != nil && newNetworkSettings.DhcpSettings.IPRange.EndAddress == "" {
```
Didainius: Is there some specific scenario when it is specified like that? As far as I have seen ranges usually have start and end addresses populated.

govcd/vapp.go
```diff
@@ -786,6 +798,328 @@ func (vapp *VApp) AddIsolatedNetwork(newIsolatedNetworkSettings *VappNetworkSett
...
+		}
+	}
+
+	if newNetworkSettings.FirewallEnabled != nil {
+		if networkFeatures == nil {
+			networkFeatures = &types.NetworkFeatures{}
+		}
+		networkFeatures.FirewallService = &types.FirewallService{IsEnabled: *newNetworkSettings.FirewallEnabled}
+	}
+	if newNetworkSettings.FirewallEnabled != nil && newNetworkSettings.NatEnabled != nil {
```
Didainius: Does it imply that firewall must be enabled so that NAT rules can be enabled? This looks to be handled a bit differently in `AddOrgNetworkAsync`

govcd/vapp.go
```diff
@@ -786,6 +798,328 @@ func (vapp *VApp) AddIsolatedNetwork(newIsolatedNetworkSettings *VappNetworkSett
...
+
+	// for case when range is one ip address
+	if newNetworkSettings.DhcpSettings != nil && newNetworkSettings.DhcpSettings.IPRange != nil && newNetworkSettings.DhcpSettings.IPRange.EndAddress == "" {
+		newNetworkSettings.DhcpSettings.IPRange.EndAddress = newNetworkSettings.DhcpSettings.IPRange.StartAddress
+	}
+
+	// explicitly check if to add data, to not send any values
+	var networkFeatures *types.NetworkFeatures
+	if newNetworkSettings.DhcpSettings != nil {
+		networkFeatures = &types.NetworkFeatures{DhcpService: &types.DhcpService{
```
Didainius: This assignment looks like 1:1 copy. Maybe it could be directly assigned?

govcd/vapp.go
```diff
@@ -786,6 +798,328 @@ func (vapp *VApp) AddIsolatedNetwork(newIsolatedNetworkSettings *VappNetworkSett
...
+	if newNetworkSettings.FirewallEnabled != nil && newNetworkSettings.NatEnabled != nil {
+		networkFeatures.NatService = &types.NatService{IsEnabled: *newNetworkSettings.NatEnabled, NatType: "ipTranslation", Policy: "allowTrafficIn"}
+	}
+
+	networkConfigurations := vapp.VApp.NetworkConfigSection.NetworkConfig
+	vappConfiguration := types.VAppNetworkConfiguration{
+		NetworkName: newNetworkSettings.Name,
+		Description: newNetworkSettings.Description,
+		Configuration: &types.NetworkConfiguration{
+			FenceMode:        types.FenceModeNAT,
```
Didainius: Is it correct even when isolated network is being created?

govcd/vapp.go
```diff
@@ -786,6 +798,328 @@ func (vapp *VApp) AddIsolatedNetwork(newIsolatedNetworkSettings *VappNetworkSett
...
+			networkToUpdateIndex = index
+			break
+		}
+	}
+
+	if networkToUpdate == (types.VAppNetworkConfiguration{}) {
+		return Task{}, fmt.Errorf("not found network to update with Id %s", networkSettingsToUpdate.Id)
+	}
+
+	if networkToUpdate.Configuration.Features.NatService != nil && networkSettingsToUpdate.NatEnabled != nil {
```
Didainius: Would it be a big risk/breaking change to use `*bool` and assign directly instead of converting each value?

govcd/vapp.go
```diff
@@ -786,6 +798,328 @@ func (vapp *VApp) AddIsolatedNetwork(newIsolatedNetworkSettings *VappNetworkSett
...
+	if networkToUpdate.Configuration.ParentNetwork != nil && orgNetwork != nil && networkToUpdate.Configuration.ParentNetwork.HREF != orgNetwork.HREF {
+		networkToUpdate.Configuration.ParentNetwork = &types.Reference{HREF: orgNetwork.HREF}
+	}
+	if networkToUpdate.Configuration.ParentNetwork != nil && orgNetwork == nil {
+		networkToUpdate.Configuration.ParentNetwork = nil
+	}
+	networkToUpdate.Description = networkSettingsToUpdate.Description
+	networkToUpdate.NetworkName = networkSettingsToUpdate.Name
+	networkToUpdate.Configuration.GuestVlanAllowed = networkSettingsToUpdate.GuestVLANAllowed
+	networkToUpdate.Configuration.IPScopes.IPScope[0].Gateway = networkSettingsToUpdate.Gateway
```
Didainius: Are we not binding ourselves to first object while many IP ranges can exist?

govcd/vapp.go
```diff
@@ -786,6 +798,328 @@ func (vapp *VApp) AddIsolatedNetwork(newIsolatedNetworkSettings *VappNetworkSett
...
+	networkToUpdate.Configuration.IPScopes.IPScope[0].Netmask = networkSettingsToUpdate.NetMask
+	networkToUpdate.Configuration.IPScopes.IPScope[0].DNS1 = networkSettingsToUpdate.DNS1
+	networkToUpdate.Configuration.IPScopes.IPScope[0].DNS2 = networkSettingsToUpdate.DNS2
+	networkToUpdate.Configuration.IPScopes.IPScope[0].DNSSuffix = networkSettingsToUpdate.DNSSuffix
+	networkToUpdate.Configuration.IPScopes.IPScope[0].IPRanges = &types.IPRanges{IPRange: networkSettingsToUpdate.StaticIPRanges}
+
+	// for case when range is one ip address
+	if networkSettingsToUpdate.DhcpSettings != nil && networkSettingsToUpdate.DhcpSettings.IPRange != nil && networkSettingsToUpdate.DhcpSettings.IPRange.EndAddress == "" {
```
Didainius: Similar question to the one in other place - are there cases when only one field is specified?

govcd/system.go
vbauzysvmware: This functions used widely in terraform also due that we need get ID from href -> specific vapp network handling. Private function is slight different - regular expression is slight changed.

govcd/vapp.go
vbauzysvmware: updated

CHANGELOG.md
vbauzysvmware: Thank you. Added

govcd/vapp.go
vbauzysvmware: changed

govcd/vapp.go
vbauzysvmware: changed

govcd/vapp.go
vbauzysvmware: It's according how you provide data. `newNetworkSettings` allows you provide such way of data

govcd/vapp.go
vbauzysvmware: made to be the same in AddOrgNetworkAsync

govcd/vapp.go
vbauzysvmware: Types are different DhcpService and DhcpSettings

govcd/vapp.go
vbauzysvmware: It used by other functions - which will be impacted, needs to be verified, etc.

govcd/vapp.go
vbauzysvmware: Well, in UI you can't have more gateways, etc. So if you know how to verify need of that when we consider changing a lot this PR.

govcd/vapp.go
vbauzysvmware: There is possibility of that.

govcd/system.go
dataclouder: The only difference between the regular expressions in public and private functions is that the public one allows text after the UUID.
I think that, instead of adding a function that is 99% the same as another, we should add a parameter to match  either open ended or closed strings.
Alternatively, the two functions should have a different name, because `getUuidFromHref` matches an entity UUID (ID at the end), while `GetUuidFromHref` matches an UUID from an action HREF (ID/action at the end). 

Given that the action is discarded, I suggest this interface:

```go
func GetUuidFromHref(href string, idAtEnd bool) (string, error) {
   searchExpression := `^https://.+/([a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})`
   if idAtEnd {
      searchExpression +=`$`
   } else {
      searchExpression += `.*$`
   }
   reGetID := regexp.MustCompile(searchExpression)
   // [...]
} 
```
This way, we use just one function for two needs.

Calls to `getUuidFromHref` will become `GetUuidFromHref(href, true)`, while calls to `GetUuidFromHref` will become `GetUuidFromHref(href, false)`

lvirbalas: I'm in favor of a common function as well to avoid duplication.

govcd/vapp.go
```diff
@@ -786,6 +798,329 @@ func (vapp *VApp) AddIsolatedNetwork(newIsolatedNetworkSettings *VappNetworkSett
 
 }
 
+// AddNetwork creates isolated or nat routed(connected to Org VDC network) network for vApp.
```
dataclouder: This function (and the related `Async` function) need a better name.
In both `AddNetwork` and `AddOrgNetwork` we are passing an `OrgVdcNetwork` as a parameter, but in one case we are assigning an existing network to the vApp, and in the other we are creating a vApp network. 
I think the name in this case should be `CreateVappNetwork` or `AddVappNetwork` to avoid confusion

govcd/vapp.go
```diff
@@ -814,7 +1149,57 @@ func validateNetworkConfigSettings(networkSettings *VappNetworkSettings) error {
...
+
+	vAppNetworkConfig, err := vapp.GetNetworkConfig()
+	if err != nil {
+		return nil, fmt.Errorf("error getting vApp networks: %#v", err)
+	}
+
+	return vAppNetworkConfig, nil
+}
+
+// RemoveNetworkAsync asyncronously removes any network (be it isolated or connected to an Org Network) from vApp
```
dataclouder:
```suggestion
// RemoveNetworkAsync asynchronously removes any network (be it isolated or connected to an Org Network) from vApp
```

govcd/vapp.go
```diff
@@ -786,6 +798,329 @@ func (vapp *VApp) AddIsolatedNetwork(newIsolatedNetworkSettings *VappNetworkSett
...
+
+	vAppNetworkConfig, err := vapp.GetNetworkConfig()
+	if err != nil {
+		return nil, fmt.Errorf("error getting vApp networks: %#v", err)
+	}
+
+	return vAppNetworkConfig, nil
+}
+
+// AddNetworkAsync creates asyncronously isolated or nat routed network for vApp. Returns Task or error
```
dataclouder:
```suggestion
// AddNetworkAsync creates asynchronously isolated or nat routed network for vApp. Returns Task or error
```

govcd/system.go
vbauzysvmware: good idea, change added.

govcd/vapp.go
vbauzysvmware: improved

govcd/vapp.go
vbauzysvmware: @Didainius what is your vote for naming?

govcd/vapp.go
vbauzysvmware: Thank you. updated

govcd/vapp.go
vbauzysvmware: Thank you. updated

govcd/vapp.go
Didainius: After a discussion we proved that NAT cannot be enabled without firewall being enabled.

govcd/vapp.go
Didainius: Alright. Lets live with it.

govcd/vapp.go
```diff
@@ -814,7 +1149,57 @@ func validateNetworkConfigSettings(networkSettings *VappNetworkSettings) error {
 	return nil
 }
 
+// RemoveNetwork removes any network (be it isolated or connected to an Org Network) from vApp
+// Returns pointer to types.NetworkConfigSection or error
+func (vapp *VApp) RemoveNetwork(id string) (*types.NetworkConfigSection, error) {
```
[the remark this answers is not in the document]
vbauzysvmware: renamed

CHANGELOG.md
```diff
@@ -11,11 +11,12 @@
 * Added methods `EdgeGateway.GetAllNsxvDhcpLeases()`, `EdgeGateway.GetNsxvActiveDhcpLeaseByMac()`
   `VM.WaitForDhcpIpByNicIndexes()`, `VM.GetParentVApp()`, `VM.GetParentVdc()`
   [#283](https://github.com/vmware/go-vcloud-director/pull/283)
+* Added methods `vapp.AddNetwork`, `vapp.AddNetworkAsync`, `vapp.AddOrgNetwork`, `vapp.AddOrgNetworkAsync`, `vapp.UpdateNetwork`, `vapp.UpdateNetworkAsync`, `vapp.UpdateOrgNetwork`, `vapp.UpdateOrgNetworkAsync`, `vapp.RemoveNetwork`, `vapp.RemoveNetworkAsync` and `GetUuidFromHref` [#289](https://github.com/vmware/go-vcloud-director/pull/290)
+* Deprecated function `vapp.RemoveIsolatedNetwork`, `vapp.AddRAWNetworkConfig` and `vapp.AddIsolatedNetwork`  [#289](https://github.com/vmware/go-vcloud-director/pull/290)
```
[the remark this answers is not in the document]

govcd/vapp.go
vbauzysvmware: renamed

PR #291, opened 2020-03-01
Change bool -> *bool in types.GuestCustomizationSection

Didainius:
In order to solve https://github.com/terraform-providers/terraform-provider-vcd/issues/303 and other problems it is better to have *bool instead of bool types (to distinguish between 0 and nil values).

This PR aims to solve just that and also deprecates some functions which are not technically correct.

Test suite passed on 9.5


govcd/vm.go
```diff
@@ -459,10 +462,10 @@ func (vm *VM) Customize(computername, script string, changeSid bool) (Task, erro
 		HREF:                vm.VM.HREF,
 		Type:                types.MimeGuestCustomizationSection,
 		Info:                "Specifies Guest OS Customization Settings",
-		Enabled:             true,
+		Enabled:             takeBoolPointer(true),
```
vbauzysvmware: Maybe it is worth it also change this harcoded value too?

govcd/vapp.go
```diff
@@ -384,10 +387,10 @@ func (vapp *VApp) Customize(computername, script string, changeSid bool) (Task,
 		HREF:                vapp.VApp.Children.VM[0].HREF,
 		Type:                types.MimeGuestCustomizationSection,
 		Info:                "Specifies Guest OS Customization Settings",
-		Enabled:             true,
+		Enabled:             takeBoolPointer(true),
```
vbauzysvmware: Maybe it is worth it also change this harcoded value too?

govcd/vm.go
Didainius: This function does not have a parameter for this field, while `changeSid` had, but was missed.

govcd/vapp.go
Didainius: This function does not have a parameter for this field, while `changeSid` had, but was missed.

govcd/vm.go
```diff
@@ -445,6 +445,9 @@ func (vm *VM) BlockWhileGuestCustomizationStatus(unwantedStatus string, timeOutA
 	}
 }
 
+// Customize function allows to set computername, apply customization script and enable or disble the changeSid option
```
lvirbalas: Maybe:
```suggestion
// Customize function allows to set ComputerName, apply customization script and enable or disable the changeSid option
```

Didainius: Fixed.

PR #292, opened 2020-03-04
Update Org VDC Networks

dataclouder:
Adds the ability of updating Org VDC networks in place. Until now, we could only remove and re-create them.

Added methods `orgVdcNetwork.Update`, `orgVdcNetwork.UpdateAsync`, and `OrgVdcNetwork.Rename`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
-## 2.6.0 (Unreleased)
+## 2.7.0 (Unreleased)
+
+* Added methods `OrgVdcNetwork.Update`, `OrgVdcNetwork.UpdateAsync`, and `OrgVdcNetwork.Rename`
+* Added methods `EdgeGateway.Update` and `EdgeGateway.UpdateAsync`
+
+## 2.6.0 (March 13, 2010)
 
 * Moved `VCDClient.supportedVersions` to `VCDClient.Client.supportedVersions` [#274](https://github.com/vmware/go-vcloud-director/pull/274)    
 * Added methods `VM.AddInternalDisk`, `VM.GetInternalDiskById`, `VM.DeleteInternalDisk`, `VM.UpdateInternalDisks` and `VM.UpdateInternalDisksAsync` [#272](https://github.com/vmware/go-vcloud-director/pull/272)
@@ -6,7 +11,6 @@
 * Improved functions to not expect XML namespaces provided in argument structure [#284](https://github.com/vmware/go-vcloud-director/pull/284)
 * Change `int` and `bool` fields from types.VAppTemplateLeaseSettings and VAppLeaseSettings into pointers
 * Added method `catalog.GetVappTemplateByHref`, and expose methods `vdc.GetEdgeGatewayByHref` and `vdc.GetEdgeGatewayRecordsType`
-* Added methods `orgVdcNetwork.Update`, `orgVdcNetwork.UpdateAsync`, and `OrgVdcNetwork.Rename`
 * Added methods `adminOrg.CreateOrgVdc`, `adminOrg.CreateOrgVdcAsync` and improved existing to support Flex VDC model. These new methods are dynamic as they change invocation behind the scenes based on vCD version [#285](https://github.com/vmware/go-vcloud-director/pull/285) 
 * Deprecated functions `adminOrg.CreateVdc` and `adminOrg.CreateVdcWait` [#285](https://github.com/vmware/go-vcloud-director/pull/285)
 * Added methods `EdgeGateway.GetAllNsxvDhcpLeases()`, `EdgeGateway.GetNsxvActiveDhcpLeaseByMac()`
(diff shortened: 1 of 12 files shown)


types/v56/types.go
```diff
@@ -212,14 +212,14 @@ type NetworkConfiguration struct {
...
 	Features                       *NetworkFeatures `xml:"Features,omitempty"`
-	GuestVlanAllowed               *bool            `xml:"GuestVlanAllowed,omitempty"`
 
 	// SubInterface and DistributedInterface are mutually exclusive
 	// When they are both nil, it means the "internal" interface (the default) will be used.
 	// When one of them is set, the corresponding interface will be used.
 	// They cannot be both set (we'll get an API error if we do).
 	SubInterface         *bool `xml:"SubInterface,omitempty"`
 	DistributedInterface *bool `xml:"DistributedInterface,omitempty"`
+	GuestVlanAllowed     *bool `xml:"GuestVlanAllowed,omitempty"`
```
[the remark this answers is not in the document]
dataclouder: I haven't come across any issues with order after this change.

govcd/edgegateway_test.go
```diff
@@ -743,3 +743,51 @@ func (vcd *TestVCD) TestListEdgeGateway(check *C) {
...
+	newName := "UpdatedEGWName"
+	newDescription := "Updated description"
+
+	edge.EdgeGateway.Name = newName
+	edge.EdgeGateway.Description = newDescription
+
+	err = edge.Update()
+	check.Assert(err, IsNil)
+
+	edge, err = vcd.vdc.GetEdgeGatewayById(saveEGW.ID, true)
```
vbauzysvmware: I think `update` should also update referenced Instance, not only send data.
So `edge.Update` -> would means that after `edge.EdgeGateway.Name` I would see new name. I don't need 
`edge, err = vcd.vdc.GetEdgeGatewayById(saveEGW.ID, true)`
Or return updated instance - usually we do that.

govcd/orgvdcnetwork.go
```diff
@@ -236,3 +236,111 @@ func (vdc *Vdc) FindEdgeGatewayNameByNetwork(networkName string) (string, error)
...
+	}
+
+	return vdc.GetEdgeGatewayByName(edgeGatewayName, false)
+}
+
+// UpdateAsync will change the contents of a network using the information in the
+// receiver data structure.
+func (orgVdcNet *OrgVDCNetwork) UpdateAsync() (Task, error) {
+	if orgVdcNet.OrgVDCNetwork.HREF == "" {
+		return Task{}, fmt.Errorf("cannot update network: HREF is empty")
```
vbauzysvmware: We should update all errors to be more specific-> "cannot update Org VDC network: HREF is empty"

govcd/orgvdcnetwork_test.go
```diff
@@ -151,6 +159,34 @@ func (vcd *TestVCD) testCreateOrgVdcNetworkRouted(check *C, ipSubnet string, sub
...
+	network.OrgVDCNetwork.Configuration.IPScopes.IPScope[0].IPRanges.IPRange[0].EndAddress = updatedEndAddress
+
+	err = network.Update()
+	check.Assert(err, IsNil)
+	AddToCleanupList(updatedNetworkName,
+		"network",
+		vcd.org.Org.Name+"|"+vcd.vdc.Vdc.Name,
+		"Test_CreateOrgVdcNetworkRouted")
+
+	network, err = vcd.vdc.GetOrgVdcNetworkById(networkId, true)
```
vbauzysvmware: Similar comment here regarding updating referenced instance during update.

types/v56/types.go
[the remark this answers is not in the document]
vbauzysvmware: I checked, at least we don't need to handle GuestVlanAllowed between versions. 9.1 already supports it.

CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
-## 2.6.0 (Unreleased)
+## 2.7.0 (Unreleased)
+
+* Added methods `OrgVdcNetwork.Update`, `OrgVdcNetwork.UpdateAsync`, and `OrgVdcNetwork.Rename`
```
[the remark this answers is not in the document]
dataclouder: done

govcd/edgegateway.go
```diff
@@ -1398,3 +1404,24 @@ func getNetworkNameAndTypeByVnicIndex(vNicIndex int, vnics *types.EdgeGatewayInt
 
 	return networkName, networkType, nil
 }
+
+func (egw *EdgeGateway) UpdateAsync() (Task, error) {
```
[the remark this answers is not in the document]
dataclouder: added

govcd/edgegateway_test.go
dataclouder: Added explicit refresh at the end of Update, as agreed

govcd/orgvdcnetwork.go
dataclouder: updated

govcd/orgvdcnetwork_test.go
dataclouder: addressed

dataclouder:
> LGTM, I would only like to ask to extend the PR description (and commit message) to include narrative what was not possible before and what the additions allow to do now more specifically (from the perspective of casual user who will be reading this PR description after clicking link in the changelog after referencing it with the next release).

Previously, we could not update networks: we had to remove and re-create them. That's what the Update functions are about

PR #299, opened 2020-04-09
bugfix for vapp.RemoveNetwork() and vapp.RemoveNetworkAsync() when removing multiple network sequentially

Didainius:
This PR attempts to fix `vapp.RemoveNetworkAsync()` and `vapp.RemoveNetwork()` functions by using http DELETE call on endpoint `https://xxx/api/network/ef005324-a62b-4041-ae8c-c7bec64fb348` instead of re-applying whole vApp network configuration with a particular network excluded.
It has caused situation where re-applying whole vApp network config gave error because that config was invalid.

One particular place which got impacted by that is Terraform provider with many sequential removals - https://github.com/terraform-providers/terraform-provider-vcd/issues/473

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ NOTES:
 BUGS FIXED:
 * Fix issue in Queries with vCD 10 version, which do not return network pool or provider VDC[#293](https://github.com/vmware/go-vcloud-director/pull/293)
 * Session timeout for media, catalog item upload  [#294](https://github.com/vmware/go-vcloud-director/pull/294)
+* Fix `vapp.RemoveNetwork`, `vapp.RemoveNetworkAsync` to use `DELETE` API call instead of update
+  which can apply incorrect remaining vApp network configurations [#299](https://github.com/vmware/go-vcloud-director/pull/299)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ NOTES:
 BUGS FIXED:
 * Fix issue in Queries with vCD 10 version, which do not return network pool or provider VDC[#293](https://github.com/vmware/go-vcloud-director/pull/293)
 * Session timeout for media, catalog item upload  [#294](https://github.com/vmware/go-vcloud-director/pull/294)
+* Fix `vapp.RemoveNetwork`, `vapp.RemoveNetworkAsync` to use `DELETE` API call instead of update
```
lvirbalas, line 14: What about a comment that this is actual only for 10.1?

Didainius: It causes problems not only in 10.1. At least 10.0 and 9.5 also has problems with Terraform's binary test `vcd.TestAccVcdVAppVmMultiNIC.tf` reendering error similar to https://github.com/terraform-providers/terraform-provider-vcd/issues/473

PR #305, opened 2020-05-11
Add command-line flags to replace env variables

dataclouder:
Same as the equivalent PR in terraform-provider-vcd, this change enables command line flags that
can be used instead of environment variables.

Use "go test -tags functional -vcd-help" for a list of flags

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -332,8 +332,7 @@ While running tests, the following environment variables can be used:
 * `VCD_TOKEN` : specifies the authorization token to use instead of username/password
    (Use `./scripts/get_token.sh` to retrieve one)
 
-When both the environment variable and the command line option are possible, the environment variable gets evaluated first: is it is set,
-also the flag becomes true.
+When both the environment variable and the command line option are possible, the environment variable gets evaluated first.
 
 # Final Words
 Be careful about using our tests as these tests run on a real vcd. If you don't have 1 gb of ram and 2 vcpus available then you should not be running tests that deploy your vm/change memory and cpu. However everything created will be removed at the end of testing.
--- a/TESTING.md
+++ b/TESTING.md
@@ -309,24 +309,26 @@ func (vcd *TestVCD) Test_ComposeVApp(check *checks.C) {
 }
 ```
 
-# Environment variables
+# Environment variables and corresponding flags
 
 While running tests, the following environment variables can be used:
 
 * `GOVCD_CONFIG=/path/file`: sets an alternative configuration file for the tests.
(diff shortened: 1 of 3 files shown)


TESTING.md
```diff
@@ -309,24 +309,26 @@ func (vcd *TestVCD) Test_ComposeVApp(check *checks.C) {
 }
 ```
 
-# Environment variables
+# Environment variables and corresponding flags
 
 While running tests, the following environment variables can be used:
```
lvirbalas, line 314: We should state which is higher priority too - env vars or flags.

govcd/api_vcd_test.go
```diff
@@ -1426,3 +1446,39 @@ func (vcd *TestVCD) Test_NewRequestWitNotEncodedParamsWithApiVersion(check *C) {
 
 	fmt.Printf("Test: %s run with api Version: %s\n", check.TestName(), apiVersion)
 }
+
+func setBoolFlag(variable *bool, name, envVar, help string) {
+	if envVar != "" && os.Getenv(envVar) != "" {
+		*variable = true
+		return
+	}
+	flag.BoolVar(variable, name, *variable, help)
```
lvirbalas, line 1455: This is actually tricky to read, because `variable` is used twice in the same line. And it's not clear how can it have the default value at this early point in time too. To clarify, please add a comment about this line and also about this function.

dataclouder: The first `variable` is a pointer, which is used to pass to `flag.BoolVar`. The second one is its value, which is whatever was set when the variable was declared.
They are Booleans, thus the default value is `false`, which is good enough for `flag.BoolVar`

TESTING.md
dataclouder: Environment variables have precedence.
With this request, you have actually found a bug. Going to fix it

TESTING.md
```diff
@@ -309,27 +309,32 @@ func (vcd *TestVCD) Test_ComposeVApp(check *checks.C) {
...
+* `GOVCD_IGNORE_CLEANUP_FILE` (`-vcd-ignore-cleanup-file`): Ignore the cleanup file if it is left behind after a test failure.
     This could be useful after running a single test, when we need to check how the test behaves with the resource still
     in place.
+* `GOVCD_SHOW_REQ` (`-vcd-show-request`): shows the API request on standard output
+* `GOVCD_SHOW_RESP` (`-vcd-show-response`): shows the API response on standard output
 * `VCD_TOKEN` : specifies the authorization token to use instead of username/password
    (Use `./scripts/get_token.sh` to retrieve one)
 
+When both the environment variable and the command line option are possible, the environment variable gets evaluated first: is it is set,
+also the flag becomes true.
```
lvirbalas, line 336: Distorted sentence.

PR #314, opened 2020-05-26
Add support for LDAP and SAML groups

Didainius:
This PR adds support for Org group management by introducing the following methods:
* CreateGroup
* GetGroupByHref
* GetGroupById
* GetGroupByName
* GetGroupByNameOrId
* Delete
* Update
* NewGroup

Additionally it adds two new methods for `adminOrg` - `LdapConfigure` and `LdapDisable` to aid automated testing.

## Testing 
Two tests `test_GroupCRUD` and `test_GroupFinderGetGenericEntity` are added. They are both are run from main test `Test_LDAP` which sets up LDAP configuration in vCD - spawns Photon OS VM and uses testing LDAP container https://github.com/rroemhild/docker-test-openldap inside it (using customization scripts.

**Note.** These tests will not work when run separately because LDAP configuration must be present.
**Note.** External network with valid IPs and DNS servers must be configured because guest will not be able to pull docker image otherwise.

`make test` passed on 9.5 and 10.1. Also on 10.1 using SAML authentication (and enabling SAML group related tests in `test_GroupCRUD`.


govcd/adminorg_administration.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+	"github.com/vmware/go-vcloud-director/v2/util"
+)
+
+// LdapConfigure allows to configure LDAP mode in use by the Org
+func (adminOrg *AdminOrg) LdapConfigure(settings *types.OrgLdapSettingsType) error {
```
vbauzysvmware, line 16: I haven't found test for it. Also we usually we return entity which we created/updated

govcd/adminorg_administration.go
```diff
@@ -0,0 +1,37 @@
...
+		"error updating LDAP settings: %s", settings, nil)
+	if err != nil {
+		return fmt.Errorf("error updating LDAP mode for Org name '%s': %s", adminOrg.AdminOrg.Name, err)
+	}
+
+	return nil
+}
+
+// LdapDisable wraps LdapConfigure to disable LDAP configuration for org
+func (adminOrg *AdminOrg) LdapDisable() error {
```
vbauzysvmware, line 35: Missing test

govcd/adminorg_ldap_test.go
```diff
@@ -0,0 +1,312 @@
...
+	vcd.test_GroupCRUD(check)
+	vcd.test_GroupFinderGetGenericEntity(check)
+
+}
+
+// configureLdap creates direct network, spawns Photon OS VM with LDAP server and configures vCD to
+// use LDAP server
+func (vcd *TestVCD) configureLdap(check *C) (string, string, string) {
+	if vcd.skipAdminTests {
+		check.Skip(fmt.Sprintf(TestRequiresSysAdminPrivileges, check.TestName()))
```
vbauzysvmware, line 41: Could we do check to if external network is configured right and skip test otherwise. This would save testing time(spawning photon with LDAP).

govcd/adminorg_ldap_test.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	// Create direct network to expose LDAP server on external network
+	directNetworkName := createDirectNetwork(vcd, check)
+
+	// Launch LDAP server on external network
+	ldapHostIp, vappName, vmName := createLdapServer(vcd, check, directNetworkName)
+
+	// Configure vCD to use new LDAP server
+	orgConfigureLdap(vcd, check, ldapHostIp)
```
vbauzysvmware, line 51:
```suggestion
	configureLdapForOrg(vcd, check, ldapHostIp)
```

govcd/adminorg_ldap_test.go
```diff
@@ -0,0 +1,312 @@
...
+	check.Assert(vdc, NotNil)
+
+	// Find catalog and catalog item
+	catalog, err := org.GetCatalogByName(vcd.config.VCD.Catalog.Name, false)
+	check.Assert(err, IsNil)
+	check.Assert(catalog, NotNil)
+	catalogItem, err := catalog.GetCatalogItemByName(vcd.config.VCD.Catalog.CatalogItem, false)
+	check.Assert(err, IsNil)
+	// Skip the test if catalog item is not Photon OS
+	if !isItemPhotonOs(*catalogItem) {
```
vbauzysvmware, line 189: This check should happen in front of all testing, as major thing to have to run the test.

govcd/adminorg_ldap_test.go
Didainius: This is not really possible because it may have any configuration and I have no obvious way to cover all cases. It could even not have DNS set, but the default ones in guest could work. Besides it would mask the test run then and one could think - "I have run all the tests" while in fact some where skipped.

govcd/adminorg_ldap_test.go
Didainius: But that is exactly the same as in `createAngGetResourcesForVmCreation`. I think readability wise I will pollute the environment if pull it out.

vbauzysvmware: Let's @dataclouder  decide. If we need less running test with caveats.

dataclouder: If we need to run this check, we should either bail out as early as possible, or reorganize the auxiliary functions to return an error, allowing the callers to clean up in case of problems.
As it is now, an error (or a skip) may leave the system in an unclean state

govcd/adminorg_administration.go
Didainius: added

govcd/adminorg_administration.go
Didainius: added

govcd/adminorg_ldap_test.go
Didainius: changed

govcd/adminorg_ldap_test.go
Didainius: let it be. moved out

Makefile
```diff
@@ -7,7 +7,7 @@ default: fmtcheck vet static build
 # test runs the test suite and vets the code
 test: testunit
 	@echo "==> Running Functional Tests"
-	cd govcd && go test -tags "functional" -timeout=200m -check.vv .
+	cd govcd && go test -tags "functional" -timeout=300m -check.vv .
```
lvirbalas, line 10: Please tell more why are you adding additional one hundred minutes to the timeout :)

Didainius: It is not related to this exact group PR. I did hit timeouts in general for full suite runs in slower envs that is why I increased the "default".

PR #316, opened 2020-06-09
New functionality for handling vApp network NAT 

vbauzysvmware:
Add function needed for implementing vApp network NAT rules resource:
* vapp.UpdateNetworkNatRules

This PR is derived from https://github.com/vmware/go-vcloud-director/pull/308 so shows that PR code too.

PR #308, opened 2020-05-21
New functionality for handlig vapp network firewall

vbauzysvmware:
* Added methods needed for implementing vapp network firewall rules resource
* `vapp.UpdateNetworkFirewallRules`
* `vapp.UpdateNetworkFirewallRulesAsync`
* `vapp.GetVappNetworkById`
* `vapp.GetVappNetworkByName`
* `vapp.GetVappNetworkByNameOrId`

PR #318, opened 2020-06-18
New functionality for handling vApp network static routing

vbauzysvmware:
Add function needed for implementing vApp network static routing rules resource:
   
    vapp.UpdateNetworkStaticRouting
    vapp.UpdateNetworkStaticRoutingAsync
    vapp.RemoveAllNetworkStaticRoutes

This PR is derived from #308 and #316 so shows that PR code too.


govcd/vapp_network.go
```diff
@@ -223,3 +223,74 @@ func (vapp *VApp) RemoveAllNetworkFirewallRules(networkId string) error {
...
+	}
+	err = task.WaitTaskCompletion()
+	if err != nil {
+		return nil, fmt.Errorf("%s", combinedTaskErrorMessage(task.Task, err))
+	}
+
+	return vapp.GetVappNetworkById(networkId, false)
+}
+
+// UpdateNetworkNatRulesAsync asynchronously updates vApp network static routes.
```
Didainius, line 242: Incorrect comment. Probably a copy/paste thing :) 
```suggestion
// UpdateNetworkStaticRoutingAsync asynchronously updates vApp network static routes.
```

govcd/vapp_network.go
```diff
@@ -223,3 +223,74 @@ func (vapp *VApp) RemoveAllNetworkFirewallRules(networkId string) error {
...
+	// But `GET /network/{id}` can return org VDC network or vApp network.
+	apiEndpoint := vapp.client.VCDHREF
+	apiEndpoint.Path += "/network/" + uuid
+
+	return vapp.client.ExecuteTaskRequest(apiEndpoint.String(), http.MethodPut,
+		types.MimeVappNetwork, "error updating vApp Network static routes: %s", networkToUpdate)
+}
+
+// Allows to identify if given network config is a vApp network and not a vApp Org network
+func isVappNetwork(networkConfig *types.VAppNetwork) bool {
```
Didainius, line 274: One consideration  - I think we have similar function in Terraform codebase. Wondering if they are 100% the same and if they are - maybe it is worth to make this function public and reuse in Terraform to avoid some flops of having slightly different checks in different places.

cc @dataclouder , @lvirbalas

govcd/vapp_network.go
```diff
@@ -223,3 +223,74 @@ func (vapp *VApp) RemoveAllNetworkFirewallRules(networkId string) error {
...
+	// here we use `PUT /network/{id}` which allow to change vApp network.
+	// But `GET /network/{id}` can return org VDC network or vApp network.
+	apiEndpoint := vapp.client.VCDHREF
+	apiEndpoint.Path += "/network/" + uuid
+
+	return vapp.client.ExecuteTaskRequest(apiEndpoint.String(), http.MethodPut,
+		types.MimeVappNetwork, "error updating vApp Network static routes: %s", networkToUpdate)
+}
+
+// Allows to identify if given network config is a vApp network and not a vApp Org network
```
Didainius, line 273:
```suggestion
// isVappNetwork allows to identify if given network config is a vApp network and not a vApp Org network
```

govcd/vapp_network_test.go
```diff
@@ -316,3 +316,65 @@ func (vcd *TestVCD) Test_UpdateNetworkNatRules(check *C) {
...
+			IPAddressAllocationMode: types.IPAllocationModePool,
+			Network:                 "Test_UpdateNetworkStaticRoutes",
+			NetworkConnectionIndex:  0,
+		})
+
+	uuid, err := GetUuidFromHref(networkFound.Link.HREF, false)
+	check.Assert(err, IsNil)
+
+	result, err := vapp.UpdateNetworkStaticRouting(uuid, []*types.StaticRoute{&types.StaticRoute{Name: "test1",
+		Network: "192.168.2.0/24", NextHopIP: "192.168.2.15"}, &types.StaticRoute{Name: "test2",
```
Didainius, line 346: From my experiments at least NextHopIp must be related to network attached to avoid error:
image

This test does pass now on one env I tested, but flagging just to be sure that these hard coded values don't break tests on envs with different IPs.

govcd/vapp_network.go
vbauzysvmware: fixed

govcd/vapp_network.go
vbauzysvmware: made public.

govcd/vapp_network.go
vbauzysvmware: added

govcd/vapp_network_test.go
vbauzysvmware: Test now create own network, so should be fine.

govcd/vapp_network.go
```diff
@@ -175,9 +176,13 @@ func (vapp *VApp) UpdateNetworkNatRulesAsync(networkId string, natRules []*types
 	}
 	networkToUpdate.Xmlns = types.XMLNamespaceVCloud
 
-	if networkToUpdate.Configuration.Features.NatService == nil {
+	if networkToUpdate.Configuration.Features.NatService == nil && networkToUpdate.Configuration.Features.FirewallService == nil {
 		return Task{}, fmt.Errorf("provided network isn't connected to org network or isn't fenced")
```
lvirbalas, line 180: How do you deduce from NAT and FW services being nil that it is "network isn't connected to org network or isn't fenced"? This asks for a comment in the code.

govcd/vapp_network.go
```diff
@@ -175,9 +176,13 @@ func (vapp *VApp) UpdateNetworkNatRulesAsync(networkId string, natRules []*types
...
 	}
+	if networkToUpdate.Configuration.Features.NatService == nil {
+		networkToUpdate.Configuration.Features.NatService = &types.NatService{}
```
lvirbalas, line 183: Is it possible for the FirewallService to be nil at this point, do we need a similar check for it? If not, this asks for a comment in the code about why.

vbauzysvmware: this part already changed, please check latest changes.

govcd/vapp_network.go
vbauzysvmware: don't know what to write.. will try

vbauzysvmware: added

PR #326, opened 2020-07-20
Adjust Test_LBAppRule to work with VCD 10.2

Didainius:
VCD 10.2 throws a bit different error when invalid application rule is being submitted and current test fails on it. This PR relieves invalid application rule check to fit older versions as well as 10.2.


Didainius:
> Looks trivial, but Travis is failing :)

Yep. This was one of Travis problems. Rebuild helped it.

PR #327, opened 2020-07-21
Update VDC dynamic func to handle API version 35.0

dataclouder:
Update recognized versions to handle 35.0 and (provisionally) 10.2.
Add a wrapper function to handle VDC creation when then highest version is not handled yet.

### DIFF
--- a/govcd/adminvdc.go
+++ b/govcd/adminvdc.go
@@ -56,19 +56,19 @@ var vdcVersionedFuncsV97 = vdcVersionedFuncs{
 var vdcVersionedFuncsByVcdVersion = map[string]vdcVersionedFuncs{
 	"vdc9.5":  vdcVersionedFuncsV95,
 	"vdc9.7":  vdcVersionedFuncsV97,
-	"vdc10.0": vdcVersionedFuncsV97,
-	"vdc10.1": vdcVersionedFuncsV97,
+
+	// If we add a new function to this list, we also need to update the "default" entry
+	"default": vdcVersionedFuncsV97,
 }
 
 // getVdcVersionedFuncsByVdcVersion is a wrapper function that retrieves the requested versioned VDC function
-// When the wanted version does not exist in the map, it returns the latest available one.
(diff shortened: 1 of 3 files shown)


govcd/adminvdc.go
```diff
@@ -52,15 +52,26 @@ var vdcVersionedFuncsV97 = vdcVersionedFuncs{
...
 	"vdc9.5":  vdcVersionedFuncsV95,
 	"vdc9.7":  vdcVersionedFuncsV97,
 	"vdc10.0": vdcVersionedFuncsV97,
 	"vdc10.1": vdcVersionedFuncsV97,
 }
 
+// getVdcVersionedFuncsByVdcVersion is a wrapper function that retrieves the requested versioned VDC function
+// When the wanted version does not exist in the map, it returns the latest available one.
+func getVdcVersionedFuncsByVdcVersion(version string) vdcVersionedFuncs {
+	highest := vdcVersionedFuncsV97
```
lvirbalas, line 66: Hm. Looks like we shouldn't be hard-coding this. What about choosing the last entry in the `vdcVersionedFuncsByVcdVersion` map instead? That should give you the highest version automatically.

dataclouder: A map doesn't have a **last** entry.
We need to change the `highest` when we create a new VDC function.  I can add a comment about that in `vdcVersionedFuncsByVcdVersion`

vbauzysvmware: one possible variation is to add `highest/default` in the map and fetch it when specific fetch fails.

dataclouder: Updated with default entry.

lvirbalas: > 	// If we add a new function to this list, we also need to update the "default" entry
	"default": vdcVersionedFuncsV97,

What about mentioning in the comment that this is the highest version for clarity?

dataclouder: added

PR #332, opened 2020-08-26
Introduce NSX-T support for Org Vdc

Didainius:
This PR lays some groundwork to introduce NSX-T support in testing. Org Vdc API calls are still using legacy API in UI and it does work properly with API v32.0. The main work here is to add some tests and configuration capability to set NSX-T related resources (network pool, provider vdc).


dataclouder (approved):
LGTM. Tests were successful with 10.1 and 10.2

lvirbalas (approved):
LGTM :)

govcd/adminvdc_nsxt_test.go
```diff
@@ -0,0 +1,140 @@
+// +build org functional nsxt ALL
```
vbauzysvmware, line 1: remove or uncomment ?

vbauzysvmware (approved):
Does it mean that tests decides to run or not only by configuration exist or not - is there isn't why to check if VCD backed by NSXT or not?

govcd/adminvdc_nsxt_test.go
Didainius: These are build tags for test. They have to be commented by convention.

Didainius:
> Does it mean that tests decides to run or not only by configuration exist or not - is there isn't why to check if VCD backed by NSXT or not?

Well yes. Configuration does hint if we are to do NSX-T related tests or not. Environments can have NSX-T and NSX-V backed VCs. Picking something by default could be risky.

vbauzysvmware:
> > Does it mean that tests decides to run or not only by configuration exist or not - is there isn't why to check if VCD backed by NSXT or not?
> 
> Well yes. Configuration does hint if we are to do NSX-T related tests or not. Environments can have NSX-T and NSX-V backed VCs. Picking something by default could be risky.

I was picking the case were I can have configuration but env doens't have NSX-T.

Didainius:
> I was picking the case were I can have configuration but env doens't have NSX-T.

Well. It will fail very quickly if it doesn't exist.

Didainius:
> > I was picking the case were I can have configuration but env doens't have NSX-T.
> 
> Well. It will fail very quickly if it doesn't exist.

We agreed that it will fail for now. We might improve it as we follow NSX-T implementation path for other resources.

PR #336, opened 2020-10-02
Add HTTP User-Agent header to all requests in SDK

Didainius:
This PR adds default HTTP `User-Agent` header with value `go-vcloud-director`. It also adds the capability to customize (or disable) this value by using `WithHttpUserAgent()` configuration option for `NewVCDClient` function.

**Note.** Testing to ensure that all outgoing requests contain HTTP header is not trivial therefore this library extends `util` package with an additional flag `PanicEmptyUserAgent` which would  cause panic in `ProcessRequestOutput` function which is hopefully logging gateway to all current and future HTTP requests. `func (vcd *TestVCD) SetUpSuite(check *C)` is configured to include verification of it by default.

Tests did pass which means (as per above note) that all logged HTTP calls have the header set.


lvirbalas (approved):
LGTM, thank you!

vbauzysvmware (approved):
LGTM. Have you already tested with terraform?

Didainius:
> LGTM. Have you already tested with terraform?

Yep. https://github.com/vmware/terraform-provider-vcd/pull/566

dataclouder (approved):

PR #339, opened 2020-10-25
VCD version

dataclouder:
Add functions to retrieve and compare VCD version
`client.GetVcdVersion`
`client.GetVcdShortVersion`
`client.GetVcdFullVersion`
`client.VersionEqualOrGreater`

This Code does not replace the functions used to compare API versions, which continue having its use.
The main uses of these functions are:

* Accessing the VCD version through API, and displaying it in logs, test output, diagnostics.
* Comparing the current version with a given (precise) version, in cases when the API version proves to be too broad (e.g. when a bug fix is introduced in a minor or revision version, and we want to change behavior based on such fix)


vbauzysvmware (approved):
LGTM

Didainius (approved):

lvirbalas (approved):
LGTM! Please add a sentence to PR description and merge message how what we had before was not enough.

dataclouder:
> LGTM! Please add a sentence to PR description and merge message how what we had before was not enough.

Added

PR #340, opened 2020-10-26
Add functions to allow to update Vdc storage profile

vbauzysvmware:
This new function will allow to update VDC storage profile.

* adminVdc.UpdateStorageProfile
* fix incorrect types.vdcStorageProfile usage. Create Vdc, update Storage profile and get Storage profile has different types.


types/v56/types.go
```diff
@@ -445,49 +445,96 @@ type AdminVdc struct {
...
+type VdcStorageProfile struct {
+	Xmlns                     string                         `xml:"xmlns,attr"`
+	Name                      string                         `xml:"name,attr"`
+	Enabled                   bool                           `xml:"Enabled,omitempty"`
+	Units                     string                         `xml:"Units"`
+	Limit                     int64                          `xml:"Limit"`
+	Default                   bool                           `xml:"Default"`
+	IopsSettings              *VdcStorageProfileIopsSettings `xml:"IopsSettingsint64"`
+	StorageUsedMB             int64                          `xml:"StorageUsedMB"`
+	IopsAllocated             int64                          `xml:"IopsAllocated"`
```
lvirbalas, line 476: Hard to understand the diff in GitHub, but do I get right, that in reality we have added these three new fields to the old `VdcStorageProfile` struct?

types/v56/types.go
```diff
@@ -445,49 +445,96 @@ type AdminVdc struct {
...
+	Xmlns                    string                     `xml:"xmlns,attr"`
+	Name                     string                     `xml:"name,attr"`
+	Description              string                     `xml:"Description,omitempty"`
+	AllocationModel          string                     `xml:"AllocationModel"` // Flex supported from 32.0
+	ComputeCapacity          []*ComputeCapacity         `xml:"ComputeCapacity"`
+	NicQuota                 int                        `xml:"NicQuota,omitempty"`
+	NetworkQuota             int                        `xml:"NetworkQuota,omitempty"`
+	VmQuota                  int                        `xml:"VmQuota,omitempty"`
+	IsEnabled                bool                       `xml:"IsEnabled,omitempty"`
+	VdcStorageProfile        []*CreateVdcStorageProfile `xml:"VdcStorageProfile"`
```
lvirbalas, line 526: This is getting confusing. Please add a comment why "CreateVdcStorageProfile" in "VdcStorageProfile".

lvirbalas (changes requested):

types/v56/types.go
vbauzysvmware: it was hard to grasp for me, until I figured out that API uses 3 types or VDC storage profiles which are almost incidental. Some of them need different XML root name. Every type has links to API description type. In general CreateType is used when creating Org VDC - which has only a few fields, VdcStorageProfile is used for GET and AdminVdcStorageProfile for update.

types/v56/types.go
vbauzysvmware: Answer in below comment

types/v56/types.go
lvirbalas: Please add the explanation as comment in code.

vbauzysvmware: Where too? all this type are referenced in different locations and every type already has explanation for what it used and reference to API.

govcd/adminvdc_test.go
```diff
@@ -219,3 +219,48 @@ func (vcd *TestVCD) Test_UpdateVdcFlex(check *C) {
 	check.Assert(*updatedVdc.AdminVdc.IsElastic, Equals, true)
 	check.Assert(*updatedVdc.AdminVdc.IncludeMemoryOverhead, Equals, false)
 }
+
+// Tests VDC storage profile update
+func (vcd *TestVCD) Test_UpdateStorageProfile(check *C) {
```
Didainius, line 224: Name is misleading. Maybe something similar to:
```suggestion
func (vcd *TestVCD) Test_VDCUpdateStorageProfile(check *C) {
```

types/v56/types.go
Didainius: Could we possible call it `AttachVdcStorageProfile` or something similar. The `CreateVdcStorageProfile` sounds as if it is going to create a storage profile but in fact it is not (storage profiles are defined in pVdcs)

govcd/adminvdc_test.go
```diff
@@ -219,3 +219,48 @@ func (vcd *TestVCD) Test_UpdateVdcFlex(check *C) {
...
+	check.Assert(err, IsNil)
+	check.Assert(storageProfileId, NotNil)
+
+	updatedVdc, err := adminVdc.UpdateStorageProfile(storageProfileId, &types.AdminVdcStorageProfile{
+		Name:                      foundStorageProfile.ProviderVdcStorageProfile.Name,
+		Default:                   true,
+		Limit:                     9081,
+		Enabled:                   takeBoolPointer(true),
+		Units:                     "MB",
+		IopsSettings:              nil,
```
Didainius, line 251: Can we test IOPS settings?

govcd/adminvdc.go
```diff
@@ -400,3 +400,24 @@ func (vdc *AdminVdc) GetVappList() []*types.ResourceReference {
...
+	queryUrl := vdc.client.VCDHREF
+	queryUrl.Path += "/admin/vdcStorageProfile/" + storageProfileId
+
+	storageProfile.Xmlns = types.XMLNamespaceVCloud
+	updateAdminVdcStorageProfile := &types.AdminVdcStorageProfile{}
+
+	_, err := vdc.client.ExecuteRequest(queryUrl.String(), http.MethodPut,
+		types.MimeStorageProfile, "error updating VDC storage profile: %s", storageProfile, updateAdminVdcStorageProfile)
+	if err != nil {
+		return nil, fmt.Errorf("cannot update VDC storage profil, error: %s", err)
```
Didainius, line 419:
```suggestion
		return nil, fmt.Errorf("cannot update VDC storage profile, error: %s", err)
```

Didainius (changes requested):
Still needs a master merge.

types/v56/types.go
vbauzysvmware: Well any more better namings? I understand your reasoning. I tried to match to createOrgVdc logic. If other agrees and there isn't better ides - I will change. @datacharmer @lvirbalas

govcd/adminvdc_test.go
vbauzysvmware: changed

govcd/adminvdc_test.go
vbauzysvmware: I could try, but it wasn't my intention to support that - I haven't analyzed or try to figure out what to add and what to expect.

govcd/adminvdc.go
vbauzysvmware: thank you. Changed

types/v56/types.go
dataclouder: The current naming is not acceptable because is a double breaking change.
We are renaming an existing structure (`VdcStorageProfile`) to `CreateVdcStorageProfile`, and what is worse, we are giving the old name `VdcStorageProfile` to a new structure.

vbauzysvmware: Well previously existing type usage wasn't correct - it was bug. I made changes to have least impact as possible. Func interfaces haven't changed. Only one thing changes is that `OrgVdc` internal Type `VdcStorageProfile` changed to `CreateVdcStorageProfile` .

dataclouder: The right procedure in this case would be to deprecate the existing structure and creating a new one. As it is we are introducing two levels of confusion with this change.

vbauzysvmware: Well result will be I have to introduce new functions for createOrgVdc, createOrgVdcType then +4 VdcStorageProfile types and new functions for getStorageProfilHref and everything will be with "silly namings" and deprecate 2 functions. Overall this minimum I recalled in my head. @lvirbalas  what to do?

lvirbalas: After discussion we renamed to `VdcStorageConfiguration` to be inline with `VdcConfiguration` - structs that are used to **create** the VDC.

@dataclouder do you have an alternative proposal without deprecations? One alternative I can think of is actually leaving old name as is, but adding a suffix or prefix "Read" to the new structs:
`VdcStorageProfileRead` and `AdminVdcStorageProfileRead`.

Either way, we need to be consistent with these kind of prefixes going forward.

vbauzysvmware: if we do that, we break get function definitions. So more breakage

vbauzysvmware:
> Previous comment was a lot more clear. Now it's confusing. Please clarify the comment whether this struct is used for creating VDC.

added

types/v56/types.go
dataclouder: >leaving old name as is, but adding a suffix or prefix "Read" to the new structs:
VdcStorageProfileRead and AdminVdcStorageProfileRead.

That I could live with. And it would make it possible to deprecate the old one.

lvirbalas: Let's go with `VdcStorageProfileConfiguration`, just last ask to add breaking changes section to changelog.

lvirbalas (approved):
LGTM.

dataclouder (approved):

govcd/adminvdc_test.go
Didainius: Ok, lets skip it for now if it is not needed

Didainius (approved):

PR #341, opened 2020-11-09
Enable usage of bearer tokens

dataclouder:
The SDK now accepts both the old (deprecated) authorization token and the bearer token.
The usage is transparent: it doesn't require new options or configuration. If we enter a bearer token instead of an authorization token, it gets used.

How to test:
1. run `./scripts/get_token.sh`
2. Copy the bearer token
3. Paste the token into the configuration file
4. Set the environment variable `GOVCD_LOG_PASSWORDS=1`
5. Run a quick test (such as `Test_GetVcdVersion`)
6. Check the logs, making sure that the `Authorization` header now contains the bearer token and that the `X-Vmware-Vcloud-Token-Type` header contains "Bearer".
7. Change the token to the authorization token (also from step 1)
8. Run steps 4-6 (the token used is the short one)
9. Remove the token from the configuration file
10. Run steps 4-5
11. Check the logs, making sure that `Authorization` header contains the bearer token.

This PR also adds a transparent fallback connection using `cloudapi/1.0.0/sessions` when the call to `/api/session` has been disabled. No action is needed to activate it. When the fallback call happens, we'll find trace information in the logs:
```
2020/11/10 15:35:16 [TRACE] Connecting to VCD using cloudapi
2020/11/10 15:35:16 [TRACE] URL https://bos1-vcloud-static-xxx-xxx.eng.vmware.com/cloudapi/1.0.0/sessions/provider
```

To test the fallback functionality, disable access to `api/sessions` following the instructions in [this page](https://docs.vmware.com/en/VMware-Cloud-Director/10.0/com.vmware.vcloud.install.doc/GUID-84390C8F-E8C5-4137-A1A5-53EC27FE0024.html) and then run any test without token.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Added functions to retrieve and use VCD version `client.GetVcdVersion`, `client.GetVcdShortVersion`, `client.GetVcdFullVersion`, `client.VersionEqualOrGreater` [#339](https://github.com/vmware/go-vcloud-director/pull/339)
 * Added methods `VM.UpdateStorageProfile`, `VM.UpdateStorageProfileAsync` [#338](https://github.com/vmware/go-vcloud-director/pull/338)
+* Added support for bearer tokens [#341](https://github.com/vmware/go-vcloud-director/pull/341)
 
 ## 2.9.0 (October 15, 2020)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 * Added functions to retrieve and use VCD version `client.GetVcdVersion`, `client.GetVcdShortVersion`, `client.GetVcdFullVersion`, `client.VersionEqualOrGreater` [#339](https://github.com/vmware/go-vcloud-director/pull/339)
 * Added methods `VM.UpdateStorageProfile`, `VM.UpdateStorageProfileAsync` [#338](https://github.com/vmware/go-vcloud-director/pull/338)
 * Added methods `adminVdc.UpdateStorageProfile` [#340](https://github.com/vmware/go-vcloud-director/pull/340)
-* Added support for bearer tokens [#341](https://github.com/vmware/go-vcloud-director/pull/341)
+* Added transparent support for bearer tokens [#341](https://github.com/vmware/go-vcloud-director/pull/341)
 * Added transparent connection using `cloudapi/1.0.0/sessions` when access through `api/sessions` is disabled
 
 BREAKING CHANGES:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@
 BREAKING CHANGES:
 
 * type.VdcConfiguration (used for creation) changed the type for storage profile from `[]*VdcStorageProfile` to `[]*VdcStorageProfileConfiguration`
->>>>>>> master
 
 ## 2.9.0 (October 15, 2020)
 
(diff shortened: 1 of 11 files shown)


util/logging.go
```diff
@@ -79,7 +79,7 @@ var (
 	LogHttpResponse bool = true
 
 	// List of tags to be excluded from logging
-	skipTags = []string{"SupportedVersions", "ovf:License"}
+	skipTags = []string{"ovf:License"}
```
dataclouder, line 82: the SupportedVersions call in recent releases has become short and bearable, while up to 9.5 it was 600+ lines.

vbauzysvmware (approved):
LGTM. 
Maybe we already can update sample configuration and state that
```
    # The vCD address, in the format https://vCD_IP/api
    # or https://vCD_host_name/api
    url: https://11.111.1.111/api
```
should be 

```
    url: https://11.111.1.111/openapi
```

dataclouder:
> LGTM.
> Maybe we already can update sample configuration and state that
> 
> ```
>     # The vCD address, in the format https://vCD_IP/api
>     # or https://vCD_host_name/api
>     url: https://11.111.1.111/api
> ```
> 
> should be
> 
> ```
>     url: https://11.111.1.111/openapi
> ```

Not yet. 
This may work for the initial connection, and we would get a valid token, but there are three problems:
* It won't work in 9.7
* the connection call is different for provider and tenant (see `get_token.sh` for an example)
* many calls in our SDK still require the `/api` format.

Didainius (approved):
My testing worked out well.

govcd/api_vcd.go
```diff
@@ -68,14 +94,28 @@ func (vcdCli *VCDClient) vcdAuthorize(user, pass, org string) (*http.Response, e
...
+
 	if err != nil {
 		return nil, err
 	}
 	defer resp.Body.Close()
 	// Store the authorization header
-	vcdCli.Client.VCDToken = resp.Header.Get(AuthorizationHeader)
-	vcdCli.Client.VCDAuthHeader = AuthorizationHeader
+	vcdCli.Client.VCDToken = resp.Header.Get(BearerTokenHeader)
+	vcdCli.Client.VCDAuthHeader = BearerTokenHeader
```
lvirbalas, line 118: I'm a bit confused by this part. Have we removed the authorization token completely in favor of the bearer token?

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 * Added functions to retrieve and use VCD version `client.GetVcdVersion`, `client.GetVcdShortVersion`, `client.GetVcdFullVersion`, `client.VersionEqualOrGreater` [#339](https://github.com/vmware/go-vcloud-director/pull/339)
 * Added methods `VM.UpdateStorageProfile`, `VM.UpdateStorageProfileAsync` [#338](https://github.com/vmware/go-vcloud-director/pull/338)
 * Added methods `adminVdc.UpdateStorageProfile` [#340](https://github.com/vmware/go-vcloud-director/pull/340)
+* Added support for bearer tokens [#341](https://github.com/vmware/go-vcloud-director/pull/341)
```
lvirbalas, line 6: Would it be worth mentioning that the selection between old and new token types are done automatically? I like that in this PR :)

govcd/api_vcd.go
dataclouder: No.
We have changed the token that we use to keep the client working. Previously, we injected the authorization token at connection time in the client, so that it would be used for further operations. Now we use the bearer token for the same purpose.

However, if you connect using the old authorization token, it will be used like before.

CHANGELOG.md
dataclouder: added "transparent" to signify that the token just works.

lvirbalas (approved):
LGTM!

PR #345, opened 2020-12-01
Add support for handling catalog storage profiles

Didainius:
It can be tricky to use catalogs while working with multiple VDCs without specifying particular storage profile  (https://github.com/vmware/terraform-provider-vcd/issues/598)

This PR adds functions to SDK so that one can handle Storage profiles for catalog during creation and update phases.

It improves functions `func (adminCatalog *AdminCatalog) Update()` and adds 
* `func (adminOrg *AdminOrg) CreateCatalogWithStorageProfile`
* `func (org *Org) CreateCatalogWithStorageProfile`

Additionally these functions are added for convenience:
* `func (adminOrg *AdminOrg) GetAllStorageProfileReferences`
* `func (adminOrg *AdminOrg) GetStorageProfileReferenceById`
* `func (adminOrg *AdminOrg) GetAllVDCs`

**Note**. Test suite passed on at least one env.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 * Added functions `edge.GetLbAppRules`, `edge.GetLbServerPools`, `edge.GetLbAppProfiles`, `edge.GetNsxvNatRules`, `client.GetOrgList`
 * Exported private function `client.maxSupportedVersion` to `client.MaxSupportedVersion`
 * Able to upload an OVF without ovf:size defined in File part. Some bug fix for uploading OVA/OVF. [#331](https://github.com/vmware/go-vcloud-director/pull/331)
-* Add support for handling catalog storage profiles (`adminOrg.CreateCatalogWithStorageProfile`,
+* Add support for handling catalog storage profile (`adminOrg.CreateCatalogWithStorageProfile`,
 `org.CreateCatalogWithStorageProfile`, `adminCatalog.Update`) [#345](https://github.com/vmware/go-vcloud-director/pull/345)
 * Add convenience functions `AdminOrg.GetAllStorageProfileReferences`, `AdminOrg.GetStorageProfileReferenceById`, `AdminOrg.GetAllVDCs`  [#345](https://github.com/vmware/go-vcloud-director/pull/345)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 * Added functions `edge.GetLbAppRules`, `edge.GetLbServerPools`, `edge.GetLbAppProfiles`, `edge.GetNsxvNatRules`, `client.GetOrgList`
 * Exported private function `client.maxSupportedVersion` to `client.MaxSupportedVersion`
 * Able to upload an OVF without ovf:size defined in File part. Some bug fix for uploading OVA/OVF. [#331](https://github.com/vmware/go-vcloud-director/pull/331)
+* Add support for handling catalog storage profiles (`adminOrg.CreateCatalogWithStorageProfile`,
+`org.CreateCatalogWithStorageProfile`, `adminCatalog.Update`) [#345](https://github.com/vmware/go-vcloud-director/pull/345)
+* Add convenience functions `AdminOrg.GetAllStorageProfileReferences`, `AdminOrg.GetStorageProfileReferenceById`, `AdminOrg.GetAllVDCs`  [#345](https://github.com/vmware/go-vcloud-director/pull/345)
 
 BREAKING CHANGES:
 
--- a/govcd/adminorg.go
+++ b/govcd/adminorg.go
@@ -42,6 +42,57 @@ func (adminOrg *AdminOrg) CreateCatalog(name, description string) (AdminCatalog,
 	return CreateCatalog(adminOrg.client, adminOrg.AdminOrg.Link, name, description)
 }
 
+// CreateCatalogWithStorageProfile is like CreateCatalog, but allows to specify storage profile
+func (adminOrg *AdminOrg) CreateCatalogWithStorageProfile(name, description string, storageProfiles *types.CatalogStorageProfiles) (AdminCatalog, error) {
+	return CreateCatalogWithStorageProfile(adminOrg.client, adminOrg.AdminOrg.Link, name, description, storageProfiles)
+}
+
+// GetAllVDCs returns all child VDCs
+func (adminOrg *AdminOrg) GetAllVDCs(refresh bool) ([]*Vdc, error) {
+	if refresh {
+		err := adminOrg.Refresh()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	allVdcs := make([]*Vdc, len(adminOrg.AdminOrg.Vdcs.Vdcs))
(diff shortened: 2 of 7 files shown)


govcd/adminorg.go
```diff
@@ -42,6 +42,74 @@ func (adminOrg *AdminOrg) CreateCatalog(name, description string) (AdminCatalog,
 	return CreateCatalog(adminOrg.client, adminOrg.AdminOrg.Link, name, description)
 }
 
+// CreateCatalogWithStorageProfile is like CreateCatalog, but allows to specify storage profile
+func (adminOrg *AdminOrg) CreateCatalogWithStorageProfile(name, description string, storageProfiles *types.CatalogStorageProfiles) (AdminCatalog, error) {
```
vbauzysvmware, line 46: Let's follow our guidelines and return referenced AdminCatalog

govcd/adminorg.go
```diff
@@ -42,6 +42,74 @@ func (adminOrg *AdminOrg) CreateCatalog(name, description string) (AdminCatalog,
...
+	return CreateCatalogWithStorageProfile(adminOrg.client, adminOrg.AdminOrg.Link, name, description, storageProfiles)
```
vbauzysvmware, line 47: does this mean that catalog can have assigned multi storage profiles?

vbauzysvmware (changes requested):
Nice improvements for handling catalog

govcd/adminorg.go
```diff
@@ -42,6 +42,74 @@ func (adminOrg *AdminOrg) CreateCatalog(name, description string) (AdminCatalog,
...
+}
+
+// GetAllVDCs returns all child VDCs
```
dataclouder, line 50: I would say "all depending VDCs" because using "child" as adjective for plural recipients doesn't sound right

govcd/adminorg.go
```diff
@@ -42,6 +42,74 @@ func (adminOrg *AdminOrg) CreateCatalog(name, description string) (AdminCatalog,
...
+			return nil, fmt.Errorf("error retrieving VDC '%s': %s", vdc.Vdc.Name, err)
+		}
+		allVdcs[vdcIndex] = vdc
+
+	}
+
+	return allVdcs, nil
+}
+
+// GetAllStorageProfileReferences traverses all child VDCs and returns a slice of storage profile references
```
dataclouder, line 72: "all depending VDCs"

govcd/adminorg.go
```diff
@@ -42,6 +42,74 @@ func (adminOrg *AdminOrg) CreateCatalog(name, description string) (AdminCatalog,
...
+	for _, vdc := range allVdcs {
+		if len(vdc.Vdc.VdcStorageProfiles.VdcStorageProfile) > 0 {
+			allStorageProfileReferences = append(allStorageProfileReferences, vdc.Vdc.VdcStorageProfiles.VdcStorageProfile...)
+		}
+	}
+
+	return allStorageProfileReferences, nil
+}
+
+func (adminOrg *AdminOrg) GetStorageProfileReferenceById(id string, refresh bool) (*types.Reference, error) {
```
dataclouder, line 96: Missing comment

govcd/org.go
```diff
@@ -91,13 +91,19 @@ func (org *Org) GetVdcByName(vdcname string) (Vdc, error) {
 }
 
 func CreateCatalog(client *Client, links types.LinkList, Name, Description string) (AdminCatalog, error) {
```
dataclouder, line 93: missing func description

types/v56/types.go
```diff
@@ -603,7 +603,7 @@ type ComputeCapacity struct {
 // Description: A reference to a resource. Contains an href attribute and optional name and type attributes.
 // Since: 0.9
 type Reference struct {
-	HREF string `xml:"href,attr"`
+	HREF string `xml:"href,attr,omitempty"`
```
dataclouder, line 606: In which case does a Reference return an empty HREF?

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@
 * Added functions `edge.GetLbAppRules`, `edge.GetLbServerPools`, `edge.GetLbAppProfiles`, `edge.GetNsxvNatRules`, `client.GetOrgList`
 * Exported private function `client.maxSupportedVersion` to `client.MaxSupportedVersion`
 * Able to upload an OVF without ovf:size defined in File part. Some bug fix for uploading OVA/OVF. [#331](https://github.com/vmware/go-vcloud-director/pull/331)
+* Add support for handling catalog storage profiles (`adminOrg.CreateCatalogWithStorageProfile`,
```
dataclouder, line 11: I thought that catalog can only have ONE storage profile.

dataclouder (changes requested):

types/v56/types.go
Didainius: In fact this was not for return, but for "POST"/"PUT". When attaching storage profile to catalog I wanted to use IDs (because they are what we use and having only ID is sufficient for the task). What happened is it would send ID and empty HREF field (`href=""`)which caused validation error.

CHANGELOG.md
Didainius: I had this as a general term "storage profiles" and yes - it is true that it only handles one. Will change.

govcd/adminorg.go
Didainius: Nope, but API  structure is such. The other way is I could always wrap it in this function. I just accepted this structure as it looked more convenient (and maybe future proof although I know it may never happen).

govcd/adminorg.go
Didainius: Can do. The reason here was that it is "sister" function of `CreateCatalog`.

vbauzysvmware: Well, that "sister" is granny with older standard :)

govcd/adminorg.go
vbauzysvmware: Well due this naming and type I got strange impression that multi storage profiles supported:)

dataclouder (approved):
LGTM

govcd/adminorg.go
Didainius: Yes, the structure accepts a slice, but I left it as I see it as a possibility for future. If it can use "any of all", why can't it use "any of specified".

CHANGELOG.md
Didainius: Changed.

govcd/org.go
Didainius: Added

govcd/adminorg.go
Didainius: Added

govcd/adminorg.go
Didainius: Fixed

govcd/adminorg.go
Didainius: Fixed

govcd/adminorg.go
Didainius: Ok, let them betray the family :)

vbauzysvmware (approved):

lvirbalas (approved):
Looks neat!

PR #348, opened 2020-12-18
Introduce generic OpenAPI entity cleanup for tests

Didainius:
I was going to introduce these changes with Org VDC network PRs but because this touches many points I have split this before going with Org VDC network PR.

OpenAPI has a standartized working approach which allows to improve on current entity cleanup operations in tests.

Because _delete_ endpoints for OpenAPI are constructed in the same fashion - all items can be deleted using low level OpenAPI function `OpenApiDeleteItem` provided endpoint (together with ID) are known. And these are easy to know once object is created.

This PR sets up code for generic openapi entity cleanup  (by adding two new cleanup functions `AddToCleanupListOpenApi` and `PrependToCleanupListOpenApi`)and also uses it in external network V2
```go
	// Use generic "OpenApiEntity" resource cleanup type
	openApiEndpoint := endpoint + createdNet.ExternalNetwork.ID
	AddToCleanupListOpenApi(createdNet.ExternalNetwork.Name, "OpenApiEntity",  check.TestName(), openApiEndpoint)
```
The above code generates such entry in cleanup list:
```json
  "EntityList": [
   {
    "Name": "testNet",
    "EntityType": "OpenApiEntity",
    "Parent": "",
    "CreatedBy": "TestVCD.Test_CreateExternalNetworkV2Nsxt",
    "OpenApiEndpoint": "1.0.0/externalNetworks/urn:vcloud:network:0d2d7953-1277-427e-b9b0-47710771286d"
   }
  ]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.11.0 (Unreleased)
+
+BUGS FIXED:
+* Made IPAddress field for IPAddresses struct to array [#350](https://github.com/vmware/go-vcloud-director/pull/350)
+
 ## 2.10.0 (December 18, 2020)
 
 * Added functions to retrieve and use VCD version `client.GetVcdVersion`, `client.GetVcdShortVersion`, `client.GetVcdFullVersion`, `client.VersionEqualOrGreater` [#339](https://github.com/vmware/go-vcloud-director/pull/339)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 BUGS FIXED:
 * Made IPAddress field for IPAddresses struct to array [#350](https://github.com/vmware/go-vcloud-director/pull/350)
 
+IMPROVEMENTS:
+* Introduce generic OpenAPI entity cleanup for tests [348](https://github.com/vmware/go-vcloud-director/pull/348)
 ## 2.10.0 (December 18, 2020)
 
 * Added functions to retrieve and use VCD version `client.GetVcdVersion`, `client.GetVcdShortVersion`, `client.GetVcdFullVersion`, `client.VersionEqualOrGreater` [#339](https://github.com/vmware/go-vcloud-director/pull/339)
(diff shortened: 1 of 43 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 BUGS FIXED:
 * Made IPAddress field for IPAddresses struct to array [#350](https://github.com/vmware/go-vcloud-director/pull/350)
 
+IMPROVEMENTS:
+* Introduce generic OpenAPI entity cleanup for tests [348](https://github.com/vmware/go-vcloud-director/pull/348)
 ## 2.10.0 (December 18, 2020)
```
lvirbalas, line 8: New line missing?

Didainius: Yes. Added.

lvirbalas (commented):
I have only one more comment while looking at forty files changed :)

Maybe a separate function for the new functionality? Something like this:

```
func AddToCleanupList(name, entityType, parent, createdBy)
func AddToCleanupListURL(name, entityType, parent, createdBy, openApiEndpoint string)
```

`AddToCleanupList` - would only call with the last parameter empty `AddToCleanupListURL(..., "")`

This is just a thought. I don't mind changing forty files if you and @dataclouder think code is easier to manage with a single function long-term.

Didainius:
> I have only one more comment while looking at forty files changed :)
> 
> Maybe a separate function for the new functionality? Something like this:
> 
> ```
> func AddToCleanupList(name, entityType, parent, createdBy)
> func AddToCleanupListURL(name, entityType, parent, createdBy, openApiEndpoint string)
> ```
> 
> `AddToCleanupList` - would only call with the last parameter empty `AddToCleanupListURL(..., "")`
> 
> This is just a thought. I don't mind changing forty files if you and @dataclouder think code is easier to manage with a single function long-term.

I was on the edge between this solution and the other. I guess one more thought like that weighs id down. Let me change it.

lvirbalas (approved):

dataclouder (approved):

PR #354, opened 2021-01-22
Introduce OpenAPI based Org VDC network management functions for NSX-V and NSX-T

Didainius:
This PR introduces:
* `OpenApiOrgVdcNetwork` structure and type meant to handle Org VDC networks for both NSX-T and NSX-V backed.
* Structure `NsxtImportableSwitch` meant for lookup "importable switches" (NSX-T segments) to be used for NSX-T Imported networks. Methods `GetNsxtImportableSwitchByName` and `GetAllNsxtImportableSwitches` aid that
* Methods `vdc.IsNsxt` and `vdc.IsNsxv` help to verify which type of network provider this VDC is backed.

Additionally it ads:
* nsxtVdc shorthand in `TestVCD` struct for testing
* Two new config fields `edgeGateway` (to use existing edge gateway) and `unusedSegment` which is needed to test our Imported NSX-T Org VDC network
* Converts VirtualQuantity int->int64 to avoid variable overflow with big sizes (https://github.com/vmware/terraform-provider-vcd/issues/626)


govcd/api_vcd_test.go
```diff
@@ -566,8 +569,13 @@ func (vcd *TestVCD) SetUpSuite(check *C) {
 		panic(err)
 	}
 
-	// This is set explicitly to make sure that catalog is using storage profile from correct VDC
-	setCatalogStorageProfile(vcd, check)
+	// configure NSX-T VDC for convenience if it is specified in configuration
+	if config.VCD.Nsxt.Vdc != "" {
+		vcd.nsxtVdc, err = vcd.org.GetVDCByName(config.VCD.Nsxt.Vdc, false)
+		if err != nil {
+			panic(fmt.Errorf("error geting NSX-T Vdc '%s': %s", config.VCD.Nsxt.Vdc, err))
```
lvirbalas, line 576: Nit:

```suggestion
			panic(fmt.Errorf("error geting NSX-T VDC '%s': %s", config.VCD.Nsxt.Vdc, err))
```

govcd/openapi_org_network.go
```diff
@@ -0,0 +1,270 @@
...
+
+	err = orgVdcNet.client.OpenApiPutItem(minimumApiVersion, urlRef, nil, OrgVdcNetworkConfig, returnEgw.OpenApiOrgVdcNetwork)
+	if err != nil {
+		return nil, fmt.Errorf("error updating Org Vdc network: %s", err)
+	}
+
+	return returnEgw, nil
+}
+
+// Delete allows to delete Org VDC  network
```
lvirbalas, line 120: Nit:

```suggestion
// Delete allows to delete Org VDC network
```

govcd/openapi_org_network.go
```diff
@@ -0,0 +1,270 @@
...
+func (orgVdcNet *OpenApiOrgVdcNetwork) Delete() error {
+	endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointOrgVdcNetworks
+	minimumApiVersion, err := orgVdcNet.client.checkOpenApiEndpointCompatibility(endpoint)
+	if err != nil {
+		return err
+	}
+
+	if orgVdcNet.OpenApiOrgVdcNetwork.ID == "" {
+		return fmt.Errorf("cannot delete Org Vdc network without id")
```
lvirbalas, line 129:
```suggestion
		return fmt.Errorf("cannot delete Org VDC network without ID")
```

govcd/openapi_org_network.go
```diff
@@ -0,0 +1,270 @@
...
+
+	urlRef, err := orgVdcNet.client.OpenApiBuildEndpoint(endpoint, orgVdcNet.OpenApiOrgVdcNetwork.ID)
+	if err != nil {
+		return err
+	}
+
+	err = orgVdcNet.client.OpenApiDeleteItem(minimumApiVersion, urlRef, nil)
+
+	if err != nil {
+		return fmt.Errorf("error deleting Org Vdc: %s", err)
```
lvirbalas, line 140:
```suggestion
		return fmt.Errorf("error deleting Org VDC: %s", err)
```

govcd/openapi_org_network.go
```diff
@@ -0,0 +1,270 @@
...
+// func (vdc *Vdc) GetOpenApiOrgVdcNetworkById(id string) (*OpenApiOrgVdcNetwork, error)
+func getOpenApiOrgVdcNetworkById(client *Client, id string, queryParameters url.Values) (*OpenApiOrgVdcNetwork, error) {
+	endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointOrgVdcNetworks
+	minimumApiVersion, err := client.checkOpenApiEndpointCompatibility(endpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	if id == "" {
+		return nil, fmt.Errorf("empty Org VDC network id")
```
lvirbalas, line 182:
```suggestion
		return nil, fmt.Errorf("empty Org VDC network ID")
```

govcd/openapi_org_network.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	return egw, nil
+}
+
+// returnSingleOpenApiOrgVdcNetwork helps to reduce code duplication for `GetOpenApiOrgVdcNetworkByName` functions with different
+// receivers
+func returnSingleOpenApiOrgVdcNetwork(name string, allEdges []*OpenApiOrgVdcNetwork) (*OpenApiOrgVdcNetwork, error) {
+	if len(allEdges) > 1 {
+		return nil, fmt.Errorf("got more than 1 Org Vdc network by name '%s' %d", name, len(allEdges))
```
lvirbalas, line 207: Suggesting to keep user messages human-oriented:
```suggestion
		return nil, fmt.Errorf("got more than 1 Org VDC network by name '%s' %d", name, len(allEdges))
```
```suggestion
		return nil, fmt.Errorf("got more than one Org Vdc network by name '%s' %d", name, len(allEdges))
```

govcd/openapi_org_network.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	if len(allEdges) < 1 {
+		return nil, fmt.Errorf("%s: got 0 Org Vdc network by name '%s'", ErrorEntityNotFound, name)
```
lvirbalas, line 211:
```suggestion
		return nil, fmt.Errorf("%s: got zero Org Vdc networks by name '%s'", ErrorEntityNotFound, name)
```

govcd/openapi_org_network.go
```diff
@@ -0,0 +1,270 @@
...
+
+	// Enforce maximum pageSize to be 32 as API endpoint throws error if it is > 32
+	pageSizeString := queryParameters.Get("pageSize")
+
+	switch pageSizeString {
+	// If no pageSize is specified it must be set to 32 as by default low level API function OpenApiGetAllItems sets 128
+	case "":
+		queryParameters.Set("pageSize", "32")
+
+		// If pageSize is specified ensure it is not >32
```
lvirbalas, line 232: This comment seems misplaced.

govcd/vapp_test.go
```diff
@@ -143,7 +143,7 @@ func (vcd *TestVCD) Test_ChangeCPUCountWithCore(check *C) {
 		check.Skip("Skipping test because vApp was not successfully created at setup")
 	}
 
-	currentCpus := 0
+	currentCpus := int64(0)
```
lvirbalas, line 146: Wow, surprised we need to do this conversion here!

govcd/openapi_org_network.go
```diff
@@ -0,0 +1,270 @@
...
+	"net/url"
+	"strconv"
+
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+)
+
+// OpenApiOrgVdcNetwork uses OpenAPI endpoint to operate both - NSX-T and NSX-V Org VDC networks
+type OpenApiOrgVdcNetwork struct {
+	OpenApiOrgVdcNetwork *types.OpenApiOrgVdcNetwork
+	client               *Client
```
lvirbalas, line 18: Private/lower `client` case on purpose?

lvirbalas (changes requested):
Looks great, just a few nits :)

govcd/api_vcd_test.go
Didainius: Changed. I will get there one day. I do remember it most of the times now.

govcd/openapi_org_network.go
Didainius: Fixed.

govcd/openapi_org_network.go
Didainius: Changed here and in many other places.

govcd/openapi_org_network.go
Didainius: Changed here and in many other places.

govcd/openapi_org_network.go
Didainius: Changed.

govcd/openapi_org_network.go
Didainius: Changed

govcd/openapi_org_network.go
Didainius: This is in correct place. I might need to improve the comment then. The point is that above `case ""` sets pageSize to 32 if none was specified.
The `default` case will check what value was provided and if it is `>32` it will set it to `32`. This is needed because this particular endpoint throws an ugly error if `pageSize>32`

govcd/vapp_test.go
Didainius: Yep, we had to do it. Also in some places Assert compares not only values, but types as well and throws an error if it is not so.

govcd/openapi_org_network.go
Didainius: Yes. In fact we have it so in many/almost all places. VM, vApp, VDC, etc...

govcd/openapi_org_network.go
lvirbalas: Ah, so then the indentation is too big, as it should be inline with `default:` as right now it looks as if the comment belongs to the `case "":`

lvirbalas (approved):
LGTM!

dataclouder (approved):

Issue #363, opened 2021-03-22
GetEdgeGatewayRecordsType (and potentially other resources) not paginating


CmdrSharp:
Recently ran into an issue when Terraforming (using https://github.com/vmware/terraform-provider-vcd) which led me to dig deeper into this.
The issue was that randomly, when creating routed networks, Terraform would complain that edge gateways which indeed exist, do not. It would error out as such:
```
Error: [routed network read] error retrieving edge gateway namegoeshere: [ENF] entity not found
```

When digging into this, I decided to clone your example code and simply run the following code at the end, to receive a list of which edges are retrieved.

```go
	edgeGatewayList, err := vdc.GetEdgeGatewayRecordsType(true)
	if err != nil {
		fmt.Println(err)
		os.Exit(1)
	}

	for _, ert := range edgeGatewayList.EdgeGatewayRecord {
		fmt.Println(ert.Name)
	}
```

This indeed only returned the first 25 (same order as if I ran the query against /api/admin/vdc/{vdc_id}/edgeGateways myself).

Perhaps this function (and maybe other resource functions that I suspect may have the same issue) should auto-paginate when required?

dataclouder:
Fixed in PR #364

PR #364, opened 2021-03-23
Fix edge gateway list retrieval

dataclouder:
Fix Issue #363
Edge gateway retrieval was cut at the first 25 items.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.12.0 (unreleased)
+
+* Added method `vdc.QueryEdgeGateway` [#364](https://github.com/vmware/go-vcloud-director/pull/364)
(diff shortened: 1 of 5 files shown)


lvirbalas (commented):
Why a new function as opposed to fixing the existing one to return all items?

dataclouder:
> Why a new function as opposed to fixing the existing one to return all items?

It was a good opportunity to normalize the name with the convention that a query returning a list should be called `Query+EntityType+List`

lvirbalas (approved):
LGTM!

BTW, how many usages of the old function do we have in the Terraform project?

Didainius (approved):
Thanks

dataclouder:
> BTW, how many usages of the old function do we have in the Terraform project?

One. And it was changed

PR #370, opened 2021-04-25
Empty requests should not contain xml.Header

Didainius:
Closes #https://github.com/vmware/terraform-provider-vcd/issues/661 

Some WAF configurations block requests when the body is empty, but contains `xml.Header`. This PR adds improvement to only sends `xml.Header` when payload is not empty.

This had no impact on test runs.

### DIFF
--- a/govcd/api.go
+++ b/govcd/api.go
@@ -601,8 +601,11 @@ func executeRequestCustomErr(pathURL string, params map[string]string, requestTy
 	url, _ := url.ParseRequestURI(pathURL)
 
 	var req *http.Request
-	switch requestType {
-	case http.MethodPost, http.MethodPut:
+	switch {
+	// Only send data (and xml.Header) if method is POST or PUT and the payload is actually provided to avoid sending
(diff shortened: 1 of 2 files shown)


govcd/api.go
```diff
@@ -601,9 +601,10 @@ func executeRequestCustomErr(pathURL string, params map[string]string, requestTy
 	url, _ := url.ParseRequestURI(pathURL)
 
 	var req *http.Request
-	switch requestType {
-	case http.MethodPost, http.MethodPut:
-
+	switch {
+	// Only send data (and xml.Header) if the payload is actually provided to avoid sending empty body with XML header
+	//(some Web Application Firewalls block requests when empty XML header is set but not body provided)
```
lvirbalas, line 606:
```suggestion
	// (some Web Application Firewalls block requests when empty XML header is set but not body provided)
```

lvirbalas (approved):
Nit and LGTM! Hopefully, this doesn't cause other firewalls to complain :)

dataclouder (approved):
LGTM

PR #372, opened 2021-04-28
Fix vapp description and add vapp.Rename

dataclouder:
* Added parameter `description` to method `vdc.ComposeRawVapp` (This method is used in `terraform-provider-vcd` to create a vApp.)
* Added methods `vapp.Rename`, `vapp.UpdateDescription`, `vapp.UpdateNameDescription`. This make it possible to update the vApp's name and description in Terraform.

### DIFF
--- a/govcd/common_test.go
+++ b/govcd/common_test.go
@@ -1,7 +1,7 @@
 // +build api functional catalog vapp gateway network org query extnetwork task vm vdc system disk lb lbAppRule lbAppProfile lbServerPool lbServiceMonitor lbVirtualServer user nsxv affinity ALL
 
 /*
- * Copyright 2019 VMware, Inc.  All rights reserved.  Licensed under the Apache v2 License.
+ * Copyright 2021 VMware, Inc.  All rights reserved.  Licensed under the Apache v2 License.
  */
 
 package govcd
@@ -672,7 +672,7 @@ func deleteVapp(vcd *TestVCD, name string) error {
 // makeEmptyVapp creates a given vApp without any VM
 func makeEmptyVapp(vdc *Vdc, name string) (*VApp, error) {
 
-	err := vdc.ComposeRawVApp(name, "")
+	err := vdc.ComposeRawVApp(name, name)
 	if err != nil {
 		return nil, err
 	}
--- a/govcd/common_test.go
+++ b/govcd/common_test.go
@@ -670,9 +670,9 @@ func deleteVapp(vcd *TestVCD, name string) error {
 }
 
 // makeEmptyVapp creates a given vApp without any VM
-func makeEmptyVapp(vdc *Vdc, name string) (*VApp, error) {
+func makeEmptyVapp(vdc *Vdc, name string, description string) (*VApp, error) {
 
-	err := vdc.ComposeRawVApp(name, "description of "+name)
+	err := vdc.ComposeRawVApp(name, description)
 	if err != nil {
 		return nil, err
 	}
--- a/govcd/common_test.go
+++ b/govcd/common_test.go
@@ -672,7 +672,7 @@ func deleteVapp(vcd *TestVCD, name string) error {
 // makeEmptyVapp creates a given vApp without any VM
 func makeEmptyVapp(vdc *Vdc, name string) (*VApp, error) {
 
-	err := vdc.ComposeRawVApp(name, name)
+	err := vdc.ComposeRawVApp(name, "description of "+name)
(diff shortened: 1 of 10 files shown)


govcd/common_test.go
```diff
@@ -672,7 +672,7 @@ func deleteVapp(vcd *TestVCD, name string) error {
 // makeEmptyVapp creates a given vApp without any VM
 func makeEmptyVapp(vdc *Vdc, name string) (*VApp, error) {
 
-	err := vdc.ComposeRawVApp(name)
+	err := vdc.ComposeRawVApp(name, name)
```
lvirbalas, line 675: Should the second parameter hold desription?

govcd/vapp.go
```diff
@@ -1375,3 +1376,71 @@ func (vapp *VApp) getOrgInfo() (orgInfoType, error) {
...
+
+// UpdateNameDescription can change the name and the description of a vApp
+// If name or description are empty, they are not changed.
+func (vapp *VApp) UpdateNameDescription(newName, newDescription string) error {
+	if vapp == nil || vapp.VApp.HREF == "" {
+		return fmt.Errorf("vapp or href cannot be empty")
+	}
+
+	// Skip update if we are using the original values
+	if (newName == vapp.VApp.Name || newName == "") && (newDescription == vapp.VApp.Description || newDescription == "") {
```
lvirbalas, line 1388: How would one set description to empty?

govcd/vapp_test.go
```diff
@@ -1557,3 +1558,101 @@ func (vcd *TestVCD) Test_AddNewVMWithComputeCapacity(check *C) {
...
+	AddToCleanupList(vappName, "vapp", "", "Test_RenameVapp")
+
+	check.Assert(vapp.VApp.Name, Equals, vappName)
+	check.Assert(vapp.VApp.Description, Equals, vappName)
+
+	// Need a slight delay for the vApp to get the links that are needed for renaming
+	time.Sleep(time.Second)
+
+	// change description
+	vcd.testUpdateVapp(check, vapp, "", newVappDescription, nil)
```
lvirbalas, line 1628: I may have missed it, but we need a test which tries to re-set description to empty.

govcd/common_test.go
dataclouder: This is a testing function that is not using the description. Since I changed the signature of `ComposeRawVapp`, I added the vApp name as description.

govcd/vapp.go
dataclouder: Changed to allow description removal

govcd/vapp_test.go
dataclouder: added

govcd/common_test.go
lvirbalas: Nit, but you could use something like `"Description of " + name` - this way in the unlikely case a person would look into this populated field, it would be clear that it works and is not a bug.

dataclouder: Changed.
However, the life span of a vApp created by this function is usually measured in minutes. The likelihood of someone looking into it is negligible.

lvirbalas: That's why it's a nit :) BTW, I think gofmt is needed there.

lvirbalas (approved):
LGTM!

govcd/vapp.go
```diff
@@ -1375,3 +1376,71 @@ func (vapp *VApp) getOrgInfo() (orgInfoType, error) {
...
+		return nil
+	}
+
+	opType := "application/vnd.vmware.vcloud.recomposeVAppParams+xml"
```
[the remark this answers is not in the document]
dataclouder: Yes. Fixed

govcd/vapp.go
```diff
@@ -1375,3 +1376,71 @@ func (vapp *VApp) getOrgInfo() (orgInfoType, error) {
 	}
 	return getOrgInfo(vapp.client, vdc.Vdc.Link, vapp.VApp.ID, vapp.VApp.Name, "vApp")
 }
+
+// UpdateNameDescription can change the name and the description of a vApp
+// If name or description are empty, they are not changed.
+func (vapp *VApp) UpdateNameDescription(newName, newDescription string) error {
+	if vapp == nil || vapp.VApp.HREF == "" {
+		return fmt.Errorf("vapp or href cannot be empty")
```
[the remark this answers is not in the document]
dataclouder: changed

Didainius (approved):
It works!

PR #375, opened 2021-05-11
Add AdminOrgVdcStorageProfile for querying

lelvisl:
## Added:
- QueryResultAdminOrgVdcStorageProfileRecordType type.
- Constants for querying: QtOrgVdcStorageProfile and QtAdminOrgVdcStorageProfile.
- Funcs for querying : QueryAdminOrgVdcStorageProfileByID and QueryOrgVdcStorageProfileByID.

QueryOrgVdcStorageProfileByID (non Admin) not workin (tested wiht vCloud 10.0 and 10.2.2).
May be query isn't correct.

### DIFF
--- a/govcd/sample_govcd_test_config.yaml
+++ b/govcd/sample_govcd_test_config.yaml
@@ -185,3 +185,8 @@ media:
   mediaPath: ../test-resources/test.iso
   # Existing media in test system
   mediaName: uploadedMediaName
+misc:
+  # By default tests in this repository pick LDAP container 'rroemhild/test-openldap'. As docker throttles downloads
+  # it can help to host the image on local registry and pull it from there. This variable overrides default container
+  # location when set.
+  #ldapContainer: custom-registry.yyy/directory/test-openldap:latest
--- a/govcd/sample_govcd_test_config.yaml
+++ b/govcd/sample_govcd_test_config.yaml
@@ -102,8 +102,6 @@ vcd:
     storageProfile:
         # First storage profile (mandatory)
         storageProfile1: Development
-        # ID of storage profile. If omitted, some tests will be skipped.
-        storageProfile1ID: 8263c1b5-6a40-404f-84a3-72343fce7c20
         # Second storage profile. If omitted, some tests will be skipped.
         storageProfile2: "*"
     # An edge gateway
--- a/govcd/sample_govcd_test_config.yaml
+++ b/govcd/sample_govcd_test_config.yaml
@@ -102,6 +102,8 @@ vcd:
     storageProfile:
         # First storage profile (mandatory)
         storageProfile1: Development
+        # ID of storage profile. If omitted, some tests will be skipped.
+        storageProfile1ID: 8263c1b5-6a40-404f-84a3-72343fce7c20
         # Second storage profile. If omitted, some tests will be skipped.
         storageProfile2: "*"
     # An edge gateway
--- a/govcd/system.go
+++ b/govcd/system.go
@@ -986,8 +986,11 @@ func QueryAdminOrgVdcStorageProfileByID(vcdCli *VCDClient, id string) (*types.Qu
 	if err != nil {
 		return nil, err
 	}
-	if len(results.Results.AdminOrgVdcStorageProfileRecord) != 1 {
-		return nil, errors.New("error querying storage profile")
+	if len(results.Results.AdminOrgVdcStorageProfileRecord) == 0 {
+		return nil, ErrorEntityNotFound
+	}
+	if len(results.Results.AdminOrgVdcStorageProfileRecord) > 1 {
+		return nil, fmt.Errorf("more than one Storage Profil found with ID %s", id)
 	}
 	return results.Results.AdminOrgVdcStorageProfileRecord[0], nil
 }
@@ -1002,8 +1005,11 @@ func QueryOrgVdcStorageProfileByID(vcdCli *VCDClient, id string) (*types.QueryRe
 	if err != nil {
 		return nil, err
 	}
-	if len(results.Results.OrgVdcStorageProfileRecord) != 1 {
-		return nil, errors.New("error querying storage profile")
+	if len(results.Results.OrgVdcStorageProfileRecord) == 0 {
+		return nil, ErrorEntityNotFound
+	}
+	if len(results.Results.OrgVdcStorageProfileRecord) > 1 {
+		return nil, fmt.Errorf("more than one Storage Profil found with ID %s", id)
 	}
 	return results.Results.OrgVdcStorageProfileRecord[0], nil
 }
--- a/govcd/system.go
+++ b/govcd/system.go
@@ -975,3 +975,29 @@ func (vcdCli *VCDClient) GetOrgList() (*types.OrgList, error) {
 	}
 	return orgList, nil
 }
+
+// QueryAdminOrgVdcStorageProfileByID finds a StorageProfile of VDC by ID as admin
+func QueryAdminOrgVdcStorageProfileByID(vcdCli *VCDClient, id string) ([]*types.QueryResultAdminOrgVdcStorageProfileRecordType, error) {
+	results, err := vcdCli.QueryWithNotEncodedParams(nil, map[string]string{
+		"type":          types.QtAdminOrgVdcStorageProfile,
+		"filter":        fmt.Sprintf("id==%s", url.QueryEscape(id)),
+		"filterEncoded": "true",
+	})
+	if err != nil {
+		return nil, err
+	}
+	return results.Results.AdminOrgVdcStorageProfileRecord, nil
+}
+
+// QueryOrgVdcStorageProfileByID finds a StorageProfile of VDC by ID
+func QueryOrgVdcStorageProfileByID(vcdCli *VCDClient, id string) ([]*types.QueryResultOrgVdcStorageProfileRecordType, error) {
+	results, err := vcdCli.QueryWithNotEncodedParams(nil, map[string]string{
+		"type":          types.QtOrgVdcStorageProfile,
+		"filter":        fmt.Sprintf("id==%s", url.QueryEscape(id)),
+		"filterEncoded": "true",
+	})
+	if err != nil {
+		return nil, err
+	}
+	return results.Results.OrgVdcStorageProfileRecord, nil
+}
--- a/govcd/system.go
+++ b/govcd/system.go
@@ -977,7 +977,7 @@ func (vcdCli *VCDClient) GetOrgList() (*types.OrgList, error) {
 }
 
 // QueryAdminOrgVdcStorageProfileByID finds a StorageProfile of VDC by ID as admin
-func QueryAdminOrgVdcStorageProfileByID(vcdCli *VCDClient, id string) ([]*types.QueryResultAdminOrgVdcStorageProfileRecordType, error) {
+func QueryAdminOrgVdcStorageProfileByID(vcdCli *VCDClient, id string) (*types.QueryResultAdminOrgVdcStorageProfileRecordType, error) {
 	results, err := vcdCli.QueryWithNotEncodedParams(nil, map[string]string{
 		"type":          types.QtAdminOrgVdcStorageProfile,
 		"filter":        fmt.Sprintf("id==%s", url.QueryEscape(id)),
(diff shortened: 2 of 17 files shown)


lelvisl:
`QueryOrgVdcStorageProfileByID (non Admin) not workin (tested wiht vCloud 10.0 and 10.2.2)` - wokring well if you are not provider admin. 
As i think, this PR can be merged.

lelvisl:
Can we merge this PR or we need to discuss something?

types/v56/constants.go
```diff
@@ -236,22 +236,25 @@ const (
...
+	QtMedia                     = "media"                     // media item
+	QtVm                        = "vm"                        // Virtual machine
+	QtAdminVm                   = "adminVM"                   // Virtual machine as admin
+	QtVapp                      = "vApp"                      // vApp
+	QtAdminVapp                 = "adminVApp"                 // vApp as admin
+	QtOrgVdc                    = "orgVdc"                    // Org VDC
+	QtAdminOrgVdc               = "adminOrgVdc"               // Org VDC as admin
+	QtOrgVdcStorageProfile      = "orgVdcStorageProfile"      // StorageProfile of VDC
+	QtAdminOrgVdcStorageProfile = "adminOrgVdcStorageProfile" // StorageProfile of VDC as admin
+
```
Didainius, line 257: No need for empty line

types/v56/types.go
```diff
@@ -2147,6 +2147,7 @@ type QueryResultRecordsType struct {
 	VAppRecord                      []*QueryResultVAppRecordType                      `xml:"VAppRecord"`                      // A record representing a VApp result.
 	AdminVAppRecord                 []*QueryResultVAppRecordType                      `xml:"AdminVAppRecord"`                 // A record representing a VApp result as admin.
 	OrgVdcStorageProfileRecord      []*QueryResultOrgVdcStorageProfileRecordType      `xml:"OrgVdcStorageProfileRecord"`      // A record representing storage profiles
+	AdminOrgVdcStorageProfileRecord []*QueryResultAdminOrgVdcStorageProfileRecordType `xml:"AdminOrgVdcStorageProfileRecord"` // A record representing storage profiles
```
Didainius, line 2150:
```suggestion
	AdminOrgVdcStorageProfileRecord []*QueryResultAdminOrgVdcStorageProfileRecordType `xml:"AdminOrgVdcStorageProfileRecord"` // A record representing Admin storage profiles
```

Didainius (commented):
Sorry for this drag. Going for review and testing.

Would you be able to add a line to `CHANGELOG.md` about this?
Also - any chances you could put a few tests into `system_test.go` (there are quite a few similar ones if you need inspiration :))

lelvisl:
Thanks for your review! Can you look again please?

govcd/sample_govcd_test_config.yaml
```diff
@@ -102,6 +102,8 @@ vcd:
     storageProfile:
         # First storage profile (mandatory)
         storageProfile1: Development
+        # ID of storage profile. If omitted, some tests will be skipped.
+        storageProfile1ID: 8263c1b5-6a40-404f-84a3-72343fce7c20
```
Didainius, line 106: Thanks for adding tests!
I will just ask to try and avoid adding this value to config. You could in fact retrieve a storage profile ID in your test  by using some of existing functions below to get ID by using Name:
```
func (vdc *Vdc) FindStorageProfileReference(name string) (types.Reference, error) {
func (adminOrg *AdminOrg) GetAllStorageProfileReferences(refresh bool) ([]*types.Reference, error) {
```

That way we don't have to introduce a new variable into test config, which is a struggle to populate.

lelvisl: 👌

lelvisl: fixed

govcd/system.go
```diff
@@ -975,3 +975,29 @@ func (vcdCli *VCDClient) GetOrgList() (*types.OrgList, error) {
 	}
 	return orgList, nil
 }
+
+// QueryAdminOrgVdcStorageProfileByID finds a StorageProfile of VDC by ID as admin
+func QueryAdminOrgVdcStorageProfileByID(vcdCli *VCDClient, id string) ([]*types.QueryResultAdminOrgVdcStorageProfileRecordType, error) {
```
Didainius, line 980: Thanks for this. Now just pardon, but I just tried to play with code and one thing hit me.
I think the signature could be:
```suggestion
func QueryAdminOrgVdcStorageProfileByID(vcdCli *VCDClient, id string) (*types.QueryResultAdminOrgVdcStorageProfileRecordType, error) {
```

That is - you don't need to return a slice of results because you are querying by ID which guarantees that there is either exactly one result, or not.

A slice could be if it was `QueryOrgVdcStorageProfile` which doesn't guarantee one object.

Didainius:
Not sure if you saw that, but I put a comment: https://github.com/vmware/go-vcloud-director/pull/375#discussion_r641518590

govcd/system.go
lelvisl: Can i then rename this method to `GetAdminOrgVdcStorageProfileByID`?

lelvisl: Or implement `QueryOrgVdcStorageProfile` with params and `GetAdminOrgVdcStorageProfileByID` that use previous?

Didainius: `GetX` in this SDK usually mean that it is not an object from query endpoint, but from its legitimate endpoint (e.g. `AdminOrg.GetVDCById` would not return a `Query` result, but the VDC struct itself. Query and object endpoints usually return different data that is why we try to keep `Query` and `Get` (to differentiate these things). This is not 100% due to historic reasons, but that is a good thing to follow.

Or - if you want to implement a function that returns all objects then it could be `QueryAdminOrgVdcStorageProfile`. But if you simply leave the code as it is (and not apply ID filter) you will hit a problem - by default it returns only the first page of data (which is usually up to 25 items). It is not easy to spot this on testing envs where not many resources exist, but we have already hit that and fixed some of the problems here. The trick here is to use `client.cumulativeQuery` internally which handles pagination and makes sure all pages are retrieved. (You could look through functions like `QueryEdgeGatewayList` or `QueryAllVdcs` to see how it is implemented.)

lelvisl: Agree with you and changed signature.

About `QueryAdminOrgVdcStorageProfile` - i think we need separate issue for discuss.

govcd/system.go
```diff
@@ -999,5 +1002,8 @@ func QueryOrgVdcStorageProfileByID(vcdCli *VCDClient, id string) ([]*types.Query
 	if err != nil {
 		return nil, err
 	}
-	return results.Results.OrgVdcStorageProfileRecord, nil
+	if len(results.Results.OrgVdcStorageProfileRecord) != 1 {
```
Didainius, line 1005: I hope you don't mind, but one more spot.

There is a particular sentinel error definition `ErrorEntityNotFound` in this SDK. When there are no results (and we look for an exact thing - we do return `ErrorEntityNotFound`. On the other hand if there are more than one results (which is unlikely in this case because you filter by ID - we do not return such error.

This helps in code to handle a specific error case (no element found instead of some connection, malformed id, etc. error). To ease consumption of such error two additional helper functions are in place for validation - `ContainsNotFound` and `IsNotFound`. 


In some cases it is a very convenient way to then write check like that:

```
if ContainsNotFound(err) { // internally it validates if err != nil
   // handle unavailable item
}

if err != nil { // Some other error occured and must be handled differently.

}
```

Excerpt from `QueryVmById`
```
	if len(foundVM) == 0 {
	}
	if len(foundVM) > 1 {
		return nil, fmt.Errorf("more than one VM found with ID %s", id)
	}
```

Didainius (commented):
One last convenience spot from me and I agree `QueryAdminOrgVdcStorageProfile` can be left for later on.

govcd/system.go
lelvisl: no problems, fixed.

govcd/system.go
```diff
@@ -975,3 +975,41 @@ func (vcdCli *VCDClient) GetOrgList() (*types.OrgList, error) {
...
+		"filterEncoded": "true",
+	})
+	if err != nil {
+		return nil, err
+	}
+	if len(results.Results.AdminOrgVdcStorageProfileRecord) == 0 {
+		return nil, ErrorEntityNotFound
+	}
+	if len(results.Results.AdminOrgVdcStorageProfileRecord) > 1 {
+		return nil, fmt.Errorf("more than one Storage Profil found with ID %s", id)
```
Didainius, line 993:
```suggestion
		return nil, fmt.Errorf("more than one Storage Profile found with ID %s", id)
```

govcd/system.go
```diff
@@ -975,3 +975,41 @@ func (vcdCli *VCDClient) GetOrgList() (*types.OrgList, error) {
...
+		"filterEncoded": "true",
+	})
+	if err != nil {
+		return nil, err
+	}
+	if len(results.Results.OrgVdcStorageProfileRecord) == 0 {
+		return nil, ErrorEntityNotFound
+	}
+	if len(results.Results.OrgVdcStorageProfileRecord) > 1 {
+		return nil, fmt.Errorf("more than one Storage Profil found with ID %s", id)
```
Didainius, line 1012:
```suggestion
		return nil, fmt.Errorf("more than one Storage Profile found with ID %s", id)
```

Didainius (approved):
Thanks. Approved and I left two typo comments and will wait for other team members to have a glimpse before we can merge.

govcd/system.go
lelvisl: oh, sorry

lvirbalas (approved):
LGTM and thanks for contributing!

lelvisl:
@dataclouder can you please review this PR?

govcd/system.go
```diff
@@ -975,3 +975,41 @@ func (vcdCli *VCDClient) GetOrgList() (*types.OrgList, error) {
...
+		return nil, ErrorEntityNotFound
+	}
+	if len(results.Results.AdminOrgVdcStorageProfileRecord) > 1 {
+		return nil, fmt.Errorf("more than one Storage Profile found with ID %s", id)
+	}
+	return results.Results.AdminOrgVdcStorageProfileRecord[0], nil
+}
+
+// QueryOrgVdcStorageProfileByID finds a StorageProfile of VDC by ID
+func QueryOrgVdcStorageProfileByID(vcdCli *VCDClient, id string) (*types.QueryResultOrgVdcStorageProfileRecordType, error) {
```
dataclouder, line 999: Missing test for this function.

Moreover, since we are reviewing this query type, the original structure – which was added in 2016 by another team – is missing several fields. 
Here is what I get when fetching an `OrgVdcStorageProfileRecord`
```xml
<OrgVdcStorageProfileRecord
diskIopsDefault="0"
diskIopsEnabled="false"
diskIopsMax="0"
diskIopsPerGbMax="0"
iopsAllocated="0"
iopsLimit="0"
isDefaultStorageProfile="true"
isEnabled="true"
isVdcBusy="false"
name="*"
numberOfConditions="0"
storageLimitMB="0"
storageUsedMB="34817"
vdc="https://example.com/api/vdc/677720eb-28b0-4bab-b108-38f4a2238db2"
vdcName="vdc-datacloud"
href="https://example.com/api/vdcStorageProfile/fb2c5d4f-78f4-4b49-afa5-6b34c79de529"/>
```
Several fields are missing from the original declaration. When you add the test for this function, you should either add such fields, or provide a good reason for skipping them.

dataclouder (changes requested):

dataclouder:
Another issue: You should merge your PR with master, as some important pieces are missing. When you run the test for  `QueryOrgVdcStorageProfileByID`, for example, you will need tenant credentials, and with the current code authentication will fail. (The bug was fixed in PR #376 )

lelvisl:
@dataclouder i merged upstream into my brunch and updated all fields.

About tests:
Please look at https://github.com/vmware/go-vcloud-director/issues/384. Test suit now is provider oriented, i can't (or don't know) how to setup both test successful from single user and i add skip in case of `Client.IsSysAdmin` (but it's not looking good for me).

govcd/system.go
lelvisl: Add tests and fields, but look please my last comment

lelvisl:
How about this change? Close? Or you want this changes? If yes - please give some feedback. 

As i see - PR https://github.com/vmware/go-vcloud-director/pull/393 doesn't solve this problem

Didainius:
@lelvisl, can you have a look at #499. We're finally hoping to resolve this and close. I merged up with latest, impoved tests.

PR #371, opened 2021-04-28
Drop support for EOL VCD 9.7 and bump default API version to v33.0

Didainius:
The purpose of this PR is to bump default API version used to V33.0 (which effectively drops support for VCD 9.7 which is EOL).

There are a few consequences to this:
* Drop legacy authentication mechanism (`vcdAuthorize`) and use only new Cloud API provided (`vcdCloudApiAuthorize`) as API V33.0 is sufficient for it. **Note** This is transparent to users
* `org.GetVDCByName` was refactored because `<Org>` structure does not contain child VDCs anymore and it has to be retrieved using query. As a result  `QueryResultOrgVdcRecordType` and `org.QueryOrgVdcList()` were introduced. **Note** `org.QueryOrgVdcList` requires to use `X-VMWARE-VCLOUD-AUTH-CONTEXT` and `X-VMWARE-VCLOUD-TENANT-CONTEXT` headers therefore some private functions were adjusted to allow setting headers
* `org.GetCatalogByName` changed approach because `<Org>` structure does not list child Catalogs anymore and they have to be retrieved using a separate call (`org.QueryCatalogList`)
* `types.Disk.Size` (allowed to provide size in KiloBytes) is no longer supported and is renamed to `types.Disk.SizeMb`
* Field `types.DiskRecordType.SizeB` is replaced with `types.DiskRecordType.SizeMb` as size in Kilobytes is not
  supported in V33.0
* Introduced Query function `client.QueryWithNotEncodedParamsHeaders` which allows to specify additional headers (used by `org.QueryOrgVdcList()` internally)
* Hopefully removed all `V32.0` conditional requirements and also specific private functions:
  * `addEmptyVmAsyncV97`
  * `addEmptyVmV97`
  * `getVMByHrefV97`
 * Added `VCDClient.QueryAllVdcs` to list all VDCs in VDC (Only using a `System` user)

Test suite configuration parameters:
* Removed `disk.size` and `disk.sizeForUpdate` as in fact they are always kept minimal and have no value.

PR #376, opened 2021-05-12
Do not send  'org' filter when used with Org users (fixes `queryCatalogByX` and `queryVDCByX`

Didainius:
After API V33.0 bump (#371) `GetCatalogById`, `GetCatalogByName`, `GetVDCById`, `GetVDCByName` started using queries behind the scenes. Apparently query filters do not accept `org` filter when being used by Org user. (it does work and is useful with System user).

This PR adjusts filtering to avoid sending `org` filter when used by Org user.


dataclouder (approved):
LGTM.
We should probably add a test that retrieves the list of VDC/Catalogs as org user. Currently, we are testing indirectly from terraform-provider-vcd.

PR #378, opened 2021-05-18
Add support for NSX Application Port Profiles

Didainius:
This PR adds types `NsxtAppPortProfile` and `types.NsxtAppPortProfile` together with `org.CreateNsxtAppPortProfile`, `org.GetAllNsxtAppPortProfiles`, `org.GetNsxtAppPortProfileByName`, `NsxtAppPortProfile.Update`, `NsxtAppPortProfile.Delete` and adds tests on top of it.

NsxtAppPortProfile uses OpenAPI endpoint to operate NSX-T Application Port Profiles
It can have 3 types of scopes:
* SYSTEM - Read-only (The ones that are provided by SYSTEM). Constant `types.ApplicationPortProfileScopeSystem`
* PROVIDER - Created by Provider on a particular network provider (NSX-T manager). Constant `types.ApplicationPortProfileScopeProvider`
* TENANT (Created by Tenant at Org VDC level). Constant `types.ApplicationPortProfileScopeTenant`

Additionally:
* Add a workaround to `OpenApiGetAllItems` internal function `openApiGetAllPages` to work around a VCD bug where some endpoints do not return `nextPage` header to follow and show `pageCount` as 0 while there are multiple pages available.
* Fix LDAP related tests to map correct port in latest versions of LDAP container
* Add optional test config variable `Misc.LdapContainer` to override default docker container location. (can be used to overcome throttling issues)
* Prevent nil pointer dereference errors in some tests when vApp is not available
* Test_GetDiskByHref supplied fake UUID of invalid length. Some API versions fail to validate it. Adjusted to include fake ID of a valid length
* Test_VdcUpdateStorageProfile used pvdcStorageProfile name in VDC storage profile. This is not always correct and was adjusted to map correctly.
* Adds two new `types.VAppStatuses` - `PARTIALLY_POWERED_OFF` and `PARTIALLY_SUSPENDED`

Note. Test suites passed on 10, 10.1, 10.2

### DIFF
--- a/govcd/nsxt_application_profile_test.go
+++ b/govcd/nsxt_application_profile_test.go
@@ -1,0 +1,182 @@
+// +build network nsxt functional openapi ALL
+
+package govcd
+
+import (
+	"fmt"
+	"net/url"
+
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+	. "gopkg.in/check.v1"
+)
+
+func (vcd *TestVCD) Test_NsxtApplicationPortProfileProvider(check *C) {
+	skipNoNsxtConfiguration(vcd, check)
+	skipOpenApiEndpointTest(vcd, check, types.OpenApiPathVersion1_0_0+types.OpenApiEndpointAppPortProfiles)
+
+	appPortProfileConfig := getAppProfileProvider(vcd, check)
+	testAppPortProfile(appPortProfileConfig, types.ApplicationPortProfileScopeProvider, vcd, check)
+
+}
+
+func (vcd *TestVCD) Test_NsxtApplicationPortProfileTenant(check *C) {
+	skipNoNsxtConfiguration(vcd, check)
+	skipOpenApiEndpointTest(vcd, check, types.OpenApiPathVersion1_0_0+types.OpenApiEndpointAppPortProfiles)
+
+	appPortProfileConfig := getAppProfileTenant(vcd, check)
+	testAppPortProfile(appPortProfileConfig, types.ApplicationPortProfileScopeTenant, vcd, check)
+
+}
+
+func (vcd *TestVCD) Test_NsxtApplicationPortProfileReadSystem(check *C) {
+	skipNoNsxtConfiguration(vcd, check)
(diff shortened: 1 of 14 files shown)


govcd/nsxt_application_profile_test.go
```diff
@@ -0,0 +1,182 @@
...
+		Scope:           types.ApplicationPortProfileScopeProvider,
+	}
+	return appPortProfileConfig
+}
+
+func getAppProfileTenant(vcd *TestVCD, check *C) *types.NsxtAppPortProfile {
+	org, err := vcd.client.GetOrgByName(vcd.config.VCD.Org)
+	check.Assert(err, IsNil)
+
+	// For PROVIDER scope application port profile must have ContextEntityId set as NSX-T Managers URN and no Org
```
lvirbalas, line 82: Just checking is the comment correct? The function seems to be for tenant.
Few lines below Description is also referring to Provider.

BTW, where does the all-capitalized form of "PROVIDER" and "TENANT" come from? Looks to be "yelling" - not sure if that's really needed here :)

lvirbalas (commented):
Just a small comment.

govcd/nsxt_application_profile_test.go
Didainius: I probably have overused capitalization :). These are `scope`values being set in API.

Basically there are 3 scopes `system`, `provider` and `tenant`

* `system` - read-only (built-in application profiles that are provided by `system` scope)
* `provider` - can be created and filtered by System user on a particular network provider (NSX-T manager). This means that whatever is set - should be available to all tenants using the same NSX-T manager
* `tenant` can be created at tenant level (org VDC)

lvirbalas: Please make a decision on this together with @dataclouder

lvirbalas (approved):
I have left that one conversation open for when @dataclouder is reviewing this PR. LGTM, otherwise!

dataclouder (approved):

govcd/nsxt_application_profile_test.go
dataclouder: If the API demands uppercase, I prefer keeping the original value in the comments and docs

PR #382, opened 2021-06-09
Add NSX-T NAT rule support

Didainius:
Implements NSX-T NAT rule support by adding `NsxtNatRule` and `types.NsxtNatRule` as well as methods `edge.GetAllNsxtNatRules`,`edge.GetNsxtNatRuleByName`, `edge.GetNsxtNatRuleById`, `edge.CreateNatRule`, `nsxtNatRule.Update`, `nsxtNatRule.Delete`, `nsxtNatRule.IsEqualTo`

A few caveats:
* API does not return ID of created object therefore it must be manually looked up by comparing as much fields as it is possible. To aid this `nsxtNatRule.IsEqualTo` is introduced to perform additional validations. `edge.CreateNatRule` also has non-standard task tracking and then uses `nsxtNatRule.IsEqualTo` internally to find ID of newly created rule.
* There are two new fields introduced in API V35.2+ (VCD 10.2.2+) - `FirewalMatch` and `Priority`. In order to support these fields all methods above will elevevate API version to exactly V35.2 (not higher) so that those fields can be used. There is a specific check `Test_NsxtNatDnatFirewallMatchPriority` to test exactly that
* New NAT rule type `REFLEXIVE` is supported in API V36.0. It also requires rule type to be specified in `Type` field instead of `RuleType`
* There is one field which was deprecated `InternalPort` and was changed to `DnatExternalPort` between existing versions. Because these functions automatically elevate API - one must know which field to use. Tests `Test_NsxtNatDnatInternalPort` and `Test_NsxtNatDnatExternalPort` test out this functionality.

Additionally:
* Functions `queryOrgVdcByName`, `queryOrgVdcById`,  `queryCatalogByName` and `queryCatalogById` are improved to include `orgName` as filter (improves filtering accuracy in cases when `org` ID field is not accepted)

Note. Tests pass with `nsxt` tag on all NSX-T supported environments.

### DIFF
--- a/govcd/nsxt_nat_rule.go
+++ b/govcd/nsxt_nat_rule.go
@@ -38,10 +38,7 @@ func (egw *NsxtEdgeGateway) GetAllNatRules(queryParameters url.Values) ([]*NsxtN
 		return nil, err
 	}
 
-	// Fields FirewallMatch and Priority require API version 35.2 to be set therefore version is elevated if API supports
-	if client.APIVCDMaxVersionIs(">= 35.2") {
-		apiVersion = "35.2"
-	}
+	apiVersion = elevateNsxtNatRuleApiVersion(apiVersion, client)
 
 	urlRef, err := client.OpenApiBuildEndpoint(fmt.Sprintf(endpoint, egw.EdgeGateway.ID))
 	if err != nil {
@@ -128,10 +125,7 @@ func (egw *NsxtEdgeGateway) CreateNatRule(natRuleConfig *types.NsxtNatRule) (*Ns
 		return nil, err
 	}
 
-	// Fields FirewallMatch and Priority require API version 35.2 to be set therefore version is elevated if API supports
-	if client.APIVCDMaxVersionIs(">= 35.2") {
-		apiVersion = "35.2"
-	}
+	apiVersion = elevateNsxtNatRuleApiVersion(apiVersion, client)
 
 	// Insert Edge Gateway ID into endpoint path edgeGateways/%s/nat/rules
 	urlRef, err := client.OpenApiBuildEndpoint(fmt.Sprintf(endpoint, egw.EdgeGateway.ID))
@@ -175,10 +169,7 @@ func (nsxtNat *NsxtNatRule) Update(natRuleConfig *types.NsxtNatRule) (*NsxtNatRu
 		return nil, err
 	}
 
-	// Fields FirewallMatch and Priority require API version 35.2 to be set therefore version is elevated if API supports
-	if client.APIVCDMaxVersionIs(">= 35.2") {
-		apiVersion = "35.2"
-	}
+	apiVersion = elevateNsxtNatRuleApiVersion(apiVersion, client)
 
 	if nsxtNat.NsxtNatRule.ID == "" {
 		return nil, fmt.Errorf("cannot update NSX-T NAT Rule without ID")
@@ -212,10 +203,7 @@ func (nsxtNat *NsxtNatRule) Delete() error {
 		return err
 	}
 
-	// Fields FirewallMatch and Priority require API version 35.2 to be set therefore version is elevated if API supports
-	if client.APIVCDMaxVersionIs(">= 35.2") {
-		apiVersion = "35.2"
-	}
+	apiVersion = elevateNsxtNatRuleApiVersion(apiVersion, client)
 
 	if nsxtNat.NsxtNatRule.ID == "" {
 		return fmt.Errorf("cannot delete NSX-T NAT rule without ID")
@@ -277,4 +265,22 @@ func natRulesEqual(first, second *types.NsxtNatRule) bool {
 	}
 
 	return false
+}
+
+// elevateNsxtNatRuleApiVersion helps to elevate API version to consume newer NSX-T NAT Rule features
+// API V35.2+ support new fields FirewallMatch and Priority
+// API V36.0+ supports new RuleType - REFLEXIVE
+func elevateNsxtNatRuleApiVersion(apiVersion string, client *Client) string {
+
+	// Fields FirewallMatch and Priority require API version 35.2 to be set therefore version is elevated if API supports
+	if client.APIVCDMaxVersionIs(">= 35.2") {
+		apiVersion = "35.2"
+	}
+
+	// RuleType REFLEXIVE requires API V36.0
+	if client.APIVCDMaxVersionIs(">= 36.0") {
+		apiVersion = "36.0"
+	}
+
+	return apiVersion
 }
--- a/govcd/nsxt_nat_rule.go
+++ b/govcd/nsxt_nat_rule.go
@@ -12,7 +12,8 @@ import (
 	"github.com/vmware/go-vcloud-director/v2/util"
 )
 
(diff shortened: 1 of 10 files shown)


lvirbalas (approved):

govcd/nsxt_nat_rule.go
```diff
@@ -0,0 +1,289 @@
...
+	if err != nil {
+		return fmt.Errorf("error deleting NSX-T NAT Rule: %s", err)
+	}
+
+	return nil
+}
+
+// IsEqualTo allows to check if a rule has exactly the same fields (except ID) to the supplied rule
+// This validation is very tricky because minor version changes impact how fields are return.
+// This function relies on most common and stable fields:
```
dataclouder, line 231: I understand that `Type` and `RuleType` may not both be there, but a comparison should be still viable. What am I missing?

Also, could you make a note on why these fields are not reliably comparable?

* InternalPort,
* FirewallMatch,
* Priority,      
* Version,

govcd/nsxt_nat_rule.go
```diff
@@ -0,0 +1,289 @@
...
+	if first.Name == second.Name &&
+		first.Enabled == second.Enabled &&
+		first.Description == second.Description &&
+
+		first.ExternalAddresses == second.ExternalAddresses &&
+		first.InternalAddresses == second.InternalAddresses &&
+		// Match either both application profiles being empty/nil, or both having the same value
+		((first.ApplicationPortProfile == second.ApplicationPortProfile) ||
+			(first.ApplicationPortProfile != nil && second.ApplicationPortProfile != nil && first.ApplicationPortProfile.ID == second.ApplicationPortProfile.ID)) {
+
```
dataclouder, line 266: I think the match should be valid if one rule has `ApplicationPortProfile == nil` and the second is empty (i.e.  `ApplicationPortProfile.ID == ""`)

govcd/nsxt_nat_rule.go
```diff
@@ -0,0 +1,289 @@
...
+	if client.APIVCDMaxVersionIs(">= 35.2") {
+		apiVersion = "35.2"
+	}
+
+	// RuleType REFLEXIVE requires API V36.0
+	if client.APIVCDMaxVersionIs(">= 36.0") {
+		apiVersion = "36.0"
+	}
+
+	return apiVersion
```
dataclouder, line 288: This function assumes that `apiVersion` contains the valid version to be used when the client version is < 35.2.
But if `apiVersion` is empty or invalid, we will have an unrecoverable error.
Possibly we should retrieve apiVersion within `elevateNsxtNatRuleApiVersion` and return `(string error)`. 
Or you may have a better idea.

govcd/nsxt_nat_rule_test.go
```diff
@@ -0,0 +1,354 @@
...
+
+	appPortProfiles, err := org.GetAllNsxtAppPortProfiles(nil, types.ApplicationPortProfileScopeSystem)
+	check.Assert(err, IsNil)
+
+	natRuleDefinition := &types.NsxtNatRule{
+		Name:              check.TestName() + "dnat",
+		Description:       "description",
+		Enabled:           true,
+		RuleType:          types.NsxtNatRuleTypeDnat,
+		ExternalAddresses: edge.EdgeGateway.EdgeGatewayUplinks[0].Subnets.Values[0].PrimaryIP,
```
dataclouder, line 33: Is there any chance that a component of the path down to `PrimaryIP` could be `nil`?

dataclouder (changes requested):
Some questions to answer

govcd/nsxt_nat_rule.go
```diff
@@ -0,0 +1,289 @@
...
+
+// natRulesEqual is a helper to check if first and second supplied rules are exactly the same (except ID)
+func natRulesEqual(first, second *types.NsxtNatRule) bool {
+	util.Logger.Println("comparing NAT rule:")
+	util.Logger.Printf("%+v\n", first)
+	util.Logger.Println("against:")
+	util.Logger.Printf("%+v\n", second)
+
+	// Being an org user always returns logging as false - therefore cannot compare it.
+	// first.Logging == second.Logging &&
```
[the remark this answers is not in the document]
Didainius: Fixed.

govcd/nsxt_nat_rule.go
Didainius: >Also, could you make a note on why these fields are not reliably comparable?

Good question which led me to another round of thinking. I was carefully trying to match as many fields as possible (while still having comprehensible comparison code). Main problem for many fields is - behavior changes across API versions which can make this matching very messy - value comparisons differ based on version:

* `Type` and `RuleType` - the trick is - on API V36.0 one can set `RuleType` (deprecated) or `Type`  field value (except the new `REFLEXIVE` which explicitly requires `Type` field). After READ it would return both `RuleType` and `Type` being set with the same value, _unless_ it is `REFLEXIVE`(in that case it returns `ruleType=null` and `type=REFLEXIVE`)
* FirewallMatch - it exists only since API 35.2+ and has a default starting this version
* InternalPort - is deprecated since API V35.0+ and is replaced by DnatExternalPort
* Priority - is available only in API V35.2+
* Version - it is something that is automatically handled by API. When creating - you must specify none, but it sets version to 0. When updating one must specify the last version read, and again it will automatically increment this value after update. (probably it is meant to avoid concurrent updates)

We could probably even write smart rules for tested versions, but what I am more afraid of is that even using a known API version might break in future VCD versions as we already saw a few times before. I know that in future versions NAT should return ID after POST and that should not be required at all, but now we still have it.

For now I have put this comment as well into function docs.

Any other ideas? I will sleep on it to see if anything more pops up now as I refreshed this context.

govcd/nsxt_nat_rule.go
Didainius: The point with this function was that I would always feed it with output of 
```
endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointNsxtNatRules
apiVersion, err := client.checkOpenApiEndpointCompatibility(endpoint)
```

However. I had in mind that one day - if such approach of "stepped" elevation (elevation to known verified API versions when possible) could be more generic for those resources that need it. 

I have made a second attempt and introduced a function `client.getOpenApiHighestElevatedVersion` for internal use which is similar to `checkOpenApiEndpointCompatibility`, but is capable of picking highest possible version from a predefined list `endpointElevatedApiVersions`. 

We could just use "highest" version supported by VCD API, but that is risky as pointing it to future VCD version might just break it (if some field is removed in future API version).

Have a look and share if it looks better. This is yet hard to comprehend if it is a silver bullet as there is just one use case, but this was my early vision of how it could look in a more generic fashion.

govcd/nsxt_nat_rule_test.go
Didainius: It should not.
What happens is that Edge Gateway requires IP addresses (at least 1) from external network. It does it set as `primary_ip` by default.


Here is a screenshot from fresh created NSX-T Edge Gateway

image

Now I have added additional explicit safeguard (validation) that it is not empty. I have never seen it, but at least tests should warn early if that changes in future.

govcd/nsxt_nat_rule.go
Didainius: Possibly. I have added two more checks to validate from both sides.

dataclouder (approved):

PR #383, opened 2021-06-11
Add disk sharing functionality (v35.0) and UUID (v36.0)

arunmk:
## IMPORTANT

guidelines shown below. 

Created a PR without an issue after discussion with the contributors. The description should make it self-evident.

## Description

Related issue: (`<URL or #NUMBER of your Issue>`)

- This PR adds basic support to create shareable disks. That is done by essentially adding the `Shareable` field to the Disk structure. Additional other fields such as Encrypted and SharingType, which were added in v35 have also been pulled in. These new fields have been added in v35. If these are not present, the default value is false, i.e., the disks are not shared. If `Shareable` is set to `True`, the disk can be attached to multiple VMs. That functionality is provided in VCD 10.2 and does not need any other API change in order to attach to multiple disks. Note that all of the parameters mentioned here are optional, and sensible defaults are mentioned in VCD API to keep them backward compatible. Hence they are specified as `omitempty` in the API.

- Another UUID parameter was added in v36.0 to help discover the disk that was attached. When a disk is attached to a VM, there is no way to know which of the /dev/sd* devices correspond to the device being attached. However, there is a UUID parameter for each disk available from v36. When the command `/lib/udev/scsi_id --page=0x83 --whitelisted --device=</dev/sdX>` is run on the VM host, there is a UUID emitted that can be checked against the UUID obtained from the API, and hence the right disk can be discovered. (Note that for the host command to work, the extra option `disk.enableUUID` must be set on the VM using vmomi).Once the right device /dev/sdx is discovered, the disk can be mounted etc. This parameter is only returned from VCD and hence there is no impact to the callers apart from the additional field in the struct.

- (Optional) @Didainius could you please take a look?

### DIFF
--- a/.changes/v2.15.0/383-features.md
+++ b/.changes/v2.15.0/383-features.md
@@ -1,0 +1,4 @@
+* Added support for Shareable disks, i.e., independent disks that can be attached to multiple VMs which is available from
+  API v35.0 onwards. Also added uuid to the Disk structure which is a new member that is returned from v36.0 onwards. This
+  member holds a UUID that can be used to correlate the disk that is attached to a particular VM from the VCD side and the
+  VM host side. [GH-383]
--- a/types/v56/types.go
+++ b/types/v56/types.go
@@ -2654,6 +2654,10 @@ type Disk struct {
 	Iops            *int             `xml:"iops,attr,omitempty"`
 	BusType         string           `xml:"busType,attr,omitempty"`
 	BusSubType      string           `xml:"busSubType,attr,omitempty"`
+	Encrypted       bool             `xml:"encrypted,attr,omitempty"`
+	Shareable       bool             `xml:"shareable,attr,omitempty"`
+	SharingType     string           `xml:"sharingType,attr,omitempty"`
+	UUID            string           `xml:"uuid,attr,omitempty"`
 	Description     string           `xml:"Description,omitempty"`
 	Files           *FilesList       `xml:"Files,omitempty"`
 	Link            []*Link          `xml:"Link,omitempty"`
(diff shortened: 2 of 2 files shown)


Didainius (commented):
Sounds quite clear - just if you could improve description of this PR and also add a few lines to `CHANGELOG.md`.

lvirbalas (approved):
I'm approving in advance, but as requested by @Didainius we need a proper PR description (which will be commit message too) and an entry in the changelog.

Didainius (approved):

dataclouder (changes requested):
I think that we should not add these fields without any function that supports them or tests to make sure the addition doesn't break anything

lvirbalas:
CSE

CHANGELOG.md
```diff
@@ -81,6 +81,10 @@ IMPROVEMENTS:
   pagination [#378](https://github.com/vmware/go-vcloud-director/pull/378)
 * Improved LDAP container related tests to use correct port mapping for latest LDAP container version 
   [#378](https://github.com/vmware/go-vcloud-director/pull/378)
+* Added support for Shareable disks, i.e., independent disks that can be attached to multiple VMs which is available from
```
vbauzysvmware, line 84: Could you move this changelog change to `.changes/v2.15.0` In this folder, we create the file with PR number and store the changelog message.

vbauzysvmware (changes requested):
Last change and we merge this :)

CHANGELOG.md
arunmk: done @vbauzysvmware . please take a look at the second commit. (I had to force-push since my branch was quite outdated)

vbauzysvmware: thank you @arunmk

vbauzysvmware (approved):

PR #389, opened 2021-07-02
V2.12.0 regression fix for `org.GetCatalogByName` and `org.GetCatalogById`

Didainius:
go vcloud director does not find catalogs shared from other Orgs in methods `org.GetCatalogByName` and `org.GetCatalogById`

This PR fixes that, adds some tests including a test with uses Org Admin user and adds a changelog note for `2.12.1` bugfix release.


dataclouder (approved):

lvirbalas (approved):
Thanks for a quick turnaround!

PR #391, opened 2021-07-05
Utility scripts

dataclouder:
###  Add .changes folder 

Add a folder `.changes' and `scripts/make-changelog.sh` to avoid `CHANGELOG` confilcts .
This is similar to the organization of terraform-provider-vcd.
To re-assemble the CHANGELOG, use `./scripts-make-changelog.sh VERSION`.

To show how the CHANGELOG would be organized, I added the items for v2.12.0 and v2.12.1.

###  Allow staticcheck to run in GitHub Actions

The test with `staticcheck` had stopped working when we transitioned from Travis to GitHub Actions. Changing the trigger variable fixed it. I also added a skip comment for a function that is used on demand for debugging.

### DIFF
--- a/.changes/README.md
+++ b/.changes/README.md
@@ -1,0 +1,30 @@
+# Handling of CHANGELOG entries
+
+## PR operations
+
+* All PRs, instead of updating CHANGELOG.md directly, will create individual files in a directory .changes/$version 
+* The files will be named `xxx-section_name.md`, where xxx is the PR number, and `section_name` is one of 
+    * features
+    * improvements
+    * bug-fixes
+    * deprecations
+    * notes
+    * removals
+
+* The changes files must NOT contain the header
+
+* You can update the file `.changes/sections` to add more headers
+* To see the full change set for the current version (or an old one), use `./scripts/make-changelog.sh [version]`
+
+
+## Post release initialization
+   
+After a release, the changelog will be initialized with the following template:
+ 
+```
+    ## $VERSION (Unreleased)
+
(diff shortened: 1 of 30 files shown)


lvirbalas (commented):
What are the main differences from the way we generate changelog in Terraform?
BTW, there are typos in the PR description: "terraform-prov0wider-vcd. To re-assemple"

dataclouder:
> What are the main differences from the way we generate changelog in Terraform?
> BTW, there are typos in the PR description: "terraform-prov0wider-vcd. To re-assemple"

The main differences:
* We use a single block CHANGELOG.md that causes conflicts at almost every merge
* In terraform-provider-vcd we use headers for all items in the CHANGELOG, while in go-vcloud-director we only use "breaking changes" and improvements.

This PR aims at having both CHANGELOGs follow the same pattern.

lvirbalas (approved):
LGTM!

Only one thing: I see in the generated CHANGELOG extra new lines in between the groups of changes. Is the idea to reconcile them manually before the release? This could be human-error-prone.

dataclouder:
> I see in the generated CHANGELOG extra new lines in between the groups of changes. Is the idea to reconcile them manually before the release? This could be human-error-prone.

We can make a "perfect and final" assembly tool, but it would require using something better than shell script. The goal here is to make a good-enough draft that we can adapt manually before releasing.  We have been doing the same in terraform-provider-vcd.

Didainius (approved):
Thanks! A brief test worked.

Issue #390, opened 2021-07-03
"func (catalog *Catalog) Delete(force, recursive bool) error" ignores task and return immediatelly


Didainius:
Catalog deletion function does not take into account returned task and responds immediately while catalog deletion task is running and may even fail.

This needs to be fixed in order to properly track catalog deletion and catch any errors.

PR #392, opened 2021-07-05
Fix catalog delete

dataclouder:
Fix Issue #390: `catalog.Delete()` ignores task

Note: the file in `.changes/...` will be processed correctly after PR #391 is merged.

### DIFF
--- a/.changes/v2.13.0/392-bug-fixes.md
+++ b/.changes/v2.13.0/392-bug-fixes.md
@@ -1,2 +1,2 @@
-* Fix Issue #390: `catalog.Delete()` ignores task [GH-392]
+* Fix Issue #390: `catalog.Delete()` ignores returned task and responds immediately which could have caused failures [GH-392]
 
--- a/.changes/v2.13.0/392-bug-fixes.md
+++ b/.changes/v2.13.0/392-bug-fixes.md
@@ -1,2 +1,2 @@
-* Fix Issue #390: `catalog.Delete()` ignores task
+* Fix Issue #390: `catalog.Delete()` ignores task [GH-392]
 
--- a/.changes/v2.13.0/392-bug-fixes.md
+++ b/.changes/v2.13.0/392-bug-fixes.md
@@ -1,0 +1,2 @@
+* Fix Issue #390: `catalog.Delete()` ignores task
+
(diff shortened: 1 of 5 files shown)


govcd/catalog.go
```diff
@@ -61,13 +61,18 @@ func (catalog *Catalog) Delete(force, recursive bool) error {
...
+		return fmt.Errorf("error deleting Catalog %s: %s", catalog.Catalog.Name, err)
 	}
-
-	return nil
+	task := NewTask(catalog.client)
+	if err = decodeBody(types.BodyTypeXML, resp, task.Task); err != nil {
+		return fmt.Errorf("error decoding task response: %s", err)
+	}
+	if task.Task.Status == "error" {
+		return fmt.Errorf("catalog not properly destroyed")
```
Didainius, line 73: Is there a possible error field we could throw?

Didainius (approved):
Great fix! Might have led to some accidental errors for bundles of destroy operations.

govcd/catalog.go
dataclouder: Yes. I have added a more in formative message and enhanced `Test_DeleteCatalog` to make sure that the inner error message comes to the surface.

.changes/v2.13.0/392-bug-fixes.md
```diff
@@ -0,0 +1,2 @@
+* Fix Issue #390: `catalog.Delete()` ignores task [GH-392]
```
lvirbalas, line 1: Please add a bit more explanation of what are fixing here, so for a new reader it's clear from this line.

dataclouder: Isn't the reference to the issue – where the problem is explained in detail – not sufficient?

lvirbalas: I want to make it instantly clear for the users without the need to click.

Suggestion:

"Fix Issue #390: `catalog.Delete()` ignores returned task and responds immediately which could have caused failures [GH-392]

Inspired by:

"Catalog deletion function does not take into account returned task and responds immediately while catalog deletion task is running and may even fail.

This needs to be fixed in order to properly track catalog deletion and catch any errors."

dataclouder: Changed

lvirbalas (approved):

PR #399, opened 2021-09-15
Add support External Network backed NSX-T Segment

Didainius:
VCD 10.3 introduced a new type of NSX-T External networks - Segment backed networks which can then be consumed by Direct VDC network. This PR adds support for this type of network and related tests.

### DIFF
--- a/.changes/v2.13.0/339-improvements.md
+++ b/.changes/v2.13.0/339-improvements.md
@@ -1,3 +1,4 @@
-* External network type ExternalNetworkV2 uses elevated API (maximum available from 33.0, 35.0 and 36.0) [GH-339]
+* External network type ExternalNetworkV2 automatically elevates API version to maximum available from 33.0, 35.0 and
+  36.0 so that new functionality can be consumed [GH-339]
 * Added new field BackingTypeValue in favor of deprecated BackingType to types.ExternalNetworkV2Backing [GH-339]
 * Add new function `GetFilteredNsxtImportableSwitches` to query NSX-T Importable Switches (Segments) [GH-339] 
--- a/.changes/v2.13.0/339-improvements.md
+++ b/.changes/v2.13.0/339-improvements.md
@@ -1,4 +1,6 @@
-* External network type ExternalNetworkV2 automatically elevates API version to maximum available from 33.0, 35.0 and
-  36.0 so that new functionality can be consumed [GH-339]
+* External network type ExternalNetworkV2 automatically elevates API version to maximum available out of 33.0, 35.0 and
+  36.0, so that new functionality can be consumed. It uses a controlled version elevation mechanism to consume the newer
+  features, but at the same time remain tested by not choosing the latest untested version blindly (more information in
+  openapi_endpoints.go) [GH-339]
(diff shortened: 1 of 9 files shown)


.changes/v2.13.0/339-improvements.md
```diff
@@ -0,0 +1,3 @@
+* External network type ExternalNetworkV2 uses elevated API (maximum available from 33.0, 35.0 and 36.0) [GH-339]
```
lvirbalas, line 1: This sentence is geeky :) Could we somehow make it more friendly for the casual reader? Few suggestions:
- "elevated API ..." - maybe add "version"?
- What about explaining what "elevated" means in this context in more detail?
Thanks!

lvirbalas (commented):
Just one ask for the "elevated API" concept description.

.changes/v2.13.0/339-improvements.md
Didainius: I have tried to improve the message. How does it look now?

One thing I am not sure about is if we should put any internal links to docs. I could - but we never done this before I think.

We introduced this generic functionality for API version elevation in previous release so that NSX-T Nat Rules can consume latest API.

There is a function https://github.com/Didainius/go-vcloud-director/blob/faf5992f7e4f4338cc76afff924cdc650af2409c/govcd/openapi_endpoints.go#L107 which is used in our resources and some elevations with explanations:
https://github.com/Didainius/go-vcloud-director/blob/faf5992f7e4f4338cc76afff924cdc650af2409c/govcd/openapi_endpoints.go#L59-L71

govcd/external_network_v2_test.go
```diff
@@ -158,17 +184,21 @@ func testExternalNetworkV2(name, backingType, backingId, NetworkProviderId strin
...
 				NetworkProvider: types.NetworkProvider{
-					// Name: vcd.config.Nsxt.Manager,
 					ID: NetworkProviderId,
 				},
 			},
 		}},
 	}
 
-	return neT
+	// Starting with VCD 10.2 field BackingType is deprecated in favor of BackingTypeValue, and only it accepts new values
```
dataclouder, line 194:
```suggestion
	// Starting with VCD 10.2 field BackingType is deprecated in favor of BackingTypeValue, and it only accepts new values
```

govcd/nsxt_importable_switch.go
```diff
@@ -65,30 +65,76 @@ func (vdc *Vdc) GetAllNsxtImportableSwitches() ([]*NsxtImportableSwitch, error)
...
 	if err != nil {
 		return nil, fmt.Errorf("could not get UUID from URN '%s': %s", vdc.Vdc.ID, err)
 	}
+	filter := map[string]string{"orgVdc": orgVdcId}
+
+	return getFilteredNsxtImportableSwitches(filter, vdc.client)
+}
+
+// GetFilteredNsxtImportableSwitches returns all available importable switches.
+// One of filters is required (requires plain UUID - not URN):
```
dataclouder, line 79:
```suggestion
// One of the filters below is required (using plain UUID - not URN):
```

dataclouder (changes requested):

dataclouder (approved):

lvirbalas (approved):

PR #409, opened 2021-11-19
Add session info to govcd logs

dataclouder:
This PR adds detailed session information to the logs.
The information recorded is:

* VCD version
* User name
* Org name
* Roles used by the connection
* Rights associated with all roles in the connection
* Connection type (using password/token/API token)

The above data gets written even if the connection uses a token

No input is required from users: if  logging is enabled, the session info gets written into the current log.

Once this PR is merged, terraform-provider-vcd will automatically benefit from it.

You can get the session information from the log using:

```
awk '/START SESSION/, /END SESSION/' go-vcloud-director.log
```


vbauzysvmware (approved):

lvirbalas (approved):
Thanks!

Didainius (approved):

PR #411, opened 2021-11-26
Add support for NSX-T ALB Pools

Didainius:
This PR adds support for NSX-T ALB Pools via type NsxtAlbPool and its relevant functions.

* Moves embedded certificates to separate file so that they can be shared with multiple tests (tags)


Didainius:
Closed in favor of #414

PR #412, opened 2021-12-01
Fix broken request payload logging

Didainius:
While fixing unhandled errors we accidentally masked one variable. More details in https://github.com/vmware/go-vcloud-director/pull/412/files#r759886479

### DIFF
--- a/govcd/openapi.go
+++ b/govcd/openapi.go
@@ -634,8 +634,9 @@ func (client *Client) newOpenApiRequest(apiVersion string, params url.Values, me
 	// If the body contains data - try to read all contents for logging and re-create another
 	// io.Reader with all contents to use it down the line
 	var readBody []byte
+	var err error
 	if body != nil {
-		readBody, err := ioutil.ReadAll(body)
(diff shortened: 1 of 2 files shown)


govcd/openapi.go
```diff
@@ -634,8 +634,9 @@ func (client *Client) newOpenApiRequest(apiVersion string, params url.Values, me
 	// If the body contains data - try to read all contents for logging and re-create another
 	// io.Reader with all contents to use it down the line
 	var readBody []byte
+	var err error
 	if body != nil {
-		readBody, err := ioutil.ReadAll(body)
+		readBody, err = ioutil.ReadAll(body)
```
Didainius, line 639: Previously we initialized `readBody` withing `if` block scope while the one defined in line 636 remained empty. Then at line 683 we actually used that empty initialized `readBody` to set it as payload and logged it in line 685.

vbauzysvmware (approved):

lvirbalas (approved):

dataclouder (approved):

PR #413, opened 2021-12-01
Add VM Shutdown support

lelvisl:
Shutdown triggers a VM undeploy and shutdown action. "Shut Down Guest OS" action in UI behaves this way.


lelvisl:
@Didainius please look at this.

Didainius:
> @Didainius please look at this.

Sorry @lelvisl ,
Will make my attempt today.

Didainius:
Ok, so this makes sense and and I checked the payload of "Shutdown VM" and it is as you say. Probably my points are:
* We need to test this (I can help here as I am not sure you are able to run the test).
* Do you what which vApp status does it get to? Is it `8:  "POWERED_OFF",` after success?
* Do you know what happens if there are no guest tools running - does it fail in the end?

Out of curiosity - I think you mentioned some time ago already, but are you using the SDK separatelly (not via Terraform)?

lelvisl:
- **tests:** i think i can test this, but i need to know - what do you want to test? Example - shutdown vm and get status of vm after this (vm with installed tools)
- **status:** this method on VM and vm will get status POWERED_OFF, vapp may be "POWERED_OFF"  or "PARTIALLY_POWERED_OFF" (may be "MIXED"), you can read here - https://blogs.vmware.com/vsphere/2012/04/partially-powered-off-vapp-vms.html
- **no guest tools:** yes, it fail in the end. 


And yes, i'm using the SDK separatelly.

.changes/v2.14.0/413-features.md
```diff
@@ -0,0 +1,2 @@
+* Added method `VM.Shutdown` to shut down guest os
```
lvirbalas, line 1:
```suggestion
* Added method `VM.Shutdown` to shut down guest OS
```

lvirbalas (changes requested):
One nit and please add tests.

Didainius:
@lelvisl,
We'd like to move this forward and bring in to place.

Would you be able to merge with latest `main` branch and make the test. IMO the test itself should be based on VM status before/after as there's not much to do.

Didainius:
@lelvisl I am closing this one and open to your comments in #496 which adds tests.

PR #415, opened 2021-12-12
Change refresh_token to use request body

dataclouder:
Although the operations in go-vcloud-director are logged
with the sensitive information filtered off, when using
the API token in the URL, it gets recorded in the VCD
logs. To avoid this logging, which may result in
token disclosure when a support bundle is generated,
we change the refresh_token operation to use the request body
instead of URL parameters.

### DIFF
--- a/govcd/api_token.go
+++ b/govcd/api_token.go
@@ -5,6 +5,7 @@
 package govcd
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


vbauzysvmware (approved):
thank you.
Is 
```
// sanitizedUrl removes sensitive API token from URL
func sanitizedUrl(url string) string {
```
Still needed?

Didainius (approved):

lvirbalas (approved):

dataclouder:
> thank you. Is
> 
> ```
> // sanitizedUrl removes sensitive API token from URL
> func sanitizedUrl(url string) string {
> ```
> 
> Still needed?

No. It is not needed anymore. I removed it

lelvisl:
Why not use url.Values from golang std lib? 
Example: https://go.dev/play/p/SRmq7Ckn3-n

dataclouder:
> Why not use url.Values from golang std lib? Example: https://go.dev/play/p/SRmq7Ckn3-n

Because it's set in legacy code. Changing igt would require rewriting most of the API functions, and that would break compatibility.
In our handling of openAPI operations we are using url.Values

PR #418, opened 2021-12-16
Fix issue when internal disk update would clean VM description

vbauzysvmware:
[Ref: #728](https://github.com/vmware/terraform-provider-vcd/issues/728)

This fix solves the issue when in terraform VM and internal disk resources are used and internal disk update happens (like resize). So on the second terrafom apply VM description is missing and an update on VM is initiated. Also which currently requires VM power off and power on.


lvirbalas:
Looks like this PR contains other PRs, hence is not reviewable in this state.

vbauzysvmware:
Ready for review

.changes/v2.14.0/418-bug-fixes.md
```diff
@@ -0,0 +1,2 @@
+* Fix Issue #728: `vm.UpdateInternalDisksAsync()` didn't send VM description and as result would delete it [GH-418]
```
lvirbalas, line 1:
```suggestion
* Fix Issue #728: `vm.UpdateInternalDisksAsync()` didn't send VM description and as a result would delete it [GH-418]
```

lvirbalas (commented):
Interesting. Are there any other fields that we are not sending during disk update, which could have a similar effect?

.changes/v2.14.0/418-bug-fixes.md
vbauzysvmware: updated

vbauzysvmware:
> Interesting. Are there any other fields that we are not sending during disk update, which could have a similar effect?

I am not aware of any anymore.

lvirbalas (approved):

govcd/vm_test.go
```diff
@@ -1087,6 +1091,11 @@ func (vcd *TestVCD) Test_UpdateInternalDisk(check *C) {
 	check.Assert(disk.BusNumber, Equals, diskSettings.BusNumber)
 	check.Assert(disk.AdapterType, Equals, diskSettings.AdapterType)
 
+	// verify VM description still available - bug fix test
```
lvirbalas, line 1094:
```suggestion
	// verify VM description still available - test for bugfix #418
```

govcd/vm_test.go
```diff
@@ -1052,6 +1052,10 @@ func (vcd *TestVCD) Test_UpdateInternalDisk(check *C) {
 	vm, storageProfile, diskSettings, diskId, previousProvisioningValue, err := vcd.createInternalDisk(check, vmName, 1)
 	check.Assert(err, IsNil)
 
+	description := "Test_UpdateInternalDisk_Description"
```
lvirbalas, line 1055: Please also add a comment above of why this has to be done explicitly.

.changes/v2.14.0/418-bug-fixes.md
dataclouder, line 1: It's not clear what "it" refers to. Also, from the conversation in Issue #728, it doesn't seem that this fix addresses that problem.

govcd/vm_test.go
vbauzysvmware: ty. changed

govcd/vm_test.go
vbauzysvmware: added

.changes/v2.14.0/418-bug-fixes.md
vbauzysvmware: yes, it is not 100% as I had very limited information

vbauzysvmware: updated "it"

dataclouder (approved):

Didainius (approved):
Tests passed. Go ahead.

PR #419, opened 2021-12-22
bugfix: Do not send Weight in ChangeCPUCountWithCore

Didainius:
`ChangeCPUCountWithCore` sends a hardocded `Weight` field with value 0. This might override shares in some cases. This PR adjusts `ChangeCPUCountWithCore` function to avoid sending Weight field at all.

### DIFF
--- a/types/v56/types.go
+++ b/types/v56/types.go
@@ -1597,9 +1597,10 @@ type OVFItem struct {
 	Reservation     int      `xml:"rasd:Reservation"`
 	ResourceType    int      `xml:"rasd:ResourceType"`
 	VirtualQuantity int64    `xml:"rasd:VirtualQuantity"`
-	Weight          int      `xml:"rasd:Weight,omitempty"`
-	CoresPerSocket  *int     `xml:"vmw:CoresPerSocket,omitempty"`
-	Link            *Link    `xml:"vcloud:Link"`
+	// Weight corresponds to Shares when used for CPU and/or memory settings
+	Weight         int   `xml:"rasd:Weight,omitempty"`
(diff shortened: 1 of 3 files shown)


lvirbalas (commented):
Please remind me, do the versions before 10.3.2 work with this change?

Didainius:
> Please remind me, do the versions before 10.3.2 work with this change?

They do. The point is that all previous VCD versions ignored (because of a bug) `Weight` field which was hardcoded  to be `0` in `ChangeCPUCountWithCore` function. In the latest version it actually starts accepting this field. 

Note. The `Weight` here actually stands for CPU Shares.

vbauzysvmware (approved):

types/v56/types.go
```diff
@@ -1597,7 +1597,7 @@ type OVFItem struct {
 	Reservation     int      `xml:"rasd:Reservation"`
 	ResourceType    int      `xml:"rasd:ResourceType"`
 	VirtualQuantity int64    `xml:"rasd:VirtualQuantity"`
-	Weight          int      `xml:"rasd:Weight"`
+	Weight          int      `xml:"rasd:Weight,omitempty"`
```
vbauzysvmware, line 1600: maybe worth adding comment that Weight is "shares" ?

vbauzysvmware (approved):

lvirbalas (approved):
Great, thank you!

types/v56/types.go
lvirbalas: Good suggestion, indeed 🙂

dataclouder (approved):

types/v56/types.go
Didainius: Added.

PR #422, opened 2022-01-05
Upload OVF by URL

vbauzysvmware:
Ref: https://github.com/vmware/go-vcloud-director/issues/321

Add function to upload OVF by link. Only OVF files are supported.
Added support for updating vApp template `vAppTemplate.UpdateAsync` and `vAppTemplate.Update`

PR #426, opened 2022-01-18
fix catalog cleanup in tests

vbauzysvmware:
After new tests were added, the old test started to leave the vApp template in env. This fix solves that.


Didainius (approved):
Two things:
* Title could reflect the issue better "fix catalog cleanup in tests"
* Possibly adding this issue number to catalog PR #422

lvirbalas (approved):

PR #441, opened 2022-03-03
Catalog.QueryMediaList bugfix

mikeletux:
## Description
`Catalog.QueryMediaList` method has a bug which prevents from correctly retrieve media list from a given catalog.

## Detailed description
While developing some features for the Terraform provider for VCD, I've found out that the method `Catalog.QueryMediaList` was buggy. 
When executing it, it throws an error saying that the URL cannot be parsed.
After some debugging, I saw that the method fmt.Sprintf was being used wrongly:
```
filter := fmt.Sprintf("catalog==" + url.QueryEscape(catalog.Catalog.HREF))
```

This was causing that the string returned, instead of having escaped characters appropriately, it contained things like `(INVALID)` which were crashing the execution.

I fix the issue with the following change:
```
filter := fmt.Sprintf("catalog==%s", url.QueryEscape(catalog.Catalog.HREF))
``` 

Now things are working properly. Also I saw neither `Catalog.QueryMediaList` nor `Catalog.QueryVappTemplateList` had tests, so I wrote them! 🚀

### DIFF
--- a/.changes/v2.15.0/441-bug-fixes.md
+++ b/.changes/v2.15.0/441-bug-fixes.md
@@ -1,0 +1,1 @@
+* `Catalog.QueryMediaList` method was not working because fmt.Sprintf was being misused [GH-441]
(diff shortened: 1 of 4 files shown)


adambarreiro (approved):
Great!

lvirbalas (approved):
This looks great, thanks!

PR #436, opened 2022-02-18
Independent disk new properties support added which are supported from API version 36.0

vbauzysvmware:
For: https://github.com/vmware/terraform-provider-vcd/pull/789
Dep: #383

Disk methods have now the ability to access new properties from API version 36.0. 
* DiskRecordType.SharingType
* DiskRecordType.UUID
* DiskRecordType.Encrypted
* Disk.SharingType
* Disk.UUID
* Disk.Encrypted

New method added
* Disk.GetAttachedVmsHrefs

`Vms.VmReference` changed to array to fix not correct deserialization

Tested with 10.2.2 and 10.3.1

PR #444, opened 2022-03-07
Publish to external VCD ORG and subscribe attributes on OrgGeneralSettings struct

mikeletux:
## Description

In order to fulfill the issue from the Terraform provider for VCD https://github.com/vmware/terraform-provider-vcd/issues/775 we needed to add a couple of attributes to `OrgGeneralSettings`. These are `CanPublishExternally` and `CanSubscribe`.

## Detailed description
The current go-vcd SDK lacks of functionality for reading and changing some ORG values regarding catalogs. At the moment it is possible to read and change the option "Share catalogs to other organizations" through the attribute ´AdminOrg.OrgSettings.OrgGeneralSettings.CanPublishCatalogs´ but for options "Publish external catalogs" and "Subscribe to external catalogs" is not possible.  
  
This PR adds support for both options through the attributes `CanPublishExternally` and `CanSubscribe` from `OrgGeneralSettings` struct.

image

### DIFF
--- a/.changes/v2.15.0/444-improvements.md
+++ b/.changes/v2.15.0/444-improvements.md
@@ -1,0 +1,1 @@
+* Add `CanPublishExternally` and `CanSubscribe` attributes to `OrgGeneralSettings` struct. [GH-444]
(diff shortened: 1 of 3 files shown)


lvirbalas (approved):
Assuming the ordering of XML fields in types.go is correct, LGTM! :)

For reference (just as a note):
https://github.com/vmware/go-vcloud-director/pull/436#discussion_r811589299

Didainius (approved):

PR #447, opened 2022-03-11
Openapi TestConnection new method

mikeletux:
This PR does not have any issue associated.

# Description
When creating a catalog and subscribing to an external one, VCD uses the endpoint `/cloudapi/1.0.0/testConnection/` to check before hand if the endpoint where the user is trying to subscribe the external catalog from exists.

# Detailed description
In order to implement the external catalog subscription in the Terraform provider for VCD when creating a catalog, TestConnection functionality needs to be implemented since before subscribing, VCD does this previous sanity check.  
  
This has been implemented as a `Client` method, since it doesn't belong to any VCD resource and it's a client helping tool.  
  
Also, I propose a slight change to `Client.OpenApiPostItem` since to use TestConnection, I had to do a POST request against `/cloudapi/1.0.0/testConnection/`. This results in a `200OK`, instead of 201 or 202. 
`Client.OpenApiPostItem` wasn't covering this case, so the function worked but it wasn't unmarshalling the response into the needed struct.
This last point is open to discussion, as it might be a better way to handle this.

🚀

### DIFF
--- a/govcd/api.go
+++ b/govcd/api.go
@@ -16,6 +16,7 @@ import (
 	"net/url"
 	"os"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
@@ -793,6 +794,57 @@ func (client *Client) TestConnection(testConnection types.TestConnection) (*type
 	}
 
 	return returnTestConnectionResult, nil
+}
+
+// TestConnectionWithDefaults calls TestConnection given a subscriptionURL. The rest of parameters are set as default.
+// It returns whether it could reach the server and establish SSL connection or not.
+func (client *Client) TestConnectionWithDefaults(subscriptionURL string) (bool, error) {
+	if subscriptionURL == "" {
+		return false, fmt.Errorf("TestConnectionWithDefaults needs to be passed a host. i.e. my-host.vmware.com")
+	}
+
+	url, err := url.Parse(subscriptionURL)
+	if err != nil {
+		return false, fmt.Errorf("unable to parse URL - %s", err)
+	}
+
+	// Get port
+	var port int
+	if v := url.Port(); v != "" {
+		port, err = strconv.Atoi(v)
+		if err != nil {
+			return false, fmt.Errorf("couldn't parse port provided - %s", err)
+		}
+	} else {
+		switch url.Scheme {
+		case "http":
+			port = 80
+		case "https":
+			port = 443
+		}
+	}
+
+	testConnectionConfig := types.TestConnection{
+		Host:    url.Hostname(),
+		Port:    port,
+		Secure:  takeBoolPointer(true), // Default value used by VCD UI
+		Timeout: 30,                    // Default value used by VCD UI
+	}
+
+	testConnectionResult, err := client.TestConnection(testConnectionConfig)
+	if err != nil {
+		return false, err
+	}
+
+	if !testConnectionResult.TargetProbe.CanConnect {
+		return false, fmt.Errorf("the remote host is not reachable")
+	}
+
+	if !testConnectionResult.TargetProbe.SSLHandshake {
+		return true, fmt.Errorf("unsupported or unrecognized SSL message")
+	}
+
+	return true, nil
 }
 
 // ---------------------------------------------------------------------
--- a/govcd/api.go
+++ b/govcd/api.go
@@ -772,17 +772,9 @@ func (client *Client) RemoveProvidedCustomHeaders(values map[string]string) {
 func (client *Client) TestConnection(testConnection types.TestConnection) (*types.TestConnectionResult, error) {
 	endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointTestConnection
 
-	minimumApiVersion, err := client.checkOpenApiEndpointCompatibility(endpoint)
+	apiVersion, err := client.getOpenApiHighestElevatedVersion(endpoint)
 	if err != nil {
 		return nil, err
-	}
-
-	// TestConnection in versions lower than 36.0 doesn't support sending these fields in the POST payload and
-	// fails if they are included. These code snippet avoid sending them in case they are set in those versions.
-	if client.APIVCDMaxVersionIs("< 36.0") {
-		testConnection.HostnameVerificationAlgorithm = ""
-		testConnection.AdditionalCAIssuers = nil
-		testConnection.PreConfiguredProxy = ""
 	}
 
 	urlRef, err := client.OpenApiBuildEndpoint(endpoint)
@@ -795,7 +787,7 @@ func (client *Client) TestConnection(testConnection types.TestConnection) (*type
 		ProxyProbe:  &types.ProbeResult{},
 	}
 
-	err = client.OpenApiPostItem(minimumApiVersion, urlRef, nil, testConnection, returnTestConnectionResult, nil)
+	err = client.OpenApiPostItem(apiVersion, urlRef, nil, testConnection, returnTestConnectionResult, nil)
 	if err != nil {
 		return nil, fmt.Errorf("error performing test connection: %s", err)
 	}
--- a/govcd/api.go
+++ b/govcd/api.go
@@ -768,7 +768,7 @@ func (client *Client) RemoveProvidedCustomHeaders(values map[string]string) {
 	}
 }
 
-// TestConnection Tests a connection against a VCD, including SSL handshake and hostname verification.
+// TestConnection calls API to test a connection against a VCD, including SSL handshake and hostname verification.
 func (client *Client) TestConnection(testConnection types.TestConnection) (*types.TestConnectionResult, error) {
 	endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointTestConnection
 
--- a/govcd/api.go
+++ b/govcd/api.go
@@ -768,6 +768,41 @@ func (client *Client) RemoveProvidedCustomHeaders(values map[string]string) {
 	}
 }
 
+// TestConnection Tests a connection against a VCD, including SSL handshake and hostname verification.
+func (client *Client) TestConnection(testConnection types.TestConnection) (*types.TestConnectionResult, error) {
+	endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointTestConnection
+
+	minimumApiVersion, err := client.checkOpenApiEndpointCompatibility(endpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	// TestConnection in versions lower than 36.0 doesn't support sending these fields in the POST payload and
+	// fails if they are included. These code snippet avoid sending them in case they are set in those versions.
+	if client.APIVCDMaxVersionIs("< 36.0") {
+		testConnection.HostnameVerificationAlgorithm = ""
+		testConnection.AdditionalCAIssuers = nil
+		testConnection.PreConfiguredProxy = ""
+	}
+
+	urlRef, err := client.OpenApiBuildEndpoint(endpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	returnTestConnectionResult := &types.TestConnectionResult{
+		TargetProbe: &types.ProbeResult{},
+		ProxyProbe:  &types.ProbeResult{},
+	}
+
+	err = client.OpenApiPostItem(minimumApiVersion, urlRef, nil, testConnection, returnTestConnectionResult, nil)
+	if err != nil {
+		return nil, fmt.Errorf("error performing test connection: %s", err)
+	}
+
+	return returnTestConnectionResult, nil
+}
+
 // ---------------------------------------------------------------------
 // The following functions are needed to avoid strict Coverity warnings
 // ---------------------------------------------------------------------
--- a/govcd/openapi.go
+++ b/govcd/openapi.go
@@ -267,7 +267,7 @@ func (client *Client) OpenApiPostItem(apiVersion string, urlRef *url.URL, params
 		return err
 	}
 
-	// Handle two cases of API behaviour - synchronous (response status code is 201) and asynchronous (response status
+	// Handle two cases of API behaviour - synchronous (response status code is 200 or 201) and asynchronous (response status
 	// code 202)
 	switch resp.StatusCode {
 	// Asynchronous case - must track task and get item HREF from there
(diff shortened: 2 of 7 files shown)


lvirbalas:
> In order to implement the external catalog subscription in the Terraform provider for VCD when creating a catalog, TestConnection functionality needs to be implemented since before subscribing, VCD does this previous sanity check.

I guess I'm still missing the use case: who/where/when will call the new OpenApiTestConnection function?

Thanks!

mikeletux:
> > In order to implement the external catalog subscription in the Terraform provider for VCD when creating a catalog, TestConnection functionality needs to be implemented since before subscribing, VCD does this previous sanity check.
> 
> I guess I'm still missing the use case: who/where/when will call the new OpenApiTestConnection function?
> 
> Thanks!

Yeah, maybe I didn't get much context.  
When creating a catalog using the VCD UI (and as long as you have permission to subscribe to an external catalog), you can create a catalog subscribing from an external one. Please check the first figure:

image

If you check the HTTP traffic using a traffic sniffer, when you fill everything and click the OK button, you can see two POST calls being generated by the frontend.

image

- The first one is to `TestConnection`. That endpoint checks that the remote endpoint where you are trying to subscribe the external catalog from, exists.
- The second one is the actual call that creates the catalog for that Org.

Let me know if this make it clear 😄

lvirbalas (approved):
LGTM, and thanks for explaining how we will use it at https://github.com/vmware/go-vcloud-director/pull/447#issuecomment-1064971732

govcd/openapi.go
```diff
@@ -291,7 +291,7 @@ func (client *Client) OpenApiPostItem(apiVersion string, urlRef *url.URL, params
 		}
 
 		// Synchronous task - new item body is returned in response of HTTP POST request
-	case http.StatusCreated:
+	case http.StatusCreated, http.StatusOK:
```
Didainius, line 294: As I think about it the status 200 is probably because of better compatibility with load balancer probes, which like `http.StatusOK`. 

Comment in line 269 needs an update (https://github.com/vmware/go-vcloud-director/pull/447/files#diff-585772f0a826713495acbb0951db5ff9d9f252426fca85e2e112581e5d90a29dR269)

govcd/openapi.go
```diff
@@ -291,7 +291,7 @@ func (client *Client) OpenApiPostItem(apiVersion string, urlRef *url.URL, params
...
 		util.Logger.Printf("[TRACE] Synchronous task detected, marshalling outType '%s'", reflect.TypeOf(outType))
```
Didainius, line 295:
```suggestion
		util.Logger.Printf("[TRACE] Synchronous task detected (HTTP Status '%d'), marshalling outType '%s'", resp.StatusCode, reflect.TypeOf(outType))
```

govcd/openapi.go
```diff
@@ -689,6 +689,33 @@ func (client *Client) newOpenApiRequest(apiVersion string, params url.Values, me
...
+	}
+
+	returnTestConnectionResult := &types.TestConnectionResult{
+		TargetProbe: &types.ProbeResult{},
+		ProxyProbe:  &types.ProbeResult{},
+	}
+
+	err = client.OpenApiPostItem(minimumApiVersion, urlRef, nil, testConnection, returnTestConnectionResult, nil)
+	if err != nil {
+		return nil, fmt.Errorf("error creating Edge Gateway: %s", err)
```
Didainius, line 713: It is not an edge gateway, is it?

govcd/openapi.go
```diff
@@ -689,6 +689,33 @@ func (client *Client) newOpenApiRequest(apiVersion string, params url.Values, me
 	return req
 }
 
+// OpenApiTestConnection Tests a connection, including SSL handshake and hostname verification.
+func (client *Client) OpenApiTestConnection(testConnection types.TestConnection) (*types.TestConnectionResult, error) {
```
Didainius, line 693: In fact I think this function could live in `api.go` with other `client` functions and it also could work without the `OpenApi` prefix (just `TestConnection`). The reason is that it is not really a function verifying OpenApi, but checking VCD access/readyness in general.

The `openapi.go` contains mainly the tools to interact with OpenApi itself.

Didainius (changes requested):
Two parts

### OpenAPI function adjustment
Looks good in general, but because functions in `openapi.go` are a  backbone for all OpenAPI requests I have a few minor addition requests which I left inline, but also here as I can't comment untouched code lines.

* Could you also adjust sister function `OpenApiPostItemSync` so that these behave the same way in `sync` operations. Also it's documentation. For a test you could manually submit the payload for testConnnection to see if it actually handles it well.
* A changelog entry needs a separate line about adjustment of these functions and the reason for it. 
* I already mentioned in explicit content but because this is no longer a single status code that guides the operation of these functions it would be beneficial to explicitly log status code as we pick the path.
** https://github.com/vmware/go-vcloud-director/pull/447/files#diff-585772f0a826713495acbb0951db5ff9d9f252426fca85e2e112581e5d90a29dR295
** https://github.com/vmware/go-vcloud-director/pull/447/files#diff-585772f0a826713495acbb0951db5ff9d9f252426fca85e2e112581e5d90a29dR191-R193 (here once you adjust it to handle both codes)

### OpenApiTestConnection

The test `Test_OpenApiTestConnection` has issue on 10.2 versions:
* 10.2.0 - FAILS:
```
go test  -count 1 -tags ALL -check.vv -check.f OpenApiTestConnection -timeout=155m .
START: api_vcd_test.go:477: TestVCD.SetUpSuite
Running on VCD https://HOST/api (version 10.2.0.17008054 built at 2020-10-08 20:53:08 +0000 UTC)
as user administrator@System (using password)
Skipping all vapp tests because one of the following wasn't given: Network, StorageProfile, Catalog, Catalogitem
PASS: api_vcd_test.go:477: TestVCD.SetUpSuite   19.153s

START: openapi_test.go:265: TestVCD.Test_OpenApiTestConnection
openapi_test.go:304:
    check.Assert(err, IsNil)
... value *errors.errorString = &errors.errorString{s:"error creating Edge Gateway: error in HTTP POST request: VCD_50006 - [ 71b4c62b-90f2-4b16-9668-448b7d3358e5 ] Json processing error.\n - Unrecognized field \"hostnameVerificationAlgorithm\" (class com.vmware.vcloud.rest.openapi.model.Connection), not marked as ignorable (5 known properties: \"host\", \"port\", \"proxyConnection\", \"secure\", \"timeout\"])\n at [Source: (org.apache.cxf.transport.http.AbstractHTTPDestination$1); line: 6, column: 37] (through reference chain: com.vmware.vcloud.rest.openapi.model.Connection[\"hostnameVerificationAlgorithm\"])"} ("error creating Edge Gateway: error in HTTP POST request: VCD_50006 - [ 71b4c62b-90f2-4b16-9668-448b7d3358e5 ] Json processing error.\n - Unrecognized field \"hostnameVerificationAlgorithm\" (class com.vmware.vcloud.rest.openapi.model.Connection), not marked as ignorable (5 known properties: \"host\", \"port\", \"proxyConnection\", \"secure\", \"timeout\"])\n at [Source: (org.apache.cxf.transport.http.AbstractHTTPDestination$1); line: 6, column: 37] (through reference chain: com.vmware.vcloud.rest.openapi.model.Connection[\"hostnameVerificationAlgorithm\"])")

FAIL: openapi_test.go:265: TestVCD.Test_OpenApiTestConnection

START: api_vcd_test.go:1549: TestVCD.TearDownSuite
PASS: api_vcd_test.go:1549: TestVCD.TearDownSuite       0.002s

OOPS: 0 passed, 1 FAILED
```

* 10.2.2 - FAILS
```
go test  -count 1 -tags ALL -check.vv -check.f OpenApiTestConnection -timeout=155m .
START: api_vcd_test.go:477: TestVCD.SetUpSuite
Running on VCD https://HOST/api (version 10.2.2.18634228 built at 2021-09-15 17:30:35 +0000 UTC)
as user administrator@System (using password)
Skipping all vapp tests because one of the following wasn't given: Network, StorageProfile, Catalog, Catalogitem
PASS: api_vcd_test.go:477: TestVCD.SetUpSuite   21.967s

START: openapi_test.go:265: TestVCD.Test_OpenApiTestConnection
openapi_test.go:304:
    check.Assert(err, IsNil)
... value *errors.errorString = &errors.errorString{s:"error creating Edge Gateway: error in HTTP POST request: VCD_50006 - [ e08d6df1-4eb0-42ec-b090-cda5ef4dedeb ] Json processing error.\n - Unrecognized field \"hostnameVerificationAlgorithm\" (class com.vmware.vcloud.rest.openapi.model.Connection), not marked as ignorable (5 known properties: \"host\", \"port\", \"proxyConnection\", \"secure\", \"timeout\"])\n at [Source: (org.apache.cxf.metrics.interceptors.CountingInputStream); line: 6, column: 37] (through reference chain: com.vmware.vcloud.rest.openapi.model.Connection[\"hostnameVerificationAlgorithm\"])"} ("error creating Edge Gateway: error in HTTP POST request: VCD_50006 - [ e08d6df1-4eb0-42ec-b090-cda5ef4dedeb ] Json processing error.\n - Unrecognized field \"hostnameVerificationAlgorithm\" (class com.vmware.vcloud.rest.openapi.model.Connection), not marked as ignorable (5 known properties: \"host\", \"port\", \"proxyConnection\", \"secure\", \"timeout\"])\n at [Source: (org.apache.cxf.metrics.interceptors.CountingInputStream); line: 6, column: 37] (through reference chain: com.vmware.vcloud.rest.openapi.model.Connection[\"hostnameVerificationAlgorithm\"])")

FAIL: openapi_test.go:265: TestVCD.Test_OpenApiTestConnection

START: api_vcd_test.go:1549: TestVCD.TearDownSuite
PASS: api_vcd_test.go:1549: TestVCD.TearDownSuite       0.001s

OOPS: 0 passed, 1 FAILED
--- FAIL: Test (22.90s)
```


* 10.3.2 - OK

govcd/openapi.go
mikeletux: Comment updated now showing `200OK `code as well

govcd/openapi.go
mikeletux: Done!

govcd/openapi.go
mikeletux: Copy/Paste issue. Solved and thanks for spotting that!

govcd/openapi.go
mikeletux: Already moved `OpenApiTestConnection` from openapi.go to api.go. Also renamed to `TestConnection`

mikeletux:
@Didainius some comments regarding the changes that I did.
- Apart from the changes you suggested to `Client.OpenApiPostItemSync` now the method is returning an error if the status code that is returned is not 200 or 201. This wasn't the previous behavior. I did it because the function description actually says that if return code is different from 200 or 201 it errors, and also because when writing the test if wasn't impossible to find out if POST against an endpoint that returns 200OK was handled appropriately. 
- Add a workaround to `Client.TestConnection` so if API max version is below 36.0, sets some fields not supported to empty so they are not send in the POST request and the function doesn't error (I tested this on 10.2.2. Could you please test it on 10.2.0?)
- Add a test in `Test_OpenApiInlineStructCRUDRoles` that manually builds TestConnection payload and send it using `Client.OpenApiPostItemSync` to check that this function handles correctly 200OK.

Let me know if you're concerned about anything else! 🚀

govcd/api.go
```diff
@@ -768,6 +768,41 @@ func (client *Client) RemoveProvidedCustomHeaders(values map[string]string) {
 	}
 }
 
+// TestConnection Tests a connection against a VCD, including SSL handshake and hostname verification.
```
vbauzysvmware, line 771:
```suggestion
// TestConnection calls API to test a connection against a VCD, including SSL handshake and hostname verification.
```

vbauzysvmware (commented):
I would say we need wrapper function too, which would have only minimum values for input like URL and everything else default. I guess in UI you provide URL.

govcd/api.go
mikeletux: Already changed!

mikeletux:
> I would say we need wrapper function too, which would have only minimum values for input like URL and everything else default. I guess in UI you provide URL.

This is interesting. Let's see what @lvirbalas @Didainius @adambarreiro have to say about it. 
I could grab the default values that the UI uses and write a wrapper.

Didainius:
> This is interesting. Let's see what @lvirbalas @Didainius @adambarreiro have to say about it.
> I could grab the default values that the UI uses and write a wrapper.

Go ahead if this is easy to make it (no advanced computations for fields)

vbauzysvmware:
> > I would say we need wrapper function too, which would have only minimum values for input like URL and everything else default. I guess in UI you provide URL.
> 
> This is interesting. Let's see what @lvirbalas @Didainius @adambarreiro have to say about it. I could grab the default values that the UI uses and write a wrapper.

how the current function is used in terraform? how does call look like?

mikeletux:
> how the current function is used in terraform? how does call look like?

Is not yet being used, but I'll use it when enhancing the Catalog resource, when it comes to create a new catalog subscribing from an external one. 
Before doing anything, check that the VCD instance where the catalog you're going to subscribe from, exist.

vbauzysvmware:
> > how the current function is used in terraform? how does call look like?
> 
> Is not yet being used, but I'll use it when enhancing the Catalog resource, when it comes to create a new catalog subscribing from an external one. Before doing anything, check that the VCD instance where the catalog you're going to subscribe from, exist.

So then you see how kind wrapper is better :)

mikeletux:
@vbauzysvmware added the method `TestConnectionWithDefaults` so it gets a subscription URL and returns if it was possible to connect.

govcd/openapi_endpoints.go
```diff
@@ -115,7 +121,7 @@ func (client *Client) checkOpenApiEndpointCompatibility(endpoint string) (string
 // supported API versions just like client.checkOpenApiEndpointCompatibility().
 //
 // The advantage of this functions is that it provides a controlled API elevation instead of just picking the highest
-// which could be risky and untested (especially if new API version is released after release of package consuming this
+// which could be risky and untested (especially if new API version hich could be risky and untested (especially if new API version is released after release of package consuming this
```
Didainius, line 124: This remaining thing to fix.

Didainius (approved):
Thanks for fixes. I am approving - just fix the typo in a separate comment.

vbauzysvmware (approved):
Thank you

govcd/openapi_endpoints.go
mikeletux: Rephrased that so it makes sense now. Thanks for spotting that 🚀

adambarreiro (approved):

Didainius:
Closing in favor of #501

PR #449, opened 2022-03-18
Fix group.Update() when users list is populated

adambarreiro:
## Description

Let's consider that we're working with LDAP users and groups. When a group is read from LDAP, and it contains users, the attribute UsersList is populated with the names of these users. The problem is, if the group is updated, and the users list is sent back, VCD will throw this error:

```
HTTP 400 Bad Request: Invalid content was found starting with element 'UsersList'. No child element is expected at this point
```

The proposed fix is to send this attribute always to `nil` in the `group.Update()` function.

I've also checked what happens with `user.Update()`, as common sense would suggest that the behaviour is the same, but it is not. In fact, there's already a check kind of similar to the one proposed, in **user.go:414**.

In this case, `user.Update()` worked out of the box, so I haven't touched that part. What I did, in the other hand, is enrich the group test with an `user.Update()` call, to have this behaviour covered as well.

### DIFF
--- a/.changes/v2.15.0/439-improvements.md
+++ b/.changes/v2.15.0/439-improvements.md
@@ -1,2 +1,3 @@
 * Add support for `User` entities imported from LDAP, with `IsExternal` property [GH-439]
 * Add support for users list attribute for `Group` [GH-439]
+* Improve `group.Update()` to avoid sending the users list to VCD to avoid unwanted errors [GH-449]


govcd/group.go
```diff
@@ -183,3 +183,21 @@ func validateDeleteGroup(group *types.Group) error {
 
 	return nil
 }
+
+// copyWithoutUserList returns a copy of the given group, with the UserList attribute set to nil.
+// This can and should be used to interact with VCD after a group read from the LDAP,
+// as having this list populated will return an error 400 as VCD doesn't expect this list to be updatable.
+func copyWithoutUserList(group *types.Group) *types.Group {
```
Didainius, line 190: Could you put one test step for this  function.  (could live in test_GroupUserListIsPopulated).


Something like this:
```go
originalGroup.Userlist = nil
check.Assert(copiedGroupWithoutUserList, DeepEquals, originalGroup)
```

This should allow us to remind improving this function if we ever add additional fields to `types.Group`.

Didainius:
Also add a note to changelog.

lvirbalas (approved):
p.s. Found during one of the test-fests.

govcd/group.go
adambarreiro: Done, ran the test with the new addition and it passes, please check now.

adambarreiro:
> Also add a note to changelog.

Done, thanks!

Didainius (approved):
Thanks. LDAP related tests pass.

vbauzysvmware (approved):

PR #450, opened 2022-03-21
Add method AdminOrg.FindCatalogRecords 

mikeletux:
There is no issue associated with this PR.
The purpose of this PR is to provide the right method to retrieve information about catalog records. This method will be used for the Terraform provider to set more information about catalogs.

## Description
govcd lacks of methods to query catalogRecords by name. This PR provides the method `AdminOrg.FindCatalogRecords` that does that.

## Detailed description
Previously there was just one method called `AdminOrg.QueryCatalogList` that was returning all catalog records for a given AdminOrg. There was no way to query catalog/s using their name.
With `AdminOrg.FindCatalogRecords`, given a name, the SDK can retrieve just the specific catalog record.
This is usefull for Terraform provider for VCD. Please check PR https://github.com/vmware/terraform-provider-vcd/pull/800

### DIFF
--- a/govcd/adminorg.go
+++ b/govcd/adminorg.go
@@ -763,9 +763,14 @@ func (adminOrg *AdminOrg) QueryCatalogList() ([]*types.CatalogRecord, error) {
 // FindCatalogRecords given a catalog name, retrieves the catalogRecords for a given organization
 func (adminOrg *AdminOrg) FindCatalogRecords(name string) ([]*types.CatalogRecord, error) {
 	util.Logger.Printf("[DEBUG] QueryCatalogList with org name %s", adminOrg.AdminOrg.Name)
-	queryType := types.QtCatalog
+
 	if adminOrg.client.IsSysAdmin {
-		queryType = types.QtAdminCatalog
+		// Set tenant context
+		adminOrg.client.SetCustomHeader(map[string]string{
+			types.HeaderAuthContext:   adminOrg.TenantContext.OrgName,
+			types.HeaderTenantContext: adminOrg.TenantContext.OrgId,
+		})
+		defer adminOrg.client.RemoveCustomHeader()
 	}
 
 	var filter string
@@ -774,8 +779,8 @@ func (adminOrg *AdminOrg) FindCatalogRecords(name string) ([]*types.CatalogRecor
 		filter = fmt.Sprintf("%s;name==%s", filter, name)
 	}
 
-	results, err := adminOrg.client.cumulativeQuery(queryType, nil, map[string]string{
-		"type":          queryType,
+	results, err := adminOrg.client.cumulativeQuery(types.QtCatalog, nil, map[string]string{
+		"type":          types.QtCatalog,
 		"filter":        filter,
 		"filterEncoded": "true",
 	})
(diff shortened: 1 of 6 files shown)


vbauzysvmware (approved):
LGTM

govcd/adminorg_test.go
```diff
@@ -190,6 +190,11 @@ func (vcd *TestVCD) TestOrg_AdminOrg_QueryCatalogList(check *C) {
 	catalogsInAdminOrg, err := adminOrg.QueryCatalogList()
 	check.Assert(err, IsNil)
 
+	// gets a specific catalog as an adminOrg
+	singleCatalogInAdminOrg, err := adminOrg.FindCatalogRecords(vcd.config.VCD.Catalog.Name)
+	check.Assert(err, IsNil)
```
vbauzysvmware, line 195: check that singleCatalogInAdminOrg is not nil :)

govcd/adminorg.go
```diff
@@ -757,27 +757,42 @@ func (adminOrg *AdminOrg) GetVdcByName(vdcname string) (Vdc, error) {
...
+
+// FindCatalogRecords given a catalog name, retrieves the catalogRecords for a given organization
+func (adminOrg *AdminOrg) FindCatalogRecords(name string) ([]*types.CatalogRecord, error) {
 	util.Logger.Printf("[DEBUG] QueryCatalogList with org name %s", adminOrg.AdminOrg.Name)
-	queryType := types.QtCatalog
+
 	if adminOrg.client.IsSysAdmin {
-		queryType = types.QtAdminCatalog
+		// Set tenant context
+		adminOrg.client.SetCustomHeader(map[string]string{
```
Didainius, line 769: There is a risk of setting header value globally (even for a short term).

The `client` is always used as a pointer - that means it is shared throughout everywhere. Now in case a user is doing any other request before the `defer adminOrg.client.RemoveCustomHeader()` gets into effect - his query would be affected by these headers.

In Terraform case - there could be some other resource doing query - and if it happens to happen at exactly the same time when header is set globally, but not yet removed - its query will be impacted. 

It would be a lot safer to look for an option to directly embed these headers into that particular request.

mikeletux: yes, this is very right and thanks for pointing this out. I'll figure out the way to set headers just for this query and not for the whole client.

lvirbalas:
> go-vcd lacks of methods to query catalogRecords by name

Note: correct abbreviation is **govcd** (no hyphen).

lvirbalas (approved):
LGTM, but please merge only when https://github.com/vmware/go-vcloud-director/pull/450/files#r831825823 is figured out :)

govcd/adminorg.go
mikeletux: @Didainius I've totally changed the approach here.
I've added several methods `*WithHeaders` that allow to do queries sending headers. I also kept the old one for keeping backwards compatibility and avoid breaking the API.

Having done that, now client has a method called `cumulativeQueryWithHeaders` that allows us to do cumulative queries adding headers, the ones that can be used to set tenant context if needed.

Now, these headers are only sent in this query so we avoid the issues you were pointing out in you last comment.

mikeletux:
⚠️ Still need to write changelog for new methods

govcd/adminorg.go
Didainius: Thanks - I like this very much - just have a run for a few tests to be sure we have no impact.

Didainius (approved):

mikeletux:
Tests passed, merging shortly!

PR #452, opened 2022-03-29
Add support for Distributed Firewall and Network Context Profile lookup

Didainius:
This PR adds support for Distributed Firewall Rules and Network Context Profile lookup functions (VCD does not yet support creation of Netwok Context Profiles in UI).

* Add support for for Network Context Profile lookup using `GetAllNetworkContextProfiles` and
  `GetNetworkContextProfilesByNameScopeAndContext` functions
* Add support for NSX-T Distributed Firewall rule management using type `DistributedFirewall` and
`VdcGroup.GetDistributedFirewall`, `VdcGroup.UpdateDistributedFirewall`,
`VdcGroup.DeleteAllDistributedFirewallRules`, `DistributedFirewall.DeleteAllRules`

Note. Tests with tags `"network nsxt vdcGroup edgegateway"` passed on 10.2.0, 10.2.2.2, 10.3.2

### DIFF
--- a/govcd/nsxt_distributed_firewall.go
+++ b/govcd/nsxt_distributed_firewall.go
@@ -16,6 +16,9 @@ type DistributedFirewall struct {
 }
 
 // GetDistributedFirewall retrieves Distributed Firewall in a VDC Group which contains all rules
+//
+// Note. This function works only with `default` policy as this was the only supported when this
+// functions was created
 func (vdcGroup *VdcGroup) GetDistributedFirewall() (*DistributedFirewall, error) {
 	client := vdcGroup.client
 	endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointVdcGroupsDfwRules
@@ -24,6 +27,7 @@ func (vdcGroup *VdcGroup) GetDistributedFirewall() (*DistributedFirewall, error)
 		return nil, err
 	}
 
+	// "default" policy is hardcoded because there is no other policy supported
 	urlRef, err := client.OpenApiBuildEndpoint(fmt.Sprintf(endpoint, vdcGroup.VdcGroup.Id, "default"))
 	if err != nil {
 		return nil, err
@@ -44,6 +48,9 @@ func (vdcGroup *VdcGroup) GetDistributedFirewall() (*DistributedFirewall, error)
 }
 
 // UpdateDistributedFirewall updates Distributed Firewall in a VDC Group
+//
+// Note. This function works only with `default` policy as this was the only supported when this
+// functions was created
 func (vdcGroup *VdcGroup) UpdateDistributedFirewall(dfwRules *types.DistributedFirewallRules) (*DistributedFirewall, error) {
 	client := vdcGroup.client
 	endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointVdcGroupsDfwRules
@@ -73,12 +80,18 @@ func (vdcGroup *VdcGroup) UpdateDistributedFirewall(dfwRules *types.DistributedF
 }
 
 // DeleteAllDistributedFirewallRules removes all Distributed Firewall rules
+//
+// Note. This function works only with `default` policy as this was the only supported when this
+// functions was created
 func (vdcGroup *VdcGroup) DeleteAllDistributedFirewallRules() error {
 	_, err := vdcGroup.UpdateDistributedFirewall(&types.DistributedFirewallRules{})
 	return err
 }
 
 // DeleteAllRules removes all Distributed Firewall rules
+//
+// Note. This function works only with `default` policy as this was the only supported when this
+// functions was created
 func (firewall *DistributedFirewall) DeleteAllRules() error {
 	if firewall.VdcGroup != nil && firewall.VdcGroup.VdcGroup != nil && firewall.VdcGroup.VdcGroup.Id == "" {
 		return errors.New("empty VDC Group ID for parent VDC Group")
--- a/govcd/nsxt_distributed_firewall.go
+++ b/govcd/nsxt_distributed_firewall.go
@@ -7,13 +7,15 @@ import (
 	"github.com/vmware/go-vcloud-director/v2/types/v56"
 )
 
-// DistributedFirewall contains a types.DfwFirewallRule
+// DistributedFirewall contains a types.DistributedFirewallRules which handles Distributed Firewall
+// rules in a VDC Group
 type DistributedFirewall struct {
 	DistributedFirewallRuleContainer *types.DistributedFirewallRules
 	client                           *Client
 	VdcGroup                         *VdcGroup
 }
 
+// GetDistributedFirewall retrieves Distributed Firewall in a VDC Group which contains all rules
 func (vdcGroup *VdcGroup) GetDistributedFirewall() (*DistributedFirewall, error) {
 	client := vdcGroup.client
 	endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointVdcGroupsDfwRules
@@ -41,6 +43,7 @@ func (vdcGroup *VdcGroup) GetDistributedFirewall() (*DistributedFirewall, error)
 	return returnObject, nil
 }
 
+// UpdateDistributedFirewall updates Distributed Firewall in a VDC Group
 func (vdcGroup *VdcGroup) UpdateDistributedFirewall(dfwRules *types.DistributedFirewallRules) (*DistributedFirewall, error) {
 	client := vdcGroup.client
 	endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointVdcGroupsDfwRules
@@ -49,6 +52,7 @@ func (vdcGroup *VdcGroup) UpdateDistributedFirewall(dfwRules *types.DistributedF
 		return nil, err
 	}
 
+	// "default" policy is hardcoded because there is no other policy supported
 	urlRef, err := client.OpenApiBuildEndpoint(fmt.Sprintf(endpoint, vdcGroup.VdcGroup.Id, "default"))
 	if err != nil {
 		return nil, err
@@ -68,11 +72,13 @@ func (vdcGroup *VdcGroup) UpdateDistributedFirewall(dfwRules *types.DistributedF
 	return returnObject, nil
 }
 
+// DeleteAllDistributedFirewallRules removes all Distributed Firewall rules
 func (vdcGroup *VdcGroup) DeleteAllDistributedFirewallRules() error {
 	_, err := vdcGroup.UpdateDistributedFirewall(&types.DistributedFirewallRules{})
 	return err
 }
 
+// DeleteAllRules removes all Distributed Firewall rules
 func (firewall *DistributedFirewall) DeleteAllRules() error {
 	if firewall.VdcGroup != nil && firewall.VdcGroup.VdcGroup != nil && firewall.VdcGroup.VdcGroup.Id == "" {
 		return errors.New("empty VDC Group ID for parent VDC Group")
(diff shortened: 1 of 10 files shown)


govcd/nsxt_distributed_firewall.go
```diff
@@ -0,0 +1,88 @@
...
+// GetDistributedFirewall retrieves Distributed Firewall in a VDC Group which contains all rules
+func (vdcGroup *VdcGroup) GetDistributedFirewall() (*DistributedFirewall, error) {
+	client := vdcGroup.client
+	endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointVdcGroupsDfwRules
+	apiVersion, err := client.getOpenApiHighestElevatedVersion(endpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	urlRef, err := client.OpenApiBuildEndpoint(fmt.Sprintf(endpoint, vdcGroup.VdcGroup.Id, "default"))
```
adambarreiro, line 27: Do you think that we should mention somewhere that only `default` FW policy is supported at the moment, or is it too obvious?

govcd/nsxt_distributed_firewall.go
```diff
@@ -0,0 +1,88 @@
...
+// UpdateDistributedFirewall updates Distributed Firewall in a VDC Group
+func (vdcGroup *VdcGroup) UpdateDistributedFirewall(dfwRules *types.DistributedFirewallRules) (*DistributedFirewall, error) {
+	client := vdcGroup.client
+	endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointVdcGroupsDfwRules
+	apiVersion, err := client.getOpenApiHighestElevatedVersion(endpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	// "default" policy is hardcoded because there is no other policy supported
```
adambarreiro, line 55: Maybe we could put this comment also in line 27?

govcd/nsxt_distributed_firewall_test.go
```diff
@@ -0,0 +1,272 @@
...
+	err = vdcGroup.Delete()
+	check.Assert(err, IsNil)
+	err = vdc.DeleteWait(true, true)
+	check.Assert(err, IsNil)
+}
+
+// createDistributedFirewallDefinitions creates some randomized firewall rule configurations to match possible configurations
+func createDistributedFirewallDefinitions(check *C, vcd *TestVCD, vdcGroupId string) ([]*types.DistributedFirewallRule, *NsxtFirewallGroup, *NsxtFirewallGroup) {
+	// This number does not impact performance because all rules are created at once in the API
+	numberOfRules := 40
```
adambarreiro, line 156: Just curious, if `40` was chosen for any reason or simply a random number?

govcd/nsxt_distributed_firewall_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+	// Get existing firewall rule configuration
+	fwRules, err := vdcGroup.GetDistributedFirewall()
+	check.Assert(err, IsNil)
+	check.Assert(fwRules.DistributedFirewallRuleContainer.Values, NotNil)
+
+	// Create some prerequisites and generate firewall rule configurations to feed them into config
+	randomizedFwRuleDefs, ipSet, secGroup := createDistributedFirewallDefinitions(check, vcd, vdcGroup.VdcGroup.Id)
+
+	// Add IP Set and to cleanup list
```
adambarreiro, line 78:
```suggestion
	// Add IP Set and Security Group to cleanup list
```

govcd/nsxt_distributed_firewall_test.go
```diff
@@ -0,0 +1,272 @@
...
+	openApiEndpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointFirewallGroups + ipSet.NsxtFirewallGroup.ID
+	PrependToCleanupListOpenApi(ipSet.NsxtFirewallGroup.Name, check.TestName(), openApiEndpoint)
+	openApiEndpoint = types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointFirewallGroups + secGroup.NsxtFirewallGroup.ID
+	PrependToCleanupListOpenApi(secGroup.NsxtFirewallGroup.Name, check.TestName(), openApiEndpoint)
```
adambarreiro, line 82: I may be wrong, but isn't this done also at the end of `preCreateVdcGroupSecurityGroup` (which is called by the above `createDistributedFirewallDefinitions`)

adambarreiro (approved):

types/v56/nsxt_types.go
```diff
@@ -1106,3 +1106,101 @@ type NsxtAlbVirtualServiceApplicationProfile struct {
 	// * L4 TLS (certificate reference is mandatory)
 	Type string `json:"type"`
 }
+
+// DistributedFirewallRule represents a single Distributed firewall rule
```
lvirbalas, line 1110: Nit, keeping the style of your other comments below:

```suggestion
// DistributedFirewallRule represents a single Distributed Firewall rule
```

lvirbalas (approved):

govcd/nsxt_distributed_firewall_test.go
Didainius: Totally random. Big enough to generate enough different setups

govcd/nsxt_distributed_firewall_test.go
Didainius: Totally right. I the difference was that in these functions I had `AddToCleanupListOpenApi` (which adds to the env of cleanup), but in reality I needed `PrependToCleanupListOpenApi` which adds to the front of cleanup list.

The reason it needs to be cleaned up early is that it must be cleaned up before VDC Group as VDC Group removal fails if there stil are IP sets and/or security groups in it.

govcd/nsxt_distributed_firewall_test.go
Didainius: Removed this cleanup part at all

govcd/nsxt_distributed_firewall.go
Didainius: Added this and `note` for each function docs.

govcd/nsxt_distributed_firewall_test.go
```diff
@@ -0,0 +1,292 @@
...
+	vdc, vdcGroup := test_CreateVdcGroup(check, adminOrg, vcd)
+	check.Assert(vdc, NotNil)
+	check.Assert(vdcGroup, NotNil)
+
+	// Run firewall tests as System user
+	fmt.Println("# Running Distributed Firewall tests as 'System' user")
+	test_NsxtDistributedFirewallRules(vcd, check, vdcGroup.VdcGroup.Id, vcd.client, vdc)
+
+	// Prep Org admin user and run firewall tests
+	userName := strings.ToLower(check.TestName())
```
vbauzysvmware, line 42: we do this like this, because when running the whole suite as ORG user then this test will be ignored?

govcd/nsxt_network_context_profile.go
```diff
@@ -0,0 +1,77 @@
...
+// * Network provider ID - to get Network Context Profiles scoped for attached NSX-T environment
+// * VDC Group ID - to get Network Context Profiles scoped for attached NSX-T environment
+//
+// scope can be one off:
+// * SYSTEM
+// * PROVIDER
+// * TENANT
+func GetNetworkContextProfilesByNameScopeAndContext(client *Client, name, scope, contextId string) (*types.NsxtNetworkContextProfile, error) {
+	if name == "" || contextId == "" || scope == "" {
+		return nil, fmt.Errorf("error - 'name', 'scope' and 'contextId' must be specified")
```
vbauzysvmware, line 50: Does API error is good enough if the scope value isn't correct?

vbauzysvmware (changes requested):
Just Small things :)

mikeletux (approved):
Great test coverage, approved!

govcd/nsxt_distributed_firewall_test.go
Didainius: No. go-vcloud-director does not have an option to run tests as Org user at all so I wanted to test this out here.
The `strings.ToLower` is here because VCD will create a user in lowercase by default and user creation will fail.

govcd/nsxt_network_context_profile.go
Didainius: It is not very bad IMO. Here is an example for "scope=HORN":
```
.. value *errors.errorString = &errors.errorString{s:"error getting all pages for endpoint https://bos1-vcloud-static-171-96.eng.vmware.com/cloudapi/1.0.0/networkContextProfiles: error in HTTP GET request: BAD_REQUEST - Cannot parse the supplied filter: scope==HORN"} ("error getting all pages for endpoint https://bos1-vcloud-static-171-96.eng.vmware.com/cloudapi/1.0.0/networkContextProfiles: error in HTTP GET request: BAD_REQUEST - Cannot parse the supplied filter: scope==HORN")
```

govcd/nsxt_distributed_firewall.go
```diff
@@ -0,0 +1,101 @@
...
+
+	returnObject := &DistributedFirewall{
+		DistributedFirewallRuleContainer: &types.DistributedFirewallRules{},
+		client:                           client,
+		VdcGroup:                         vdcGroup,
+	}
+
+	err = client.OpenApiPutItem(apiVersion, urlRef, nil, dfwRules, returnObject.DistributedFirewallRuleContainer, nil)
+	if err != nil {
+		return nil, fmt.Errorf("error setting Distributed Firewall rules: %s", err)
```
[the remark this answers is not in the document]
Didainius: Fixed.

govcd/nsxt_distributed_firewall.go
```diff
@@ -0,0 +1,101 @@
...
+func (vdcGroup *VdcGroup) GetDistributedFirewall() (*DistributedFirewall, error) {
+	client := vdcGroup.client
+	endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointVdcGroupsDfwRules
+	apiVersion, err := client.getOpenApiHighestElevatedVersion(endpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	// "default" policy is hardcoded because there is no other policy supported
+	urlRef, err := client.OpenApiBuildEndpoint(fmt.Sprintf(endpoint, vdcGroup.VdcGroup.Id, "default"))
```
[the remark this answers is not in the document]
Didainius: Changed.

govcd/nsxt_distributed_firewall_test.go
vbauzysvmware: Thank you for answer.

govcd/nsxt_network_context_profile.go
vbauzysvmware: Let's hope it will be enough :)

govcd/nsxt_distributed_firewall.go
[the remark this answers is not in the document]
vbauzysvmware: but I don't see PR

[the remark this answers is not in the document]
Didainius: Indeed. I posted before I pushed as firewall test was still running. Now it is pushed and tested.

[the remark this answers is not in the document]
vbauzysvmware: That's ok, just precaution change wasn't lost :)

vbauzysvmware (approved):

govcd/nsxt_distributed_firewall_test.go
```diff
@@ -0,0 +1,292 @@
...
+		dumpDistributedFirewallRulesToScreen(randomizedFwRuleDefs)
+	}
+
+	fwUpdated, err := vdcGroup.UpdateDistributedFirewall(fwRules.DistributedFirewallRuleContainer)
+	check.Assert(err, IsNil)
+	check.Assert(fwUpdated, Not(IsNil))
+
+	check.Assert(len(fwUpdated.DistributedFirewallRuleContainer.Values), Equals, len(randomizedFwRuleDefs))
+
+	// Check that all created rules are have the same attributes and order
```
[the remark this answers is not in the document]
Didainius: Yeah, I made an error. Fixed.
"// Check that all created rules have the same attributes and order"

PR #454, opened 2022-03-30
Add generic CRUD metadata by href functions

adambarreiro:
Signed-off-by: abarreiro <abarreiro@vmware.com>

This PR relates to https://github.com/vmware/terraform-provider-vcd/issues/372, as the goal is to provide CRUD functions for metadata to resources that **don't have an associated type**, like Storage profiles.

The functions just retrieve an HREF instead, and perform the metadata operation on it.

Another way of doing this would be changing the signature to:

```
func (reference *types.Reference) AddMetadataEntryByHrefAsync(typedValue, key, value string) {
// Use reference.HREF here
```

But I followed the approach of having this method associated to `VCDClient` instead. (This is, of course, open to discussion).

### DIFF
--- a/govcd/metadata.go
+++ b/govcd/metadata.go
@@ -12,6 +12,45 @@ import (
 	"github.com/vmware/go-vcloud-director/v2/types/v56"
 )
 
+
+// GetMetadataByHref returns metadata from the given resource reference.
+func (vcdClient *VCDClient) GetMetadataByHref(href string) (*types.Metadata, error) {
+	return getMetadata(&vcdClient.Client, href)
+}
+
+// AddMetadataEntryByHref adds metadata typedValue and key/value pair provided as input to the given resource reference,
+// then waits for the task to finish.
+func (vcdClient *VCDClient) AddMetadataEntryByHref(href, typedValue, key, value string) error {
+	task, err := vcdClient.AddMetadataEntryByHrefAsync(href, typedValue, key, value)
+	if err != nil {
+		return err
+	}
+	return task.WaitTaskCompletion()
+}
+
+// AddMetadataEntryByHrefAsync adds metadata typedValue and key/value pair provided as input to the given resource reference
+// and returns the task.
+func (vcdClient *VCDClient) AddMetadataEntryByHrefAsync(href, typedValue, key, value string) (Task, error) {
+	return addMetadata(&vcdClient.Client, typedValue, key, value, href)
+}
+
+// DeleteMetadataEntryByHref deletes metadata from the given resource reference, depending on key provided as input
+// and waits for the task to finish.
+func (vcdClient *VCDClient) DeleteMetadataEntryByHref(href, key string) error {
+	task, err := vcdClient.DeleteMetadataEntryByHrefAsync(href, key)
+	if err != nil {
+		return err
+	}
+
+	return task.WaitTaskCompletion()
+}
+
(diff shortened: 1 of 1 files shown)


govcd/metadata_test.go
```diff
@@ -898,3 +899,47 @@ func (vcd *TestVCD) Test_MetadataOnVdcNetworkCRUD(check *C) {
...
+
+func (vcd *TestVCD) Test_MetadataByHrefCRUD(check *C) {
+	fmt.Printf("Running: %s\n", check.TestName())
+	storageProfileRef, err := vcd.vdc.FindStorageProfileReference(vcd.config.VCD.StorageProfile.SP1)
+	if err != nil {
+		check.Skip(fmt.Sprintf("Test_MetadataByHrefCRUD: Storage Profile %s not found. Test can't proceed", vcd.config.VCD.StorageProfile.SP1))
+		return
+	}
+
+	storageProfileAdminHref := strings.ReplaceAll(storageProfileRef.HREF, "api", "api/admin")
```
Didainius, line 911: Is it only 'admin' which can operate on metadata? What could happen if I "set" on the url with `"api/admin"` and read on non-admin?

Didainius (commented):
Missing changelog entry.

govcd/metadata_test.go
adambarreiro: In this case, for `Storage Policy`, yes, it's only the admin who can get, put, post and delete metadata. I will add also a test to get metadata with **no admin** user, good point (it's the only thing the tenant user can do for storage profile metadata)

.changes/v2.15.0/454-features.md
```diff
@@ -0,0 +1,3 @@
+* Added support to set, get and delete metadata to any supported resource via its HREF, with the methods
```
lvirbalas, line 1: What about adding a list or just an explanation about which resources support this generic metadata function and which don't?

adambarreiro: Done, please check if you're OK with the formatting or I can refine it.

lvirbalas: Looks good to me, @vbauzysvmware could you review the list and the way we provide resource names too? E.g. is "disk" specific enough, etc. Thanks.

adambarreiro: (Just for context, picked namings from [here](https://developer.vmware.com/apis/1046/vmware-cloud-director/doc/doc/landing-user_operations.html))

.changes/v2.15.0/454-features.md
```diff
@@ -0,0 +1,5 @@
+* Added support to set, get and delete metadata to the following resources via its HREF:
+  `catalog`, `catalog item`, `disk`, `edge gateway`, `media`, `network`, `org`, `PVDC`, `PVDC storage profile`, `vApp`, `vApp template`,`VDC` and `VDC storage profile`;
```
vbauzysvmware, line 2:
```suggestion
  `catalog`, `catalog item`, `independent disk`, `edge gateway`, `media`, `network`, `org`, `PVDC`, `PVDC storage profile`, `vApp`, `vApp template`,`VDC` and `VDC storage profile`;
```

vbauzysvmware (approved):

.changes/v2.15.0/454-features.md
adambarreiro: Done!

lvirbalas (approved):

PR #455, opened 2022-04-01
Waiting: Add OpenAPI metadata methods for OpenAPI VDC Networks

adambarreiro:
This PR aims for two things:
1. Add support for OpenAPI VDC Network metadata support
2. For the above, it adds OpenAPI functions to manage metadata (private and generic CRUD functions)

However, metadata support in OpenAPI is still in `alpha` status, so this PR must remain in Draft.


govcd/metadata.go
```diff
@@ -91,58 +91,16 @@ func (vapp *VApp) GetMetadata() (*types.Metadata, error) {
 	return getMetadata(vapp.client, vapp.VApp.HREF)
 }
 
-func getMetadata(client *Client, requestUri string) (*types.Metadata, error) {
```
adambarreiro, line 94: Just moved these private functions to the bottom for the sake of ordering a bit.

adambarreiro:
Going back to draft state

PR #458, opened 2022-04-04
Catalog and LDAP tests fix

mikeletux:
This PR is not associated with any issue.

## Description

Fix issues with Catalog tests, as well as LDAP test framework.

## Detailed description
This PR addresses two issues that were impacting tests. 

### Issues solved related Catalog
There were some Catalog tests that were not removing the CatalogItems being upload to Catalogs. These tests were leveraging the function `AddToCleanupList` to delete those CatalogItems. This was leading to issues because the clean-up function is executed at the end of the test framework, and other tests might not be aware about the changes being done there.
The fix introduced here is to make every tests responsible of cleaning up whatever it has uploaded to Catalogs after the testing part has been done and right before the test returns. Hence, the `AddToCleanupList` function lays as last resort for cleaning-up the environment at the end of the tests if something fails.

### Issues solved related LDAP
Turns out that when configuring LDAP settings, for whatever reason, two Org catalog publishing vars gets disable (these are `Publish external catalogs` and `Subscribe to external catalogs`). This is a know issue and VCD team is addressing it to be solved. This interference with more tests as it makes the environment non-consistent.
In the meantime, this workaround what it does is, before doing any configuration, it grabs those two values, and when all LDAP testing sub-framework is closing up, is sets those vars again to the Org and finishes.

Tested and working in VCD 10.2.2 and 10.3.2

### DIFF
--- a/.changes/v2.15.0/458-improvements.md
+++ b/.changes/v2.15.0/458-improvements.md
@@ -1,0 +1,2 @@
+* Add workaround to tests for Org Catalog publishing bug when dealing with LDAP [GH-458]
+* Add clean-up actions to some tests that were uploading vAppTemplates/medias to catalogs [GH-458]
--- a/govcd/adminorg_ldap_test.go
+++ b/govcd/adminorg_ldap_test.go
@@ -35,12 +35,33 @@ func (vcd *TestVCD) Test_LDAP(check *C) {
 	if vcd.config.VCD.ExternalNetwork == "" {
 		check.Skip("[" + check.TestName() + "] external network not provided")
 	}
+	// Due to a bug in VCD, when configuring LDAP service, Org publishing catalog settings `Publish external catalogs` and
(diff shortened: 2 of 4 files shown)


adambarreiro (approved):
Great, thanks a lot!

Didainius (approved):
Thanks - I have ran `catalog` test twice and it worked and also didn't clear down catalog publishing settings at Org level.

lvirbalas:
> Not sure if I should include changelog for this @lvirbalas @Didainius @vbauzysvmware @adambarreiro @datacharmer

Yes, please. As this is purely testing, you can add to the IMPROVEMENTS section. LGTM, otherwise :)

lvirbalas (approved):
(Just a changelog entry is needed.)

vbauzysvmware (approved):

PR #459, opened 2022-04-05
Add metadata for OpenAPI VDC networks

adambarreiro:
This PR contains a workaround for OpenAPI VDC networks **metadata** functions to work with XML endpoints while the final OpenAPI implementation remains unreleased.

The PR that contains the final implementation is #455

The other option that came in to my mind was not to do these functions that operate on a `OpenApiOrgVdcNetwork` object and just use the `OrgVDCNetwork` ones that we currently have, but the result was an spaghetti on Terraform Provider part that would be harder to clean afterwards.

This option just hides the details in the new provided functions, so refactoring later would be super easy (in fact is already done in #455, we just need to wait).

### DIFF
--- a/govcd/metadata.go
+++ b/govcd/metadata.go
@@ -842,13 +842,13 @@ func (orgVdcNetwork *OrgVDCNetwork) DeleteMetadataEntryAsync(key string) (Task, 
 
 // OpenAPI metadata functions
 
-// GetMetadata returns OrgVDCNetwork metadata.
+// GetMetadata returns OpenApiOrgVdcNetwork metadata.
 // TODO: This function is currently using XML underneath as metadata is supported in v37.0 and at the moment is in alpha state. See https://github.com/vmware/go-vcloud-director/pull/455
 func (openApiOrgVdcNetwork *OpenApiOrgVdcNetwork) GetMetadata() (*types.Metadata, error) {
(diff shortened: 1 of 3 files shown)


lvirbalas (approved):

mikeletux (approved):

govcd/metadata.go
```diff
@@ -839,3 +839,34 @@ func (orgVdcNetwork *OrgVDCNetwork) DeleteMetadataEntry(key string) error {
 func (orgVdcNetwork *OrgVDCNetwork) DeleteMetadataEntryAsync(key string) (Task, error) {
 	return deleteMetadata(orgVdcNetwork.client, key, strings.ReplaceAll(orgVdcNetwork.OrgVDCNetwork.HREF, "/api/", "/api/admin/"))
 }
+
+// OpenAPI metadata functions
+
+// GetMetadata returns OpenApiOrgVdcNetwork metadata.
+// TODO: This function is currently using XML underneath as metadata is supported in v37.0 and at the moment is in alpha state. See https://github.com/vmware/go-vcloud-director/pull/455
```
vbauzysvmware, line 846: Is comment(TODO) is correct, if not please correct in other functions also

vbauzysvmware (changes requested):

govcd/metadata.go
adambarreiro: Yep. The idea is that when v37.0 is released we can come back to this and improve these functions with the PR shown in the comment: #455 (basically removes XML in favor of OpenAPI)

adambarreiro: Improved the message to not to be confusing. Thanks!

vbauzysvmware (approved):

Didainius (approved):

PR #460, opened 2022-04-05
Add Vdc.GetNsxtAppPortProfileByName and VdcGroup.GetNsxtAppPortProfileByName for convenience

Didainius:
This PR adds two additional methods `Vdc.GetNsxtAppPortProfileByName` and `VdcGroup.GetNsxtAppPortProfileByName` for easier NSX-T Application Port Profile lookup.

Additionally it improves tests to test out lookup using VDC and VDC Group 

This will be used to improve https://github.com/vmware/terraform-provider-vcd/pull/812

### DIFF
--- a/govcd/nsxt_application_profile.go
+++ b/govcd/nsxt_application_profile.go
@@ -78,6 +78,35 @@ func (org *Org) GetNsxtAppPortProfileByName(name, scope string) (*NsxtAppPortPro
 	return getNsxtAppPortProfileByName(org.client, name, queryParameters)
 }
 
+// GetNsxtAppPortProfileByName allows users to retrieve Application Port Profiles for specific scope.
+// More details in documentation for types.NsxtAppPortProfile
+//
+// Note. Names are enforced to be unique per scope
+func (vdc *Vdc) GetNsxtAppPortProfileByName(name, scope string) (*NsxtAppPortProfile, error) {
+	queryParameters := copyOrNewUrlValues(nil)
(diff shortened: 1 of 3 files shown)


govcd/nsxt_application_profile.go
```diff
@@ -78,6 +78,35 @@ func (org *Org) GetNsxtAppPortProfileByName(name, scope string) (*NsxtAppPortPro
 	return getNsxtAppPortProfileByName(org.client, name, queryParameters)
 }
 
+// GetNsxtAppPortProfileByName allows users to retrieve Application Port Profiles for specific scope.
+// More details in documentation for types.NsxtAppPortProfile
+//
+// Note. Names are enforced to be unique per scope
+func (vdc *Vdc) GetNsxtAppPortProfileByName(name, scope string) (*NsxtAppPortProfile, error) {
+	queryParameters := copyOrNewUrlValues(nil)
```
vbauzysvmware, line 86: not sure need it, but looks like we could have a private function which accepts only ID which is later added to _context  :)

vbauzysvmware (approved):

govcd/nsxt_application_profile.go
Didainius: I'll probably refrain from it, because this function already calls a private function `getNsxtAppPortProfileByName` which would make it a "2 step" private function call

mikeletux (approved):
LGTM! Approved!

adambarreiro (approved):

Issue #461, opened 2022-04-05
vdc.GetMetadata() has index issues when accessing the vdcs as a system org user


bogi0704:
### Describe the bug

I am using the vdc.GetMetadata() funtion as a user belonging to the system organization. I get the following error when calling this as a system user:
```
runtime error: index out of range [1] with length 1
```
This is happening because the function GetMetadata() uses the function getAdminVdcURL to get the admin url. But the url this function gets is already an admin url, since I am accessing the Metadata as a system user.

### Reproduction steps

```bash
1. Authenticate as a system org user
2. try to call vdc.GetMetadata()
```

### Expected behavior

The vdc.GetMetadata() should be able to distinguish if it needs to get the admin url or just use the one that the vdc object has.

### Additional context

I will open a pull request to fix this issue.

adambarreiro:
Closing as done in #473

PR #462, opened 2022-04-05
added check before getting the admin url

bogi0704:
Related issue: #461 

This PR fixes a small bug which causes an index out of bounds error when calling it as a system org user. I have only added this code since there was no test for GetMetadata() itself.

### DIFF
--- a/govcd/metadata.go
+++ b/govcd/metadata.go
@@ -29,7 +29,7 @@ func (vm *VM) AddMetadata(key string, value string) (Task, error) {
 
 // GetMetadata returns Vdc metadata.
 func (vdc *Vdc) GetMetadata() (*types.Metadata, error) {
-	if strings.Contains(vdc.Vdc.HREF, "/api/admin/vdc/"){
+	if strings.Contains(vdc.Vdc.HREF, "/api/admin/vdc/") {
 		return getMetadata(vdc.client, vdc.Vdc.HREF)
 	} else {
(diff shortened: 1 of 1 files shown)


adambarreiro:
I'd bet that we have the same issue in other `metadata` functions that rely on `getAdminVdcURL(vdcURL string)`, my suggestion here would be to fix that function. Maybe instead of:

```
func getAdminVdcURL(vdcURL string) string {
	return strings.Split(vdcURL, "/api/vdc/")[0] + "/api/admin/vdc/" + strings.Split(vdcURL, "/api/vdc/")[1]
}
```

We could do:

```
func getAdminVdcURL(vdcURL string) string {
	return strings.ReplaceAll(vdcURL, "/api/vdc/", "/api/admin/vdc/"))
}
```

If I'm not wrong this should be working in all cases. What are you thoughts on this?

bogi0704:
You are right, there are probably other Methods using `getAdminVdcURL(vdcURL string)` which would benefit from this fix. I have tested your suggestion of the fix and it also works when the `vdcURL string` already contains `"/api/admin/vdc/"`. So nothing should break.

adambarreiro:
Hi,
Actually we did a huge refactor of the metadata handling, planned for v3.7.0, in PR #473. In that PR we also included this fix (in a more generic fashion). If you don't mind I think we can close this one.

PR #465, opened 2022-04-19
Add DNS option to the NSX-T Segment DHCP server

mikeletux:
This PR addresses the needs on `go-vcloud-director` to allow later feature implementation on `terraform-provider-vcd` about DHCP DNS on NSX-T routed networks  https://github.com/vmware/terraform-provider-vcd/issues/753

## Description
This PR allows govcd to be able to set and read DNS servers to/from the Dhcp configuration of a specific Org vDC routed network.

## Detailed description
Starting from VCD 10.3.1 (API v36.1) is possible to set the DNS servers that the DHCP sends when configured in a routed NSX-T network.

To allow this, a new field `DnsServers` has been added to the struct `OpenApiOrgVdcNetworkDhcp`. Also the methods `Vdc.GetOpenApiOrgVdcNetworkDhcp` and `Vdc.UpdateOpenApiOrgVdcNetworkDhcp` have been updated to use the method `Client.getOpenApiHighestElevatedVersion` instead of `Client.checkOpenApiEndpointCompatibility`, so if VCD 10.3.1 or newer is used, DNS can be set.

### DIFF
--- a/govcd/openapi_org_network_dhcp.go
+++ b/govcd/openapi_org_network_dhcp.go
@@ -27,7 +27,7 @@ func (vdc *Vdc) GetOpenApiOrgVdcNetworkDhcp(orgNetworkId string) (*OpenApiOrgVdc
 	queryParameters := queryParameterFilterAnd("orgVdc.id=="+vdc.Vdc.ID, params)
 
 	endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointOrgVdcNetworksDhcp
-	minimumApiVersion, err := client.checkOpenApiEndpointCompatibility(endpoint)
+	apiVersion, err := client.getOpenApiHighestElevatedVersion(endpoint)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


lvirbalas (approved):

adambarreiro (approved):
Great! 👍

govcd/openapi_org_network_dhcp.go
```diff
@@ -95,9 +95,6 @@ func (vdc *Vdc) UpdateOpenApiOrgVdcNetworkDhcp(orgNetworkId string, orgVdcNetwor
 
 // DeleteOpenApiOrgVdcNetworkDhcp allows to perform HTTP DELETE request on DHCP pool configuration for specified Org VDC
 // Network ID
-//
-// Note. VCD Versions before 10.2 do not allow to perform "DELETE" on DHCP pool and will return error. The way to
-// remove DHCP configuration is to recreate Org VDC network itself.
 func (vdc *Vdc) DeleteOpenApiOrgVdcNetworkDhcp(orgNetworkId string) error {
 	endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointOrgVdcNetworksDhcp
 	minimumApiVersion, err := vdc.client.checkOpenApiEndpointCompatibility(endpoint)
```
Didainius, line 100: While it shouldn't matter for DELETE I vote to use `getOpenApiHighestElevatedVersion` here as well so that all operations are using the same API version.

Didainius (approved):

govcd/openapi_org_network_dhcp.go
mikeletux: yeah, I agree. I'll change that!

lvirbalas: Thanks! I was expecting Dainius to have some comments on API version functions used too :)

mikeletux: Already changed @Didainius !!! Please check and resolve conversation if you agree 😄

vbauzysvmware (approved):

PR #478, opened 2022-05-26
Add Nsxt route advertisement methods

mikeletux:
This PR doesn't have any issue associated

## Description
This PR implements new NSX-T edge gateway methods that allow to create route advertisement in NSX-T edge gateways.

## Detailed description
This PR proposes three new methods for `NsxtEdgeGateway` that allows to manage route advertisement. These methods are:
- `NsxtEdgeGateway.GetNsxtRouteAdvertisement`
- `NsxtEdgeGateway.UpdateNsxtRouteAdvertisement`
- `NsxtEdgeGateway.DeleteNsxtRouteAdvertisement`

In order to use these methods, make sure you're using a dedicated Tier-0 gateway:
image

Or using code:
```
edge.EdgeGateway.EdgeGatewayUplinks[0].Dedicated = true
```

This PR also introduces a small change (docs) introduced by @vbauzysvmware in file [govcd/catalog_test.go](https://github.com/vmware/go-vcloud-director/pull/478/files#diff-f606b5f73940b6e9ae6b448549e5855a91680df2ad78391bbda0c7f8a8f34761)


govcd/nsxt_route_advertisement_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	org, err := vcd.client.GetOrgByName(vcd.config.VCD.Org)
+	check.Assert(err, IsNil)
+	nsxtVdc, err := org.GetVDCByName(vcd.config.VCD.Nsxt.Vdc, false)
+	check.Assert(err, IsNil)
+	edge, err := nsxtVdc.GetNsxtEdgeGatewayByName(vcd.config.VCD.Nsxt.EdgeGateway)
+	check.Assert(err, IsNil)
+
+	// Make sure we are using a dedicated Tier-0 gateway (otherwise route advertisement won't be available)
+	edge.EdgeGateway.EdgeGatewayUplinks[0].Dedicated = true
```
vbauzysvmware, line 28: let's check if this change is needed and return edge gtw to the same state as the test started to use. This allows us to avoid another test clash if in the case will use or depend on `Dedicated`

vbauzysvmware (approved):

govcd/tenant_context.go
```diff
@@ -183,3 +183,15 @@ func (vdcGroup *VdcGroup) getTenantContext() (*TenantContext, error) {
...
+func (egw *NsxtEdgeGateway) getTenantContext() (*TenantContext, error) {
+	if egw != nil && egw.EdgeGateway.Org != nil {
+		if egw.EdgeGateway.Org.Name == "" || egw.EdgeGateway.Org.ID == "" {
+			return nil, fmt.Errorf("either parent NsxtEdgeGateway Org name or ID is empty and both must be set. Org name is [%s] and Org ID is [%s]", egw.EdgeGateway.Org.Name, egw.EdgeGateway.Org.ID)
+		}
+
+		return &TenantContext{OrgId: egw.EdgeGateway.Org.ID, OrgName: egw.EdgeGateway.Org.Name}, nil
+	}
+
+	return nil, fmt.Errorf("NsxtEdgeGateway is not fully initialized. Please initialized before using this method")
```
adambarreiro, line 196: Just small typo
```suggestion
	return nil, fmt.Errorf("NsxtEdgeGateway is not fully initialized. Please initialize it before using this method")
```

adambarreiro (approved):
LGTM!

lvirbalas (approved):

govcd/nsxt_route_advertisement_test.go
Didainius: +1 to this

Make sure to restore this setting post-test because we have other tests that create additional NSX-T Edge Gateways using the same external network (T0 router). When it becomes dedicated - no other Edge Gateways can be created using the same external network. Other tests would fail.

I think even a `defer` cleanup function would be good to define here in case the test breaks in between.

mikeletux: Fixed. Please check https://github.com/mikeletux/go-vcloud-director/blob/31e5d2910eb6936914e847d408c6440585279b2d/govcd/nsxt_route_advertisement_test.go#L27-L33

govcd/tenant_context.go
mikeletux: Changed! Thanks for this catch! 🚀

types/v56/nsxt_types.go
```diff
@@ -1244,3 +1244,13 @@ type EntitySecurityTags struct {
 	// Tags is the list of tags. The value is case-agnostic and will be converted to lower-case.
 	Tags []string `json:"tags"`
 }
+
+// RouteAdvertisement lists the subnets that will be advertised so that the Edge Gateway can route out to the
+// connected external network.
+type RouteAdvertisement struct {
+	// Enable if true, means that the subnets will be advertised. The default is true.
```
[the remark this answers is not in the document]
mikeletux: Changed comment as we agreed!

Didainius (approved):

vbauzysvmware (approved):

PR #480, opened 2022-06-03
Add support for BGP Configuration on NSX-T Edge Gateways

Didainius:
This PR adds the initial structure for NSX-T Edge Gateway BGP Configuration. It reflects the "Configuration" tab in UI.

image

Additional test fixes to make test suite run better:
* Route advertisement code uncovered a VCD flaw with tenant context IDs. Apparently, VCD 10.2.X does not accept URN formatted Org ID (only UUID). VCD 10.3 works fine with both. Commit `f57dd01` introduced a global fix for this.
* Some of the previous tests were leaking vApp templates and media items that would break `Test_QueryVappTemplateList` and `Test_CatalogQueryMediaList`. These tests are now patched to dump structures for these leaked templates so that they can be caught easier. As a result of this lookup these tests were patched to properly cleanup:
  * `Test_AddAndDeleteMetadataOnMediaRecord`
  * `Test_AddNewVMFromMultiVmTemplate` 
  * `Test_AnswerVmQuestion`
* All VCD versions currently have a bug that disables Catalog publishing on Org update. As a result `Test_PublishToExternalOrganizations` always fails as it assumes that publishing is enabled. To overcome it I added explicit `patch` for `CanPublishCatalogs=true` and `CanPublishExternally=true` before this test run.
* There have been cases when ALB tests were probably hit by timing issues (looks like VCD needs a wait before kicking in a task to enable ALB on Edge Gateway). In this PR test helper `setupAlbPoolPrerequisites` is patched to perform cleanups better and avoid subsequent test failure. This should help tests to run cleaner and also find potential errors in the future.


govcd/nsxt_edgegateway_bgp_configuration.go
```diff
@@ -0,0 +1,78 @@
...
+
+package govcd
+
+import (
+	"fmt"
+
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+)
+
+func (egw *NsxtEdgeGateway) GetBgpConfiguration() (*types.EdgeBgpConfig, error) {
```
adambarreiro, line 13: Shouldn't we add documentation for the public functions?

govcd/nsxt_edgegateway_bgp_configuration.go
```diff
@@ -0,0 +1,78 @@
...
+	err = client.OpenApiPutItem(apiVersion, urlRef, nil, bgpConfig, returnObject, nil)
+	if err != nil {
+		return nil, fmt.Errorf("error setting NSX-T Edge Gateway BGP Configuration: %s", err)
+	}
+
+	return returnObject, nil
+}
+
+func (egw *NsxtEdgeGateway) DisableBgpConfiguration() error {
+	// Get existring BGP configuration so that when disabling it - other settings remain as they are
```
adambarreiro, line 69:
```suggestion
	// Get existing BGP configuration so that when disabling it - other settings remain as they are
```

govcd/nsxt_edgegateway_bgp_configuration_test.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	updatedBgpConfig, err := edge.UpdateBgpConfiguration(newBgpConfig)
+	check.Assert(err, IsNil)
+	check.Assert(updatedBgpConfig, NotNil)
+
+	newBgpConfig.Version = updatedBgpConfig.Version // Version is constantly iterated and cant be checked
+	check.Assert(updatedBgpConfig, DeepEquals, newBgpConfig)
+
+	// Check "disable" function which keeps the all fields the same, but sets "Enabled: false"
```
adambarreiro, line 54:
```suggestion
	// Check "disable" function which keeps all fields the same, but sets "Enabled: false"
```

Didainius: Fixed.

govcd/nsxt_edgegateway_bgp_configuration.go
Didainius: Fixed.

govcd/nsxt_edgegateway_bgp_configuration.go
Didainius: Totally - this is my oversight. Fixed.

adambarreiro (approved):

govcd/nsxt_edgegateway_bgp_configuration_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// Get and store existing BGP configuration
+	bgpConfig, err := edge.GetBgpConfiguration()
+	check.Assert(err, IsNil)
+	check.Assert(bgpConfig, NotNil)
+
+	// Disable BGP
+	err = edge.DisableBgpConfiguration()
+	check.Assert(err, IsNil)
+
```
vbauzysvmware, line 35: assert that disable happen missing

vbauzysvmware: I saw same at the end of test, so not sure why this needed here

govcd/tenant_context.go
```diff
@@ -152,7 +152,11 @@ func getTenantContextHeader(tenantContext *TenantContext) map[string]string {
 		return nil
 	}
 	return map[string]string{
-		types.HeaderTenantContext: tenantContext.OrgId,
+		// Some versions of VCD do not like when URN is sent for Tenant context ID - they fail with
```
vbauzysvmware, line 155: I would like to  name exact versions, not "some":))

vbauzysvmware: what about new VCD versions? is it ok to pass ID no URN? as change seem global

types/v56/nsxt_types.go
```diff
@@ -1254,3 +1254,60 @@ type RouteAdvertisement struct {
...
+	// StaleRouteTimer defines maximum time (in seconds) before stale routes are removed when BGP
+	// restarts.
+	StaleRouteTimer int `json:"staleRouteTimer"`
+}
+
+// EdgeBgpConfigVersion is part of EdgeBgpConfig type and describes current version of the entity
+// being modified
+type EdgeBgpConfigVersion struct {
+	Version int `json:"version"`
+}
```
vbauzysvmware, line 1313: note: seems to me file grow is sufficient for a start to think how to split it to more granularly domains :)

vbauzysvmware (changes requested):
Nice additions to the main PR topic. Yet, is always a philosophical question that should be in separate PR or not :)

govcd/nsxt_edgegateway_bgp_configuration_test.go
Didainius: This first `disable` happened just to be sure that it is disabled before testing configuration. BGP gets enabled by default in some cases.

govcd/tenant_context.go
Didainius: All 10.2.X versions - I have updated the comment to reflect that. Change is global as and the other tests which use tenant context validate that both approaches work.

lvirbalas (approved):
Just one ask and looks great!

.changes/v2.16.0/478-features.md
```diff
@@ -1 +1 @@
-* Add NSX-T Edge Gateway methods `NsxtEdgeGateway.GetNsxtRouteAdvertisement`, `NsxtEdgeGateway.GetNsxtRouteAdvertisementWithContext`, `NsxtEdgeGateway.UpdateNsxtRouteAdvertisement`, `NsxtEdgeGateway.UpdateNsxtRouteAdvertisementWithContext`, `NsxtEdgeGateway.DeleteNsxtRouteAdvertisement` and `NsxtEdgeGateway.DeleteNsxtRouteAdvertisementWithContext` that allows to manage NSX-T route advertisement [GH-478]
+* Add NSX-T Edge Gateway methods `NsxtEdgeGateway.GetNsxtRouteAdvertisement`, `NsxtEdgeGateway.GetNsxtRouteAdvertisementWithContext`, `NsxtEdgeGateway.UpdateNsxtRouteAdvertisement`, `NsxtEdgeGateway.UpdateNsxtRouteAdvertisementWithContext`, `NsxtEdgeGateway.DeleteNsxtRouteAdvertisement` and `NsxtEdgeGateway.DeleteNsxtRouteAdvertisementWithContext` that allows to manage NSX-T route advertisement [GH-478, GH-480]
```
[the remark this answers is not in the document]
Didainius: Fixed.

PR #485, opened 2022-06-24
Update ALB service engine group and ALB settings to support SupportedFeatureSet in VCD 10.4

adambarreiro:
This PR is not associated to any issue.

## Description

When executing the tests using VCD 10.4, I found that some of them were failing due to:

- ALB Controller does not use `LicenseType` attribute in 10.4 (v37.0)
- ALB Service Engine Group has a new `SupportedFeatureSet` in 10.4 (v37.0) that replaces the above `LicenseType`
- ALB Settings has a new `SupportedFeatureSet` in 10.4 (v37.0) that replaces the above `LicenseType`

The goal of this PR is to adapt the mentioned data structures to **make the Go SDK compatible with v37.0** of the API (VCD 10.4).

In consequence, the failing tests have been adapted and some functions now use `getOpenApiHighestElevatedVersion` to support both old and new scenarios.

## Tests

- [x] Tests pass on VCD 10.2
- [x] Tests pass on VCD 10.3
- [x] Tests pass on VCD 10.4

### DIFF
--- a/.changes/v2.16.0/485-improvements.md
+++ b/.changes/v2.16.0/485-improvements.md
@@ -1,1 +1,2 @@
-* Add `SupportedFeatureSet` attribute to `NsxtAlbServiceEngineGroup` and `NsxtAlbConfig` to support v37.0 license management for AVI Load Balancer [GH-485]
+* Add `SupportedFeatureSet` attribute to `NsxtAlbServiceEngineGroup` and `NsxtAlbConfig` to support v37.0 license management for AVI Load Balancer
+and replace `LicenseType` from `NsxtAlbController` [GH-485]
--- a/.changes/v2.16.0/485-improvements.md
+++ b/.changes/v2.16.0/485-improvements.md
@@ -1,1 +1,1 @@
-* Add `SupportedFeatureSet` attribute to `NsxtAlbServiceEngineGroup` to support v37.0 AVI Load Balancer license management [GH-485]
+* Add `SupportedFeatureSet` attribute to `NsxtAlbServiceEngineGroup` and `NsxtAlbConfig` to support v37.0 license management for AVI Load Balancer [GH-485]
--- a/.changes/v2.16.0/485-improvements.md
+++ b/.changes/v2.16.0/485-improvements.md
@@ -1,0 +1,1 @@
+* Add `SupportedFeatureSet` attribute to `NsxtAlbServiceEngineGroup` to support v37.0 AVI Load Balancer license management [GH-485]
--- a/govcd/api_vcd_versions.go
+++ b/govcd/api_vcd_versions.go
@@ -46,7 +46,8 @@ var apiVersionToVcdVersion = map[string]string{
 	"33.0": "10.0",
 	"34.0": "10.1",
 	"35.0": "10.2",
-	"36.0": "10.3", // Provisional version for non-GA release. It may change later
+	"36.0": "10.3",
+	"37.0": "10.4", // Provisional version for non-GA release. It may change later
 }
 
 // vcdVersionToApiVersion gets the max supported API version from vCD version
@@ -58,7 +59,8 @@ var vcdVersionToApiVersion = map[string]string{
 	"10.0": "33.0",
 	"10.1": "34.0",
 	"10.2": "35.0",
-	"10.3": "36.0", // Provisional version for non-GA release. It may change later
+	"10.3": "36.0",
+	"10.4": "37.0", // Provisional version for non-GA release. It may change later
 }
 
 // to make vcdVersionToApiVersion used
--- a/govcd/nsxt_alb_controllers_test.go
+++ b/govcd/nsxt_alb_controllers_test.go
@@ -63,7 +63,7 @@ func (vcd *TestVCD) Test_NsxtAlbController(check *C) {
 		Url:         vcd.config.VCD.Nsxt.NsxtAlbControllerUrl,
 		Username:    vcd.config.VCD.Nsxt.NsxtAlbControllerUser,
 		Password:    vcd.config.VCD.Nsxt.NsxtAlbControllerPassword,
-		LicenseType: "BASIC",
+		LicenseType: "BASIC", // Not used since v37.0
 	}
 	updatedController, err := controllerByUrl.Update(updateControllerDef)
 	check.Assert(err, IsNil)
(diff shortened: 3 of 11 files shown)


govcd/nsxt_alb_controllers_test.go
```diff
@@ -71,7 +71,10 @@ func (vcd *TestVCD) Test_NsxtAlbController(check *C) {
 	check.Assert(updatedController.NsxtAlbController.Description, Equals, updateControllerDef.Description)
 	check.Assert(updatedController.NsxtAlbController.Url, Equals, updateControllerDef.Url)
 	check.Assert(updatedController.NsxtAlbController.Username, Equals, updateControllerDef.Username)
-	check.Assert(updatedController.NsxtAlbController.LicenseType, Equals, updateControllerDef.LicenseType)
+	// LicenseType has been removed since v37.0
+	if vcd.client.Client.APIVCDMaxVersionIs("< 37.0") {
+		check.Assert(updatedController.NsxtAlbController.LicenseType, Equals, updateControllerDef.LicenseType)
```
adambarreiro, line 76: The funny thing here is that LicenseType is populated in v37.0 as "BASIC", no matter what payload you send. I guess it was done this way for compatibility reasons. That's why this test didn't fail, as the update operation here precisely tested a change from "ENTERPRISE" to "BASIC".

This condition was added to make the test more correct, as you can see it was passing on v37.0 because of coincidence.

adambarreiro:
Moving back to Draft as I got some test results today that showed that also changes in Edge Gateway must also be done.

.changes/v2.16.0/485-improvements.md
```diff
@@ -0,0 +1 @@
+* Add `SupportedFeatureSet` attribute to `NsxtAlbServiceEngineGroup` and `NsxtAlbConfig` to support v37.0 license management for AVI Load Balancer [GH-485]
```
adambarreiro, line 1: Should I add `LicenseType` to deprecations section? It is still present in the used structs and by the API in the newest VCD, although seems ignored, so that's why I didn't.

govcd/api_vcd_versions.go
```diff
@@ -46,7 +46,8 @@ var apiVersionToVcdVersion = map[string]string{
 	"33.0": "10.0",
 	"34.0": "10.1",
 	"35.0": "10.2",
-	"36.0": "10.3", // Provisional version for non-GA release. It may change later
+	"36.0": "10.3",
+	"37.0": "10.4", // Provisional version for non-GA release. It may change later
```
adambarreiro, line 50: Not sure if I need to change some other file.

types/v56/nsxt_types.go
```diff
@@ -704,6 +704,7 @@ type NsxtAlbController struct {
 	// LicenseType By enabling this feature, the provider acknowledges that they have independently licensed the
 	// enterprise version of the NSX AVI LB.
 	// Possible options: 'BASIC', 'ENTERPRISE'
+	// This field was removed since VCD 10.4.0 (v37.0) in favor of NsxtAlbServiceEngineGroup.SupportedFeatureSet
 	LicenseType string `json:"licenseType,omitempty"`
```
adambarreiro, line 708: I had doubts about changing to pointer here, but as the newest VCD doesn't complain about sending/receiving it, I didn't.
That's the main reason `SupportedFeatureSet` is a pointer, avoids breaking <v37.0

lvirbalas:
Please expand the Description of this PR with context of why this is needed, what has changed in VCD and in which version. Thanks!

.changes/v2.16.0/485-improvements.md
Didainius: I think it is not worth deprecating `LicenseType` yet. 
The point is - all currently supported VCD versions use it and it is correct usage from that point. Marking this field as deprecated in our schema would tell that this field is deprecated, but there is no replacement (until one goes beyond VCD 10.3). Let's see how it goes, but IMO deprecation should happen later on.

Didainius: Just an extra sentence that `SupportedFeatureSet` field in fact replaces `LicenseType` in V37.0. That one - whoever is skimming changelog should have enough information to get triggered about it (or not).

govcd/api_vcd_versions.go
Didainius: I think nothing more should be needed. In fact, these mappings were mainly meant for XML based APIs and custom version IIRC. 

OpenAPI endpoints rely on `endpointElevatedApiVersions` (which you already updated)

types/v56/nsxt_types.go
Didainius: I think we can avoid adding a pointer (which could cause someone an unnecessary change when upgrading SDK ).

Go JSON lib does not send `string` type with empty value `""` when `omitempty` is present.

govcd/nsxt_distributed_firewall_test.go
```diff
@@ -265,7 +265,7 @@ func getRandomListOfNetworkContextProfiles(check *C, vcd *TestVCD, vdcClient *VC
 	check.Assert(err, IsNil)
 	openApiRefs := make([]types.OpenApiReference, 1)
 	for _, networkContextProfile := range networkContextProfiles {
-		if strings.Contains(networkContextProfile.Description, "ALG") {
+		if strings.Contains(networkContextProfile.Description, "ALG") || strings.Contains(networkContextProfile.Description, "includes the URL categories") {
```
Didainius, line 268: FYI @lvirbalas 
We talked about this patch.

@adambarreiro ,
Could you add a brief comment here "// skipping network context profile which has hardcoded destinations and throws error when used in firewall rules with specified destinations"

types/v56/nsxt_types.go
```diff
@@ -827,10 +831,13 @@ type NsxtAlbConfig struct {
 	// LicenseType of the backing Load Balancer Cloud.
 	// * BASIC - Basic edition of the NSX Advanced Load Balancer.
 	// * ENTERPRISE - Full featured edition of the NSX Advanced Load Balancer.
+	// This field was removed since VCD 10.4.0 (v37.0) in favor of NsxtAlbConfig.SupportedFeatureSet
 	LicenseType string `json:"licenseType,omitempty"`
+	// SupportedFeatureSet was added in VCD 10.4.0 (v37.0) as substitute of NsxtAlbConfig.LicenseType.
+	// Possible values are: "STANDARD", "PREMIUM".
+	SupportedFeatureSet *string `json:"supportedFeatureSet,omitempty"`
```
Didainius, line 838: You can double-check here but I think having simple `string` will not send the field if `omitempty` is set and value is `""` (empty string)

Didainius (approved):
I left a few non major comments and approving.

govcd/nsxt_distributed_firewall_test.go
adambarreiro: Done!

PR #487, opened 2022-06-30
Add support for Dynamic Security Groups

Didainius:
This PR adds support for Dynamic Security Groups (or as they are called in API - Firewall Groups).
It reuses the same functions we already have (because IP sets, Static Security Groups and Dynamic Security Groups use the same API), but expands a few places to accommodate Dynamic Security Group handling in VCD 10.3+:
* Expand `types.NsxtFirewallGroup` with new fields introduced with API V36.0 (VCD 10.3)
* New constant `types.FirewallGroupTypeVmCriteria` to match Dynamic Security Group type of Firewall group.
* API elevation to 36.0 when available so that Dynamic security groups can be handled.
* Additional test `Test_NsxtDynamicSecurityGroup` to checkout Dynamic security group creation.
* `NsxtFirewallGroup.IsDynamicSecurityGroup()` function to validate that given Firewall Group is Dynamic Security Group
* Additional Firewall Group functions with `*VdcGroup` pointer receiver
* Internal function `getFirewallGroupTypeFilterFieldName` for embedding into Security Group management functions as new field `TypeValue` must be used instead of now deprecated one `Type` which doesn't support Dynamic Firewall Groups.

More about Dynamic Security Groups in VCD official docs and VMware Blog post:
* https://docs.vmware.com/en/VMware-Cloud-Director/10.3/VMware-Cloud-Director-Tenant-Portal-Guide/GUID-CF7FCFE1-7B1C-45CE-B8A0-774E9CD31DB5.html
* https://blogs.vmware.com/cloudprovider/2021/08/vmware-cloud-director-10-3-new-networking-security-features.html

### DIFF
--- a/.changes/v2.16.0/487-features.md
+++ b/.changes/v2.16.0/487-features.md
@@ -1,0 +1,4 @@
+* Add support for Dynamic Security Groups in VCD 10.3 by expanding `types.NsxtFirewallGroup` to
+  accommodate fields required for dynamic security groups, implemented automatic API elevation to
+  v36.0. Added New functions `VdcGroup.CreateNsxtFirewallGroup`,
+  `NsxtFirewallGroup.IsDynamicSecurityGroup` [GH-487]
--- a/.changes/v2.16.0/487-features.md
+++ b/.changes/v2.16.0/487-features.md
@@ -1,4 +1,4 @@
 * Add support for Dynamic Security Groups in VCD 10.3 by expanding `types.NsxtFirewallGroup` to
-  accommodate fields required for dynamic security groups, implemented automatic API elevation to
+  accommodate fields required for Dynamic Security Groups, implemented automatic API elevation to
   v36.0. Added New functions `VdcGroup.CreateNsxtFirewallGroup`,
   `NsxtFirewallGroup.IsDynamicSecurityGroup` [GH-487]
--- a/govcd/nsxt_firewall_group.go
+++ b/govcd/nsxt_firewall_group.go
@@ -36,6 +36,10 @@ func (egw *NsxtEdgeGateway) CreateNsxtFirewallGroup(firewallGroupConfig *types.N
 	return createNsxtFirewallGroup(egw.client, firewallGroupConfig)
 }
 
+func (vdcGroup *VdcGroup) CreateNsxtFirewallGroup(firewallGroupConfig *types.NsxtFirewallGroup) (*NsxtFirewallGroup, error) {
+	return createNsxtFirewallGroup(vdcGroup.client, firewallGroupConfig)
+}
+
 // GetAllNsxtFirewallGroups allows users to retrieve all Firewall Groups for Org
 // firewallGroupType can be one of the following:
 // * types.FirewallGroupTypeSecurityGroup - for NSX-T Security Groups
@@ -57,8 +61,9 @@ func (egw *NsxtEdgeGateway) CreateNsxtFirewallGroup(firewallGroupConfig *types.N
 // 'networkProviderId' is NSX-T manager ID
 func (org *Org) GetAllNsxtFirewallGroups(queryParameters url.Values, firewallGroupType string) ([]*NsxtFirewallGroup, error) {
 	queryParams := copyOrNewUrlValues(queryParameters)
+	filteringTypeFieldName := getFirewallGroupTypeFilterFieldName(org.client)
 	if firewallGroupType != "" {
-		queryParams = queryParameterFilterAnd("type=="+firewallGroupType, queryParams)
+		queryParams = queryParameterFilterAnd(fmt.Sprintf("%s==%s", filteringTypeFieldName, firewallGroupType), queryParameters)
 	}
 
 	return getAllNsxtFirewallGroups(org.client, queryParams)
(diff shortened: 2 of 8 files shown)


govcd/nsxt_firewall_group.go
```diff
@@ -219,7 +231,7 @@ func (firewallGroup *NsxtFirewallGroup) Update(firewallGroupConfig *types.NsxtFi
 // Delete allows users to delete NSX-T Firewall Group
 func (firewallGroup *NsxtFirewallGroup) Delete() error {
 	endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointFirewallGroups
-	minimumApiVersion, err := firewallGroup.client.checkOpenApiEndpointCompatibility(endpoint)
+	apiVersion, err := firewallGroup.client.getOpenApiHighestElevatedVersion(endpoint)
```
lvirbalas, line 234: Just being cautious: after changing the way we chose API versions across this PR, does it work with the latest 10.4?

govcd/nsxt_firewall_group_dynamic_security_group_test.go
```diff
@@ -0,0 +1,103 @@
...
+		check.Skip("Dynamic security groups require VCD 10.3+")
+	}
+
+	adminOrg, err := vcd.client.GetAdminOrgByName(vcd.config.VCD.Org)
+	check.Assert(err, IsNil)
+
+	vdcGroup, err := adminOrg.GetVdcGroupByName(vcd.config.VCD.Nsxt.VdcGroup)
+	check.Assert(err, IsNil)
+
+	// vdcGroupEdgeGateway.
```
lvirbalas, line 29: Strange comment :)

govcd/nsxt_firewall_group_dynamic_security_group_test.go
```diff
@@ -0,0 +1,103 @@
...
+	createdDynamicGroup.NsxtFirewallGroup.Name = check.TestName() + "-updated"
+
+	updatedDynamicGroup, err := createdDynamicGroup.Update(createdDynamicGroup.NsxtFirewallGroup)
+	check.Assert(err, IsNil)
+	check.Assert(updatedDynamicGroup, NotNil)
+	check.Assert(updatedDynamicGroup.NsxtFirewallGroup, DeepEquals, createdDynamicGroup.NsxtFirewallGroup)
+
+	check.Assert(updatedDynamicGroup, DeepEquals, createdDynamicGroup)
+
+	// Remove
```
lvirbalas, line 100: Remove what? :)

types/v56/nsxt_types.go
```diff
@@ -255,7 +255,7 @@ type OpenApiOrgVdcNetworkDhcpPools struct {
 type NsxtFirewallGroup struct {
 	// ID contains Firewall Group ID (URN format)
 	// e.g. urn:vcloud:firewallGroup:d7f4e0b4-b83f-4a07-9f22-d242c9c0987a
-	ID string `json:"id"`
+	ID string `json:"id,omitempty"`
```
lvirbalas, line 258: It's OK to omit ID?

lvirbalas (changes requested):
Just a few asks :)

.changes/v2.16.0/487-features.md
```diff
@@ -0,0 +1,4 @@
+* Add support for Dynamic Security Groups in VCD 10.3 by expanding `types.NsxtFirewallGroup` to
+  accommodate fields required for dynamic security groups, implemented automatic API elevation to
```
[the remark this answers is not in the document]
Didainius: Fixed.

govcd/nsxt_firewall_group.go
Didainius: Testing will cover 10.4, but if you remember - we introduced "smart" API elevation back with PR #399. 

The point of it is that it doesn't "dumbly" just pick the latest API version available. We only pick "highest supported version" from our predefined list and this PR does it here: https://github.com/vmware/go-vcloud-director/pull/487/files#diff-d48dd6aa67017584bb6e37e8e459a46748e9f2660ff686da9a84b3bb18449792R99-R102

```go
types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointFirewallGroups: {
   //"34.0", // Basic minimum required version
   "36.0", // Adds support for Dynamic Security Groups by deprecating `Type` field in favor of `TypeValue`
},
```


This definition ensures that we will not elevate API above v36.0 even if v39.0 is supported in the future. So unless VCD itself breaks API V36.0 - we are good to go. 

We only elevate to known and tested API versions.

govcd/nsxt_firewall_group_dynamic_security_group_test.go
Didainius: Yes. A leftover as I was planning on testing creation of Dynamic Security Groups using `vdcGroupEdgeGateway` entity, but in reality Dynamic Security Groups are bound to VDC Groups, not to Edge Gateways. Forgot to delete that and overlooked while self reviewing.

govcd/nsxt_firewall_group_dynamic_security_group_test.go
Didainius: Updated.

types/v56/nsxt_types.go
Didainius: Yes. In general, we usually put `omitempty` for `ID` fields. The reason is - during a "Create" operation we don't have ID and it is better not to send the field at all than send an empty string. There were some cases where API even breaks getting an `ID` field, but that is not this particular case. I just saw that I left it some time ago and thought to align it - if there was any problem - tests would catch it.

lvirbalas (approved):

adambarreiro (approved):

PR #483, opened 2022-06-10
XML tags fixed for correct unmarshalling of ResourcePoolRefs in AdminVdc and test added

ArnoSen:
## IMPORTANT

To help us process your pull request efficiently, please follow the 
guidelines shown below. 

## A Pull Request should be associated with an Issue

We wish to have discussions in Issues. A single issue may be targeted by multiple PRs.
If you're offering a new feature or fixing anything, we'd like to know beforehand in Issues,
and potentially we'll be able to point development in a particular direction.

We accept PRs without associated issues provided the change is sufficiently evident 
from the commit message. If you have typos or simple bug fixes go for it.

## Description

Related issue: [](https://github.com/vmware/go-vcloud-director/issues/482)

- (Required) Short description of changes in the PR subject

- (Required) Detailed description of changes include tests and
  documentation. If the pull request contains multiple commits with 
  detailed messages, refer to those instead

XML tags have been corrected and a test has been added that shows the xml is properly unmarshalled.

- (Optional) Names of reviewers using @ sign + name

PR #494, opened 2022-08-08
Add ResourcePoolRefs read support for AdminVdc type

Didainius:
This PR builds on top of #483 to fix the problem with `AdminVdc.ResourcePoolRefs` read capabilities (this is a read-only field).

There is a Go marshaling problem https://github.com/golang/go/issues/9519 which makes marshaling/unmarshaling non symmetric, but as this field is read-only - we care about  reading it.
So the goal is to:
* Be able to read resources pools backing VDCs
* Avoid breaking code on `create/update` operations

Test `go test -tags "org vdc" -check.vv  -timeout=255m .` passed

Closes #483

### DIFF
--- a/.changes/v2.17.0/494-improvements.md
+++ b/.changes/v2.17.0/494-improvements.md
@@ -1,0 +1,1 @@
+Fix type `types.AdminVdc.ResourcePoolRefs` to make unmarshaling work (read-only) [GH-494]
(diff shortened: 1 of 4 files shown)


lvirbalas (approved):
Nice catch!

adambarreiro (approved):
LGTM!

PR #495, opened 2022-08-10
Add vApp Template functions

adambarreiro:
Signed-off-by: abarreiro <abarreiro@vmware.com>

## IMPORTANT

guidelines shown below. 

We wish to have discussions in Issues. A single issue may be targeted by multiple PRs.
If you're offering a new feature or fixing anything, we'd like to know beforehand in Issues,
and potentially we'll be able to point development in a particular direction.

We accept PRs without associated issues provided the change is sufficiently evident 
from the commit message. If you have typos or simple bug fixes go for it.

## Description

Related issue: (`<URL or #NUMBER of your Issue>`)

- (Required) Short description of changes in the PR subject

- (Required) Detailed description of changes include tests and
  documentation. If the pull request contains multiple commits with 
  detailed messages, refer to those instead

- (Optional) Names of reviewers using @ sign + name

### DIFF
--- a/govcd/catalog.go
+++ b/govcd/catalog.go
@@ -864,6 +864,21 @@ func (cat *Catalog) GetCatalogItemByName(catalogItemName string, refresh bool) (
 	return nil, ErrorEntityNotFound
 }
 
+// GetVappTemplateByName finds a VAppTemplate by Name
+// On success, returns a pointer to the VAppTemplate structure and a nil error
+// On failure, returns a nil pointer and an error
+func (cat *Catalog) GetVappTemplateByName(vAppTemplateName string, refresh bool) (*VAppTemplate, error) {
+	catalogItem, err := cat.GetCatalogItemByName(vAppTemplateName, refresh)
+	if err != nil {
+		return nil, err
+	}
+	vAppTemplate, err := catalogItem.GetVAppTemplate()
+	if err != nil {
+		return nil, err
+	}
+	return &vAppTemplate, nil
+}
+
 // GetCatalogItemById finds a Catalog Item by ID
 // On success, returns a pointer to the CatalogItem structure and a nil error
 // On failure, returns a nil pointer and an error
@@ -884,7 +899,20 @@ func (cat *Catalog) GetCatalogItemById(catalogItemId string, refresh bool) (*Cat
 	return nil, ErrorEntityNotFound
 }
 
-// GetCatalogItemByNameOrId finds a Catalog Item by Name or ID
+// GetVappTemplateById finds a vApp Template by ID.
+// On success, returns a pointer to the VAppTemplate structure and a nil error.
+// On failure, returns a nil pointer and an error.
+func (cat *Catalog) GetVappTemplateById(vAppTemplateId string) (*VAppTemplate, error) {
+	vAppTemplateHref := fmt.Sprintf("%s/vAppTemplate/vappTemplate-%s", cat.client.VCDHREF.String(), strings.ReplaceAll(vAppTemplateId, "urn:vcloud:vapptemplate:", ""))
+	vappTemplate, err := cat.GetVappTemplateByHref(vAppTemplateHref)
+	if err != nil {
+		return nil, fmt.Errorf("could not find vApp Template with ID %s: %s", vAppTemplateId, err)
+	}
+	return vappTemplate, nil
+
+}
+
+// GetCatalogItemByNameOrId finds a Catalog Item by Name or ID.
 // On success, returns a pointer to the CatalogItem structure and a nil error
 // On failure, returns a nil pointer and an error
 func (cat *Catalog) GetCatalogItemByNameOrId(identifier string, refresh bool) (*CatalogItem, error) {
@@ -895,6 +923,19 @@ func (cat *Catalog) GetCatalogItemByNameOrId(identifier string, refresh bool) (*
 		return nil, err
 	}
 	return entity.(*CatalogItem), err
+}
+
+// GetVappTemplateByNameOrId finds a vApp Template by Name or ID.
+// On success, returns a pointer to the VAppTemplate structure and a nil error.
+// On failure, returns a nil pointer and an error.
+func (cat *Catalog) GetVappTemplateByNameOrId(identifier string, refresh bool) (*VAppTemplate, error) {
+	getByName := func(name string, refresh bool) (interface{}, error) { return cat.GetVappTemplateByName(name, refresh) }
+	getById := func(id string, refresh bool) (interface{}, error) { return cat.GetVappTemplateById(id) }
+	entity, err := getEntityByNameOrId(getByName, getById, identifier, refresh)
+	if entity == nil {
+		return nil, err
+	}
+	return entity.(*VAppTemplate), err
 }
 
 // QueryMediaList retrieves a list of media items for the catalog
--- a/govcd/vapptemplate.go
+++ b/govcd/vapptemplate.go
@@ -6,6 +6,7 @@ package govcd
 
 import (
 	"fmt"
+	"github.com/vmware/go-vcloud-director/v2/util"
 	"net/http"
 	"net/url"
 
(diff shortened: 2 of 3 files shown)


adambarreiro:
I'm checking why `make static` returns exit code 2, in my local it returns 0. 🤔

lvirbalas (approved):
Just one ask: if you have an issue number for which this PR is preparing, please add to the description.

And LGTM! :)

adambarreiro:
> Just one ask: if you have an issue number for which this PR is preparing, please add to the description.
> 
> And LGTM! :)

Couldn't find any issue, but added a phrase to point to the resulting PR in **terraform-provider-vcd** (which has a related issue indeed)

govcd/vapptemplate.go
```diff
@@ -123,3 +125,16 @@ func (vAppTemplate *VAppTemplate) UpdateAsync() (Task, error) {
 	return vAppTemplate.client.ExecuteTaskRequest(url, http.MethodPut,
 		types.MimeVAppTemplate, "error updating vApp template item: %s", vappTemplatePayload)
 }
+
+// Delete deletes the VAppTemplate Item, returning an error if the vCD call fails.
```
Didainius, line 129: Nit. The `vCD` comes from the vCloud Director era. We're not VMware Cloud Director :)

```suggestion
// Delete deletes the VAppTemplate Item, returning an error if the VCD call fails.
```

govcd/catalog.go
```diff
@@ -884,7 +899,22 @@ func (cat *Catalog) GetCatalogItemById(catalogItemId string, refresh bool) (*Cat
 	return nil, ErrorEntityNotFound
 }
 
-// GetCatalogItemByNameOrId finds a Catalog Item by Name or ID
+// GetVAppTemplateById finds a vApp Template by ID.
+// On success, returns a pointer to the VAppTemplate structure and a nil error.
+// On failure, returns a nil pointer and an error.
+func (cat *Catalog) GetVAppTemplateById(vAppTemplateId string) (*VAppTemplate, error) {
+	vappTemplateHref := cat.client.VCDHREF
+	vappTemplateHref.Path += "/vAppTemplate/vappTemplate-" + strings.ReplaceAll(vAppTemplateId, "urn:vcloud:vapptemplate:", "")
```
Didainius, line 907: You could use `extractUuid` on `vAppTemplateId` here and in `Delete` function. 

Because of this mess with UUIDs/URNs in VCD, we have a bunch of functions here in SDK to deal with that. Some are `getBareEntityUuid`, `extractUuid`, `equalIds`, `IsUuid`, `isUrn`

Didainius (approved):

govcd/vapptemplate.go
adambarreiro: Changed, thanks for spotting it!

govcd/catalog.go
adambarreiro: I finally used `getBareEntityUuid` as in this case I needed the inner UUID, not the full URN.
Thanks for the suggestion, I've noted them for subsequent changes 🙂

govcd/vapptemplate_test.go
```diff
@@ -49,36 +49,51 @@ func (vcd *TestVCD) Test_RefreshVAppTemplate(check *C) {
...
 
-	// Get VAppTemplate
-	vAppTemplate, err := catItem.GetVAppTemplate()
+	vAppTemplate, err := catalog.GetVAppTemplateByName(itemName, true)
 	check.Assert(err, IsNil)
 	check.Assert(vAppTemplate, NotNil)
 	check.Assert(vAppTemplate.VAppTemplate.Name, Equals, itemName)
 
+	// FIXME: Due to bug in OVF Link upload in VCD, this assert is skipped
+	if isOvfLink {
```
adambarreiro, line 93: @Didainius @lvirbalas (mentioning you as you already reviewed and approved and this was recently added):

I have added a test for `UploadOvfByLink` as we didn't have one, and it's the one that can reproduce a bug currently present in VCD (hence this conditional `if isOvfLink` as well)

govcd/catalog.go
```diff
@@ -864,6 +864,21 @@ func (cat *Catalog) GetCatalogItemByName(catalogItemName string, refresh bool) (
 	return nil, ErrorEntityNotFound
 }
 
+// GetVAppTemplateByName finds a VAppTemplate by Name
+// On success, returns a pointer to the VAppTemplate structure and a nil error
+// On failure, returns a nil pointer and an error
+func (cat *Catalog) GetVAppTemplateByName(vAppTemplateName string, refresh bool) (*VAppTemplate, error) {
```
dataclouder, line 870: What happens if the catalog item name and the vApp template name are different?

govcd/vapptemplate.go
```diff
@@ -123,3 +123,16 @@ func (vAppTemplate *VAppTemplate) UpdateAsync() (Task, error) {
 	return vAppTemplate.client.ExecuteTaskRequest(url, http.MethodPut,
 		types.MimeVAppTemplate, "error updating vApp template item: %s", vappTemplatePayload)
 }
+
+// Delete deletes the VAppTemplate Item, returning an error if the VCD call fails.
```
dataclouder, line 127: Where does this method act? 
Does it delete the vApp template from a catalog?
How is this function different from CatalogItem.Delete?

dataclouder (commented):
A few items to discuss

govcd/catalog.go
dataclouder: Thanks for the test. This behavior and the impossibility of un-matching names should be documented.
As for deprecating catalog item, please hold on to it. I'd like to discuss the whole plan first.

govcd/vapptemplate.go
adambarreiro: The result and behaviour would be indeed equivalent to `CatalogItem.Delete`, the only difference is the implementation as the endpoint and receiver object is the vApp Template entity.

govcd/catalog.go
adambarreiro: I've added a few lines to `VAppTemplate.Update` to state this behaviour.

govcd/vapptemplate.go
```diff
@@ -65,12 +65,42 @@ func (vAppTemplate *VAppTemplate) Refresh() error {
...
+	}
+	if len(queriedVappTemplates) != 1 {
+		return "", fmt.Errorf("found more than one vApp Template with ID %s", vAppTemplate.VAppTemplate.ID)
+	}
+	return queriedVappTemplates[0].VdcName, nil
+}
+
+// Update updates the vApp template item information.
+// VCD also updates the associated Catalog Item, in order to be in sync with the receiver vApp Template entity.
+// For example, updating a vApp Template name "A" to "B" will make VCD to also update the Catalog Item to be renamed to "B".
```
lvirbalas, line 103: Does this mean if someone has both item and template in the same HCL, and changes template details, the next plan will show inconsistency for the item?

CC: @dataclouder @Didainius

dataclouder: having "both item and template in the same HCL" would be pointless. If someone does, there will be an inconsistency, but this is why we have a cleaner game, compared to the UI, where information can be hidden and convergence of entities presented as a unified view.
When we will eventually deprecate catalog_item, this problem will disappear. For now, users know that they need either of them. If they want to break the provider, there are other similar places that offer such opportunity.
IMO, this is a non-problem.

lvirbalas: I concur. It would actually be good to have a runtime sanity check that the user is not using both types for the same entity, but given Terraform limitations that would be difficult.

adambarreiro: Ideally it should be very hard to achieve as both `vcd_catalog_item` and `vcd_catalog_vapp_template` require a name and catalog, which must be a unique pair in VCD. So, if someone uses both in same HCL, one of them should give an error on creation saying that the vApp Template exists already.

To bypass that one could use the importing feature, I guess? In that case an inconsistency could be achieved.

govcd/vdc_test.go
```diff
@@ -561,3 +561,54 @@ func assertVDCAccessSettings(wanted, received []*types.AccessSetting) error {
 	}
 	return nil
 }
+
+// TestVAppTemplateRetrieval tests that VDC receiver objects can search vApp Templates successfully.
+func (vcd *TestVCD) TestVAppTemplateRetrieval(check *C) {
```
Didainius, line 566: could you possibly convert it to use nsxt vdcs and templates, just so we have one less thing to fix when removing nsxv and testing CDS

Didainius (approved):

govcd/vdc_test.go
adambarreiro: Done, I've picked your proposed changes from [here](https://github.com/vmware/go-vcloud-director/pull/509/files#diff-ccdc0aa4095bdd61b24c396c0572e6bcb8933ca63a89e9e1e5b9d7cdcff4e5bfR144), I also modified `sample_govcd_test_config.yaml` (let me know your thoughts about it, since probably you also need it in your PR, to avoid conflicts).

Didainius: Thanks!

PR #497, opened 2022-08-16
Go 1.19 release cleanups

Didainius:
Go 1.19 is released and this PR does a few cleanup adjustments. The main goal is to merge this PR ASAP to avoid messing up other PRs due to broken `make fmt` issues.

* `make fmt` using Go 1.19 release (`fmt` automatically changes doc comment structure). This will prevent `make static` errors when running tests using Go 1.19
* Update branding `vCloud Director` -> `VMware Cloud Director`
* Go officially supports 2 last releases. With Go 1.19 being released it means that Go 1.18 is the minimum officially supported Go version and this sets our hands free to use generic in this SDK (if there is a need for it). `go.mod` is updated to reflect Go minimum version 1.18
* package `io/ioutil` is deprecated as of Go 1.16. `staticcheck` started complaining about usage of deprecated packages. As a result this PR switches packages to either `io` or `os` (still the same functions are used)
* `staticcheck` switched version naming from `2021.1.2` to `v0.3.3` in download section. This PR also updates code to fetch correct one.

### DIFF
--- a/.changes/v2.17.0/497-notes.md
+++ b/.changes/v2.17.0/497-notes.md
@@ -1,8 +1,8 @@
-* `make fmt` using Go 1.19 release (`fmt` automatically changes doc comment structure). This will
-  prevent `make static` errors when running tests using Go 1.19 [GH-497]
-* Update branding `vCloud Director` -> `VMware Cloud Director` [GH-497]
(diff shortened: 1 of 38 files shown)


lvirbalas (approved):
Thanks!

PR #505, opened 2022-09-17
Simplify Test_LDAP

dataclouder:
* Remove VM and network creation from LDAP test
* Replace ephemeral VM with permanent LDAP server in the testing VCD
* Remove functions that were only used to create the VM for this test

This is similar to the change of `TestAccVcdOrgGroup` in [PR-909](https://github.com/vmware/terraform-provider-vcd/pull/909)

The reason for this change is that what we are testing within TestLDAP is quite fast, but the *preparation* takes a long time, so much that we added a tag `skipLong` to avoid such test in given circumstances. Moreover, creating a LDAP server (which requires routing to the VM) would fail for environmental reasons that add nothing to our testing experience.
The change requires a pre-set LDAP server, which is created only once per each VCD in a separate –automated– task, thus allowing us to test the intended feature without waiting for the server to be up.

### DIFF
--- a/govcd/api_vcd_test.go
+++ b/govcd/api_vcd_test.go
@@ -163,6 +163,7 @@ type TestConfig struct {
 		ExternalNetworkPortGroup     string `yaml:"externalNetworkPortGroup,omitempty"`
 		ExternalNetworkPortGroupType string `yaml:"externalNetworkPortGroupType,omitempty"`
 		VimServer                    string `yaml:"vimServer,omitempty"`
+		LdapServer                   string `yaml:"ldap_server,omitempty"`
 		Nsxt                         struct {
 			Manager             string `yaml:"manager"`
 			Tier0router         string `yaml:"tier0router"`
--- a/govcd/api_vcd_test.go
+++ b/govcd/api_vcd_test.go
@@ -204,9 +204,6 @@ type TestConfig struct {
 		PhotonOsOvaPath  string `yaml:"photonOsOvaPath,omitempty"`
 		MediaUdfTypePath string `yaml:"mediaUdfTypePath,omitempty"`
 	} `yaml:"media"`
-	Misc struct {
-		LdapContainer string `yaml:"ldapContainer,omitempty"`
-	} `yaml:"misc"`
 }
 
 // Test struct for vcloud-director.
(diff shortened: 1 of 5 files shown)


lvirbalas (approved):
LGTM! One ask only: for our own reference please add the explanation "why" we're doing this change in PR description. Thanks!

adambarreiro (approved):
I ran the tests with the `user` tag and all passed OK. They're much faster now that they don't spin-up the LDAP container. Thanks!

govcd/api_vcd_test.go
```diff
@@ -163,6 +163,7 @@ type TestConfig struct {
 		ExternalNetworkPortGroup     string `yaml:"externalNetworkPortGroup,omitempty"`
 		ExternalNetworkPortGroupType string `yaml:"externalNetworkPortGroupType,omitempty"`
 		VimServer                    string `yaml:"vimServer,omitempty"`
+		LdapServer                   string `yaml:"ldap_server,omitempty"`
```
Didainius, line 166: The ugly part of GitHub - I can't comment on unaltered code, but could you remove `Misc.LdapContainer` field from here as we don't really need it anymore.

govcd/sample_govcd_test_config.yaml
```diff
@@ -132,6 +132,9 @@ vcd:
     # A vSphere server name for creating an external network
     vimServer: vc9
     #
+    # IP of a pre-configured LDAP server
+    # using Docker image https://github.com/rroemhild/docker-test-openldap
+    ldap_server: 10.10.10.99
```
Didainius, line 137: Just like in the test struct, could you also remove the `misc.ldapContainer` documentation below

Didainius (approved):
Tests worked great. The only query I have is to remove `Misc.LdapContainer` variable from `TestConfig` struct and from `sample_govcd_test_config.yml`

govcd/api_vcd_test.go
dataclouder: Done

govcd/sample_govcd_test_config.yaml
dataclouder: done

PR #509, opened 2022-10-05
Adjust VM code

Didainius:
This code has a few tweaks for existing VM code.
* New function `vApp.AddRawVM` to support addition of VM from type `*types.ReComposeVAppParams` instead of functions such as `AddNewVMWithComputePolicy`, `AddNewVMWithStorageProfile`, `AddNewVM` which only accept some field values and need constant improvement.
* `vm.ChangeCPUAndCoreCount` which is almost the same as `vm.ChangeCPU` except that it accepts pointers instead of integers - this gives the option to specify only one of values the other being nil, which is more convenient in some workflows (including terraform-provider-vcd)
* Type `types.RecomposeVAppParamsForEmptyVm` had missing `PowerOn` field which is used to control VM power state on creation

Additional. Apparently we had a long existing inverse action for testing flag `GOVCD_IGNORE_CLEANUP_FILE`. It would perform cleanup for previous files only if `GOVCD_IGNORE_CLEANUP_FILE` is set (which is the reverse of its meaning). This PR patches condition check to match name of the flag.


govcd/api_vcd_test.go
```diff
@@ -604,7 +606,7 @@ func (vcd *TestVCD) SetUpSuite(check *C) {
 	// Gets the persistent cleanup list from file, if exists.
 	cleanupList, err := readCleanupList()
 	if len(cleanupList) > 0 && err == nil {
-		if ignoreCleanupFile {
+		if !ignoreCleanupFile {
```
Didainius, line 609: Fixes meaning for test flag `GOVCD_IGNORE_CLEANUP_FILE`

lvirbalas (approved):
LGTM!

I'm just thinking maybe we should add a note about `GOVCD_IGNORE_CLEANUP_FILE` fix/change to TESTING section of the changelog.

govcd/api_vcd_test.go
```diff
@@ -141,8 +141,10 @@ type TestConfig struct {
 		} `yaml:"nsxt_provider_vdc"`
 		Catalog struct {
 			Name                    string `yaml:"name,omitempty"`
+			NsxtBackedCatalogName   string `yaml:"nsxtBackedCatalogName,omitempty"`
```
adambarreiro, line 144: Should we also change `sample_govcd_test_config.yaml` to explain these new fields?

Didainius: Yes - checked it out from PR #495

govcd/vm_test.go
```diff
@@ -1964,3 +2014,103 @@ func (vcd *TestVCD) Test_VMChangeMemory(check *C) {
...
+	check.Assert(vapptemplate.VAppTemplate.Children.VM[0].HREF, NotNil)
+
+	vapp, err := vcd.nsxtVdc.CreateRawVApp(check.TestName(), check.TestName())
+	check.Assert(err, IsNil)
+	check.Assert(vapp, NotNil)
+	// After a successful creation, the entity is added to the cleanup list.
+	AddToCleanupList(vapp.VApp.Name, "vapp", vcd.nsxtVdc.Vdc.Name, check.TestName())
+
+	// fmt.Println("sleeping")
+	// time.Sleep(1 * time.Minute)
```
adambarreiro, line 2038: Just checking if these are leftovers.

govcd/vm_test.go
```diff
@@ -1936,6 +1936,56 @@ func (vcd *TestVCD) Test_VMChangeCPU(check *C) {
 	check.Assert(err, IsNil)
 }
 
+func (vcd *TestVCD) Test_VMChangeCPUAndCoreCount(check *C) {
+	if vcd.skipVappTests {
+		check.Skip("Skipping test because vapp was not successfully created at setup")
```
adambarreiro, line 1941: nit
```suggestion
		check.Skip("Skipping test because vApp was not successfully created at setup")
```

govcd/vm_test.go
```diff
@@ -1964,3 +2014,103 @@ func (vcd *TestVCD) Test_VMChangeMemory(check *C) {
...
+
+	vapp, err := vcd.nsxtVdc.CreateRawVApp(check.TestName(), check.TestName())
+	check.Assert(err, IsNil)
+	check.Assert(vapp, NotNil)
+	// After a successful creation, the entity is added to the cleanup list.
+	AddToCleanupList(vapp.VApp.Name, "vapp", vcd.nsxtVdc.Vdc.Name, check.TestName())
+
+	// fmt.Println("sleeping")
+	// time.Sleep(1 * time.Minute)
+	// // Check that vApp is powered of
```
adambarreiro, line 2039:
```suggestion
	// // Check that vApp is powered off
```

govcd/vm_test.go
```diff
@@ -1964,3 +2014,103 @@ func (vcd *TestVCD) Test_VMChangeMemory(check *C) {
...
+	check.Assert(vm.VM.Name, Equals, vmDef.SourcedItem.Source.Name)
+
+	// Refresh vApp to have latest state
+	err = vapp.Refresh()
+	check.Assert(err, IsNil)
+
+	// Check that vApp did not lose its state
+	vappStatus, err = vapp.GetStatus()
+	check.Assert(err, IsNil)
+	check.Assert(vappStatus, Equals, "MIXED") //vApp is powered on, but the VM withing is powered off
```
adambarreiro, line 2093:
```suggestion
	check.Assert(vappStatus, Equals, "MIXED") // vApp is powered on, but the VM within is powered off
```

adambarreiro (approved):
Just a few nitpicks

govcd/vm_test.go
Didainius: Yes! Thanks!

govcd/vm_test.go

PR #510, opened 2022-10-05
Add access level and domain to Metadata

adambarreiro:
## Overview

This PR makes Metadata handling 100% complete. Before, we lacked from the access levels and the domain, and also the type (string, number, boolean, date) was malfunctioning.

This creates the foundations for https://github.com/vmware/terraform-provider-vcd/pull/917.

## Detailed description

First thing to notice in this PR is a modification of the Metadata types in `types.go`:

* Some changes are cosmetic, like renaming `TypedValue` to `MetadataTypedValue`. This was done to accurately reflect the exact name of the type as shown in the VCD API docs (also, `TypedValue` was misleading: _Typed value of what?_)

* `types.MetadataTypedValue` had the `XsiType` attribute not properly mapped to the XML. It was `xsi:type,attr` but this is not the way to declare XML namespaces in Go as far as I understood. I've fixed that using the namespace URI `http://www.w3.org/2001/XMLSchema-instance` as explained [here](https://stackoverflow.com/questions/45770413/attribute-with-xml-namespace). This error during marshaling/unmarshaling caused the metadata type to be never retrieved during a GET. Somehow it worked on PUT/POST, maybe as StringValue is the default. Now it works and we can retrieve and use all available types.

* Added `MetadataDomainTag` to `types.MetadataValue` (the structure used to read) and changed `types.MetadataEntry` to map the Domain to `MetadataDomainTag` as well, instead of a string. The reason is that MetadataDomainTag contains both the visibility (READONLY, READWRITE, PRIVATE) and the domain name (SYSTEM/GENERAL).

In constants I added the three visibility types: READONLY, READWRITE, PRIVATE so SDK users can use them instead of hardcoding, pretty much like we had with the value types.

Last but not least, I created `metadata_v2.go` and `metadata_v2_test.go`. The idea behind this is not to pollute current `metadata.go` as it gets more and more complicated to read as we add things there. Once we release a new major of this go-vcloud-director, we should be able to get rid of `metadata.go` and `metadata_test.go` and everything should work out of the box (or at least it would be easier).

In `metadata_v2` I refactored things to make the code cleaner, specially the non-async functions which had a lot of duplicated code. I also reordered the functions. Instead of grouping by objects (vm, vapp, vdc...), I grouped by functionality (read, then add, then merge, then delete) which IMHO makes it easier to extend and read.


govcd/filter_helpers.go
```diff
@@ -496,7 +496,7 @@ func strToRegex(s string) string {
 // If the value looks like a number, or a true/false value, the corresponding type is returned
 // Otherwise, we assume it's a string.
 // We do this because the API doesn't return the metadata type
```
adambarreiro, line 498: I just read this. Could it be due to the wrong mapping of XsiType that I changed in this PR?

govcd/metadata_v2.go
```diff
@@ -0,0 +1,757 @@
...
+// The function supports passing a value that requires a typed value that must be one of:
+// types.MetadataStringValue, types.MetadataNumberValue, types.MetadataDateTimeValue and types.MetadataBooleanValue.
+// Visibility also needs to be one of: types.MetadataReadOnlyVisibility, types.MetadataHiddenVisibility or types.MetadataReadWriteVisibility
+func addMetadataAndWait(receiver metadataAsync, key, value, typedValue, visibility string, isSystem bool) error {
+	task, err := receiver.AddMetadataEntryWithVisibilityAsync(key, value, typedValue, visibility, isSystem)
+	if err != nil {
+		return err
+	}
+
+	return task.WaitTaskCompletion()
```
adambarreiro, line 699: Note to reviewers: Original `metadata.go` had a `Refresh()` call on the receiver objects. I've removed them as it didn't make much sense (metadata is handled outside the entities).
Let me know if I'm taking something for granted.

govcd/metadata_v2_test.go
```diff
@@ -0,0 +1,490 @@
...
+
+func (vcd *TestVCD) TestVAppMetadata(check *C) {
+	fmt.Printf("Running: %s\n", check.TestName())
+	if vcd.skipVappTests {
+		check.Skip("Skipping test because vApp was not successfully created at setup")
+	}
+	testMetadataCRUDActions(vcd.vapp, check, nil)
+}
+
+// TODO: Change to new vApp Template functions
```
adambarreiro, line 93: Requires #495

adambarreiro: Done.

govcd/metadata.go
```diff
@@ -171,22 +166,50 @@ func (vdc *Vdc) AddMetadataEntry(typedValue, key, value string) error {
 	return nil
 }
 
+// AddMetadataEntry adds VDC metadata typedValue and key/value pair provided as input
+// and waits for the task to finish.
+// Deprecated: Use AdminVdc.AddMetadataEntryWithVisibility instead
+func (adminVdc *AdminVdc) AddMetadataEntry(typedValue, key, value string) error {
```
adambarreiro, line 172: I added this function, although is already deprecated, just to be able to use interfaces in the consumer side (Terraform Provider in this case), which generalizes the usage of metadata and avoids code duplication.

adambarreiro:
I don't have any idea why `make static` is failing in the GitHub Action. In my local it's passing with exit code 0. In fact, the file it complains, **metadata_v2.go**, I've formatted it several times with different tools (go fmt and GoLand) and there's no difference to what it's present in this PR.

types/v56/types.go
```diff
@@ -1445,43 +1445,64 @@ type Value struct {
...
 }
 
-type TypedValue struct {
-	XsiType string `xml:"xsi:type,attr"`
+// MetadataTypedValue is the content of a metadata entry.
+// Type: MetadataTypedValue
+// Namespace: http://www.vmware.com/vcloud/v1.5
+// Description: One of: MetadataStringValue, MetadataNumberValue, MetadataBooleanValue, MetadataDateTimeValue
+// Since: 5.1
+type MetadataTypedValue struct {
```
Didainius, line 1464: Just to avoid breaking things in non major version you could add a type alias and note it is deprecated in favor of
MetadataTypedValue

`type TypedValue = MetadataTypedValue`

Didainius (commented):
I think this looks good (and clean!) in general, but I just want to get it another view once I check the #917

types/v56/types.go
adambarreiro: You're very right, this could break things. Changed!

govcd/metadata_v2.go
```diff
@@ -0,0 +1,754 @@
...
+
+// addMetadata adds metadata to an entity.
+// If the metadata entry is of the SYSTEM domain (isSystem=true), one can set different types of Visibility:
+// types.MetadataReadOnlyVisibility, types.MetadataHiddenVisibility but NOT types.MetadataReadWriteVisibility.
+// If the metadata entry is of the GENERAL domain (isSystem=false), visibility is always types.MetadataReadWriteVisibility.
+// In terms of typedValues, that must be one of:
+// types.MetadataStringValue, types.MetadataNumberValue, types.MetadataDateTimeValue and types.MetadataBooleanValue.
+func addMetadata(client *Client, requestUri, key, value, typedValue, visibility string, isSystem bool) (Task, error) {
+
+	if isSystem && visibility == types.MetadataReadWriteVisibility {
```
adambarreiro, line 650: While self-reviewing, I was thinking that having this conditional may be adding too much logic/awareness to this function. If someone introduces a wrong combination of `visibility` and `isSystem`, probably we should let VCD return an error, just in case something changes in future in VCD side.

Let me know your thoughts.

Didainius: I am all for VCD validation if the error coming from VCD is human readable and hints what is wrong. This is usually the approach we take to avoid over-validation.

lvirbalas: Yes, strategically, we always let CD API validate and add our own validation only where it's absolutely needed.

govcd/metadata_v2.go
```diff
@@ -0,0 +1,754 @@
...
+
+package govcd
+
+import (
+	"fmt"
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+	"net/http"
+)
+
+// NOTE: This "v2" is not v2 in terms of API versioning, it's just a new way of handling metadata.
```
lvirbalas, line 13: This is my main question about this PR: do we really want "v2"? For the readers of the code, the "v2" will lead to wonder "Why v2? How it's different from v1? Etc.".

If yes, we should extend this comment with a sentence of how v2 is better than v1 :)

CC: @dataclouder @Didainius

govcd/metadata_v2.go
dataclouder: If the error is an internal system error (with usually too much information and the real cause buried deep down) we should avoid exposing users to it

adambarreiro: I re-visited this use-case and the error coming from VCD is like this:
`API Error: 500: [ 83cca9ab-826f-406b-a641-de907cf8a3fe ] visibility`
What do you think? Probably it makes sense in this case, doesn't it?

Didainius: I actually was playing a bit with it in Terraform. What I didn't like is that I have like 10 metadata items and all I get is that error. A bit ugly that I can't know which item throws this error. I can still live with it, but a bit more information could be nice. 
image

adambarreiro: Ah, didn't think about it, but my error message would be useless even in that scenario. I will change it.

adambarreiro: I refactored the code, instead of "being too smart", I just beautified the error message for this unique case.

govcd/metadata_v2.go
Didainius: I like the fact of code being split. I just don't know a better name. IMO a good enough comment should be sufficient. That it supports metadata types, visibility, etc.

Didainius (approved):
You're just missing `make fmt`.

adambarreiro:
> You're just missing `make fmt`.

I need to check that, as locally my `make fmt` does nothing and `make static` passes. Not sure why the quirk here yet (will check).

govcd/metadata_v2.go
adambarreiro: I will revisit the comment and explain a bit better. I also thought about naming it `metadata_new.go` but I'm against using `new` suffix in most of the cases as its meaning is not very objective.

adambarreiro: Changed comment!

govcd/metadata_v2.go
lvirbalas: > 	// Workaround for ugly error returned by VCD: "API Error: 500: [ <uuid> ] visibility"
>	if err != nil && strings.HasSuffix(err.Error(), "visibility") {
> 		err = fmt.Errorf("error adding metadata with key %s: visibility cannot be %s when domain is %s", key, visibility, domain)

May there be another CD error with "visibility" in it which we don't expect? Just thinking to suggest adding the original err message at the very end of this custom message for completeness.

adambarreiro: Updated 👍

govcd/metadata_v2.go
lvirbalas: It's great, thanks!

lvirbalas (approved):

PR #513, opened 2022-10-25
Adjust VM Placement Policy code

adambarreiro:
### Overview

This PR adds a new method to update Compute Policies in **VMs** with the v2.0.0 version of the API, that also adds support for VM Placement Policies.

It sets the foundations for https://github.com/vmware/terraform-provider-vcd/pull/922, as in this PR we onboard VM Placement Policies for VMs.

### DIFF
--- a/govcd/vm.go
+++ b/govcd/vm.go
@@ -1546,7 +1546,7 @@ func (vm *VM) UpdateComputePolicyV2Async(sizingPolicyId, placementPolicyId strin
 
 	computePolicy := &types.ComputePolicy{}
 
-	if sizingPolicyId != "" {
+	if strings.TrimSpace(sizingPolicyId) != "" {
 		vdcSizingPolicyHref, err := vm.client.OpenApiBuildEndpoint(types.OpenApiPathVersion2_0_0, types.OpenApiEndpointVdcComputePolicies, sizingPolicyId)
 		if err != nil {
 			return Task{}, fmt.Errorf("error constructing HREF for sizing policy")
@@ -1554,7 +1554,7 @@ func (vm *VM) UpdateComputePolicyV2Async(sizingPolicyId, placementPolicyId strin
 		computePolicy.VmSizingPolicy = &types.Reference{HREF: vdcSizingPolicyHref.String()}
 	}
 
-	if placementPolicyId != "" {
+	if strings.TrimSpace(placementPolicyId) != "" {
 		vdcPlacementPolicyHref, err := vm.client.OpenApiBuildEndpoint(types.OpenApiPathVersion2_0_0, types.OpenApiEndpointVdcComputePolicies, placementPolicyId)
 		if err != nil {
 			return Task{}, fmt.Errorf("error constructing HREF for placement policy")
--- a/govcd/vm.go
+++ b/govcd/vm.go
@@ -1487,7 +1487,29 @@ func (vm *VM) UpdateVmSpecSectionAsync(vmSettingsToUpdate *types.VmSpecSection, 
 		})
 }
 
+// UpdateComputePolicyV2 updates VM compute policy and returns refreshed VM or error.
+func (vm *VM) UpdateComputePolicyV2(sizingPolicy, placementPolicy *types.VdcComputePolicyV2) (*VM, error) {
+	task, err := vm.UpdateComputePolicyV2Async(sizingPolicy, placementPolicy)
+	if err != nil {
+		return nil, err
+	}
+
+	err = task.WaitTaskCompletion()
+	if err != nil {
+		return nil, err
+	}
+
+	err = vm.Refresh()
+	if err != nil {
+		return nil, err
+	}
+
+	return vm, nil
+
+}
+
 // UpdateComputePolicy updates VM compute policy and returns refreshed VM or error.
+// Deprecated: Use VM.UpdateComputePolicyV2 instead
 func (vm *VM) UpdateComputePolicy(computePolicy *types.VdcComputePolicy) (*VM, error) {
 	task, err := vm.UpdateComputePolicyAsync(computePolicy)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


govcd/vm.go
```diff
@@ -1487,7 +1487,29 @@ func (vm *VM) UpdateVmSpecSectionAsync(vmSettingsToUpdate *types.VmSpecSection,
 		})
 }
 
+// UpdateComputePolicyV2 updates VM compute policy and returns refreshed VM or error.
```
Didainius, line 1490: Can you just add the important bit you put in `UpdateComputePolicyV2Async`. The note that it is using v2.0.0 OpenAPI endpoint.

govcd/vm.go
```diff
@@ -1508,7 +1530,58 @@ func (vm *VM) UpdateComputePolicy(computePolicy *types.VdcComputePolicy) (*VM, e
 
 }
 
+// UpdateComputePolicyV2Async updates VM Compute policy with the given compute policies using v2.0.0 OpenAPI endpoint,
+// and returns a Task and an error. Updating with an empty compute policy ID will remove it from the VM. Both
+// policies can't be empty as the VM requires at least one policy.
+func (vm *VM) UpdateComputePolicyV2Async(sizingPolicyId, placementPolicyId string) (Task, error) {
```
Didainius, line 1536: I've got one concern about function parameters - what will we do when we get to vGpu policy. I think leaving the function as it is will force us to deprecate it. Should we better accept the container of all policies `*types.ComputePolicy`?

govcd/vm.go
```diff
@@ -1508,7 +1530,58 @@ func (vm *VM) UpdateComputePolicy(computePolicy *types.VdcComputePolicy) (*VM, e
 
 }
 
+// UpdateComputePolicyV2Async updates VM Compute policy with the given compute policies using v2.0.0 OpenAPI endpoint,
+// and returns a Task and an error. Updating with an empty compute policy ID will remove it from the VM. Both
```
Didainius, line 1534: One thing that popped my head and needs clarification - do you know if sending a `types.ComputePolicy{}` structure without placement policy structure (because `types.ComputePolicy.VmPlacementPolicy` field has `omitempty`) still causes removal of placement policy (if it was previously assigned)? 

To work around it, user could at first retrieve ComputePolicy for VM and update only the relevant fields, but we don't yet have vGpu policy fields in `types.ComputePolicy` so it would be impossible to preserve them while doing update. We could possibly add them (without any testing) just to avoid this problem (if my initial query is valid)

vGpu policy fields in `types.ComputePolicy` and that would always wipe it for someone who has manually set it

Didainius (commented):
Pass one - it is all clear, except I have this one query about possibly problems with vGpu policies (you will find it in comments)

govcd/vm.go
adambarreiro: I tested it and sending an empty `&types.ComputePolicy{}` does nothing. I was expecting an error from VCD, because whenever you try to remove all policies from a VM, UI shows an error saying "you need at least one policy". But it simply ignores the call (it performs an update in the VM but the policies remain the same).

Actually this is the reason I added this snippet:

```
	if sizingIsEmpty && placementIsEmpty {
		return Task{}, fmt.Errorf("either sizing policy ID or placement policy ID is needed")
	}
```

As I needed some sort of error to avoid removing all policies in the Terraform provider, otherwise we would get state<->real state mismatches all the time.

govcd/vm.go
adambarreiro: That is a good point! The only thing I dislike is that it will be a bit more verbose in the consumer side (Terraform provider in this case) but it will be 100% complete indeed, which is nice as it will avoid deprecations as you say.

I will change it :+1:

PS: Another option would be having a variadic function, like

```
UpdateComputePolicyV2Async(policyIds... string)
```

but we lose which policy type each one is, unless we comment it.

govcd/vm.go
adambarreiro: Done, thanks!

lvirbalas (approved):

govcd/vm.go
adambarreiro: I added a `vGpuPolicyId` parameter, although it is ignored. I couldn't find it is used in the [documentation](https://developer.vmware.com/apis/1260/vmware-cloud-director/doc/doc//types/VmType.html).

govcd/vm.go
```diff
@@ -1508,7 +1533,64 @@ func (vm *VM) UpdateComputePolicy(computePolicy *types.VdcComputePolicy) (*VM, e
...
+	if vm.VM.HREF == "" {
+		return Task{}, fmt.Errorf("cannot update VM compute policy, VM HREF is unset")
+	}
+
+	sizingIsEmpty := strings.TrimSpace(sizingPolicyId) == ""
+	placementIsEmpty := strings.TrimSpace(placementPolicyId) == ""
+	vGpuPolicyIsEmpty := strings.TrimSpace(vGpuPolicyId) == ""
+
+	if !vGpuPolicyIsEmpty {
+		util.Logger.Printf("[WARN] vGPU policies are not supported, hence %s will be ignored", vGpuPolicyId)
```
Didainius, line 1550: I wonder if we should just from an error here  to clearly state the situation. I doubt a lot of people would read logs.

Didainius (approved):
Thanks. Test passed on 10.3.3, 10.4..0

govcd/vm.go
adambarreiro: Changed!

govcd/vm_test.go
```diff
@@ -2021,6 +2174,37 @@ func (vcd *TestVCD) Test_VMChangeMemory(check *C) {
 }
 
 func (vcd *TestVCD) Test_AddRawVm(check *C) {
+	vapp, vm := createNsxtVAppAndVm(vcd, check)
+
+	// Check that vApp did not lose its state
```
dataclouder, line 2179: Before using vApp and VM, we should check that they are not nil

dataclouder (approved):
Only one minor issue. LGTM otherwise

govcd/vm_test.go
adambarreiro: Good catch! Fixed

PR #515, opened 2022-11-01
Add org users to govcd test configuration

dataclouder:
A new structure `[]tenant` is added to `TestConfig`, to allow connecting as tenant user.
When the option `-vcd-test-org-user` (same as what is used in terraform-provider-vcd) is se, the first tenant in the list is selected for connection.
To use the second tenant (or an optional third one), we can use the option `-vcd-connect-tenant=1` (or `-vcd-connect-tenant=2` if a third one exists).

### DIFF
--- a/.changes/v2.17.0/515-improvements.md
+++ b/.changes/v2.17.0/515-improvements.md
@@ -1,0 +1,1 @@
+* Add `[]tenant` structure to simplify org user testing [GH-515]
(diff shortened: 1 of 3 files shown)


adambarreiro (approved):

Didainius (approved):
Just a reminder - we also have "on the fly" option for getting a client with Org user connection - `func newOrgUserConnection`

dataclouder:
> Just a reminder - we also have "on the fly" option for getting a client with Org user connection - `func newOrgUserConnection`

Yes. I remember. This change, however, allows existing tests to run as Org user. We will need to add appropriate skip statements if a test is supposed to be used blindly by both admin and org users.
So far, we can target individual tests to run as org user, which is useful during development

lvirbalas (approved):
Looks great!

PR #516, opened 2022-11-02
Add GitHub Action to run gosec and amend already present issues

adambarreiro:
### Description

This PR is not related to any opened issue as far as I know. It adds the [`gosec`](https://github.com/securego/gosec) tool as a GitHub Action that runs on every **push** and **pull_request** event.

As we had already some issues detected in the code, adding this action would break all subsequent PR builds, hence no one could merge until solved. For that reason, this PR also solves/amends the found issues.

Thanks @dataclouder for the `gosec` idea!

- [x] Tests pass

### DIFF
--- a/samples/discover/discover.go
+++ b/samples/discover/discover.go
@@ -94,7 +94,7 @@ func check_configuration(conf Config) {
 // Retrieves the configuration from a Json or Yaml file
 func getConfig(config_file string) Config {
 	var configuration Config
-	//#nosec G304 -- We only accept YAML files, it would fail once unmarshalled if one puts a malicious file
+	// #nosec G304 -- We only accept YAML files, it would fail once unmarshalled if one puts a malicious file
 	buffer, err := os.ReadFile(config_file)
 	if err != nil {
 		fmt.Printf("Configuration file %s not found\n%s\n", config_file, err)
--- a/samples/discover/discover.go
+++ b/samples/discover/discover.go
@@ -94,6 +94,7 @@ func check_configuration(conf Config) {
 // Retrieves the configuration from a Json or Yaml file
 func getConfig(config_file string) Config {
 	var configuration Config
+	//#nosec G304 -- We only accept YAML files, it would fail once unmarshalled if one puts a malicious file
 	buffer, err := os.ReadFile(config_file)
 	if err != nil {
 		fmt.Printf("Configuration file %s not found\n%s\n", config_file, err)
--- a/samples/discover/discover.go
+++ b/samples/discover/discover.go
@@ -45,6 +45,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"path/filepath"
 
 	"gopkg.in/yaml.v2"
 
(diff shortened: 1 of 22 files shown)


util/tar.go
```diff
@@ -86,14 +91,21 @@ func Unpack(tarFile string) ([]string, string, error) {
 
 			// if it's a newFile create it
 		case tar.TypeReg:
+			// #nosec G304 -- File path is coming from inside the TAR file.
 			newFile, err := os.OpenFile(target, os.O_CREATE|os.O_RDWR, os.FileMode(header.Mode))
```
dataclouder, line 95: Actually, this could be a potential vulnerability, which we should investigate. If the tar file contains something like `'../../../../../../../../../../etc'`, which points outside the extraction directory.

See this [similar case](https://github.com/advisories/GHSA-47wr-426j-fr82)

util/tar.go
```diff
@@ -138,11 +150,16 @@ func sanitizedName(filename string) string {
 
 // GetFileContentType returns the real file type
 func GetFileContentType(file string) (string, error) { // Open File
+	// #nosec G304 -- The file is just opened to detect the content type
 	f, err := os.Open(file)
```
dataclouder, line 154: it is possible that using `filepath.Clean` would be enough to satisfy the warning

util/logging.go
```diff
@@ -112,6 +113,7 @@ func TogglePanicEmptyUserAgent(willPanic bool) {
 func newLogger(logpath string) *log.Logger {
 	var err error
 	var file *os.File
+	// #nosec G302 G304 -- The file has read permissions for groups as it is just a log
 	if OverwriteLog {
 		file, err = os.OpenFile(logpath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0640)
```
dataclouder, line 118: Possibly adding `filepath.Clean` here?
Also, what problem would we have if we restrict permissions to `0600`?

samples/discover/discover.go
```diff
@@ -94,6 +94,7 @@ func check_configuration(conf Config) {
 // Retrieves the configuration from a Json or Yaml file
 func getConfig(config_file string) Config {
 	var configuration Config
+	// #nosec G304 -- We only accept YAML files, it would fail once unmarshalled if one puts a malicious file
 	buffer, err := os.ReadFile(config_file)
```
dataclouder, line 98: `filepath.Clean` might be effective here

dataclouder (commented):
There are a few possible improvements. Let's discuss

util/tar.go
```diff
@@ -16,18 +16,23 @@ import (
...
-// Extract files to system tmp dir with name govcd+random number. Created folder with files isn't deleted.
+// Unpack extracts files to system tmp dir with name govcd+random number. Created folder with files isn't deleted.
 // Returns extracted files paths in array and path where folder with files created.
 func Unpack(tarFile string) ([]string, string, error) {
 
 	var filePaths []string
 	var dst string
 
+	// #nosec G304 -- It only accepts TAR files, otherwise it will fail later
 	reader, err := os.Open(tarFile)
```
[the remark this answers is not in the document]
adambarreiro: Fixed, thanks!

util/tar.go
adambarreiro: Fixed, thanks!

Didainius (approved):
Thanks! I am good provided this runs well

dataclouder (approved):

lvirbalas (approved):
Thanks!

Issue #514, opened 2022-11-01
GetNetworkList() truncates the network list when a VDC has more than 25 networks


RussPitcher:
### Describe the bug

When running a Terraform action against an OrgVDC with more than 25 OrgVDC networks the state refresh fails with one or more errors as follows:

`Error: error retrieving network myOrgVdcNetwork from network list`

On investigation we discovered that the query used to retrieve the network list does not manage paged responses and therefore uses the default page size of 25 records.  Any records apart from the initial 25 are not returned.  This causes the calling functions to fail when the expected network is missing.

We tracked the REST call down to the `GetNetworkList()` function in the `orgvdcnetwork.go` file and it can be seen that the pageSize parameter is not used, and the function itself does not attempt to handle paged responses.

### Reproduction steps

1. Create an OrgVDC with more than 25 terraform-managed `vdc_network_direct` objects
2. Run a `terraform plan`

### Expected behavior

The terraform plan should refresh the state successfully without errors

### Additional context

Adding a simple `"pageSize": "128"` to the query would reduce the scope of this issue (as has been done in the `QueryProviderVdcStorageProfileByName` function) but it would be better to implement proper support for paged responses.

dataclouder:
Thanks for the investigation. we were aware of the problem, but this function skipped through the cracks last time we went around fixing it.
There is a simple fix (we do have a `cumulativeQuery` that does the pagination automatically). It will require some tests, but it should be addressed easily.

dataclouder:
Fixed with PR #518

PR #518, opened 2022-11-14
Fix network list 514

dataclouder:
Address Issue #514, by making sure that the list of networks is not truncated to the default page size
It also fixes terraform-provider-vcd [Issue #919](https://github.com/vmware/terraform-provider-vcd/issues/919)


Didainius (approved):
Test pased on 3 versions and the PR makes sense - thanks!

adambarreiro (approved):
LGTM!

Issue #526, opened 2022-11-28
Missing vm OVF_Environment


odraghi:
### Is your feature request related to a problem? Please describe.

-

### Describe the solution you'd like

Add VM types for OVF_Environment and it's children . Don't forget the attribut vCenterId witch is the VM moref.
Add func VM.GetEnvironment

Notice :
 I don't find specific API call to retrieve only "Environment" section. I think that  it's needed to fetch with the vm href for that.

### Additional context

Sample extracted from a VM xml:

 <ovfenv:Environment xmlns:ve="http://www.vmware.com/schema/ovfenv" ovfenv:id="" ve:vCenterId="vm-583592">
        <ovfenv:PlatformSection>
            <ovfenv:Kind>VMware ESXi</ovfenv:Kind>
            <ovfenv:Version>6.7.0</ovfenv:Version>
            <ovfenv:Vendor>VMware, Inc.</ovfenv:Vendor>
            <ovfenv:Locale>en</ovfenv:Locale>
        </ovfenv:PlatformSection>
        <ovfenv:PropertySection>
            <ovfenv:Property ovfenv:key="vCloud_UseSysPrep" ovfenv:value="None"/>
            <ovfenv:Property ovfenv:key="vCloud_bitMask" ovfenv:value="1"/>
            <ovfenv:Property ovfenv:key="vCloud_bootproto_0" ovfenv:value="static"/>
            <ovfenv:Property ovfenv:key="vCloud_computerName" ovfenv:value="testubu"/>
            <ovfenv:Property ovfenv:key="vCloud_dns1_0" ovfenv:value="8.8.8.8"/>
            <ovfenv:Property ovfenv:key="vCloud_dns2_0" ovfenv:value=""/>
            <ovfenv:Property ovfenv:key="vCloud_gateway_0" ovfenv:value="192.168.0.1"/>
            <ovfenv:Property ovfenv:key="vCloud_ip_0" ovfenv:value="192.168.0.103"/>
            <ovfenv:Property ovfenv:key="vCloud_macaddr_0" ovfenv:value="00:50:56:01:1e:cb"/>
            <ovfenv:Property ovfenv:key="vCloud_markerid" ovfenv:value="532524b8-5141-4046-bb22-212eb307b20a"/>
            <ovfenv:Property ovfenv:key="vCloud_netmask_0" ovfenv:value="255.255.255.0"/>
[...]

PR #528, opened 2022-11-29
Add VM.GetEnvironment() to retrieve OVF Environment

odraghi:
Signed-off-by: Olivier DRAGHI <odraghi@gmail.com>

## Description
I need a way for a service running inside a VM to query cloud director as an OrgAdmin and find itself.
I think about checking the VM moref that is available with vmtools :
  vmtoolsd --cmd "info-get guestinfo.ovfenv" | grep "ve:vCenterId" | cut -d'"' -f2

Unfortunately getting ovfenv ( Environment ) was not implemeted in govdc.

Related issue: (#526)

### DIFF
--- a/govcd/vm.go
+++ b/govcd/vm.go
@@ -836,11 +836,11 @@ func (vm *VM) GetProductSectionList() (*types.ProductSectionList, error) {
 func (vm *VM) GetEnvironment() (*types.OVF_Environment, error) {
 	vmStatus, err := vm.GetStatus()
 	if err != nil {
-		return nil, fmt.Errorf("unable to get ovf environment: %s", err)
+		return nil, fmt.Errorf("unable to get OVF environment: %s", err)
 	}
 
 	if vmStatus != "POWERED_ON" {
-		return nil, fmt.Errorf("ovf environment is only available when VM is poweredOn")
+		return nil, fmt.Errorf("OVF environment is only available when VM is powered on")
 	}
 
 	return vm.VM.Environment, nil
--- a/govcd/vm.go
+++ b/govcd/vm.go
@@ -832,6 +832,21 @@ func (vm *VM) GetProductSectionList() (*types.ProductSectionList, error) {
 	return getProductSectionList(vm.client, vm.VM.HREF)
 }
 
+// GetEnvironment returns the OVF Environment. It's only available for poweredOn VM
+func (vm *VM) GetEnvironment() (*types.OVF_Environment, error) {
+	v := &types.Vm{}
+
+	if vm.VM.HREF == "" {
+		return nil, fmt.Errorf("cannot refresh, invalid reference url")
+	}
+
+	_, err := vm.client.ExecuteRequest(vm.VM.HREF, http.MethodGet,
+		types.MimeVM, "error retrieving ovf environment: %s", nil, v)
+
+	// The request was successful
+	return v.Environment, err
+}
+
 // GetGuestCustomizationSection retrieves guest customization section for a VM. It allows to read VM guest customization properties.
 func (vm *VM) GetGuestCustomizationSection() (*types.GuestCustomizationSection, error) {
 	if vm == nil || vm.VM.HREF == "" {
--- a/govcd/vm.go
+++ b/govcd/vm.go
@@ -834,17 +834,16 @@ func (vm *VM) GetProductSectionList() (*types.ProductSectionList, error) {
 
 // GetEnvironment returns the OVF Environment. It's only available for poweredOn VM
 func (vm *VM) GetEnvironment() (*types.OVF_Environment, error) {
-	v := &types.Vm{}
-
-	if vm.VM.HREF == "" {
-		return nil, fmt.Errorf("cannot refresh, invalid reference url")
+	vmStatus, err := vm.GetStatus()
+	if err != nil {
+		return nil, fmt.Errorf("unable to get ovf environment: %s", err)
 	}
 
-	_, err := vm.client.ExecuteRequest(vm.VM.HREF, http.MethodGet,
-		types.MimeVM, "error retrieving ovf environment: %s", nil, v)
+	if vmStatus != "POWERED_ON" {
+		return nil, fmt.Errorf("ovf environment is only available when VM is poweredOn")
+	}
 
-	// The request was successful
-	return v.Environment, err
+	return vm.VM.Environment, nil
 }
 
 // GetGuestCustomizationSection retrieves guest customization section for a VM. It allows to read VM guest customization properties.
--- a/govcd/vm.go
+++ b/govcd/vm.go
@@ -833,7 +833,7 @@ func (vm *VM) GetProductSectionList() (*types.ProductSectionList, error) {
 }
 
 // GetEnvironment returns the OVF Environment. It's only available for poweredOn VM
-func (vm *VM) GetEnvironment() (*types.OVF_Environment, error) {
+func (vm *VM) GetEnvironment() (*types.OvfEnvironment, error) {
 	vmStatus, err := vm.GetStatus()
 	if err != nil {
 		return nil, fmt.Errorf("unable to get OVF environment: %s", err)
--- a/govcd/vm_test.go
+++ b/govcd/vm_test.go
@@ -2281,3 +2281,55 @@ func createNsxtVAppAndVm(vcd *TestVCD, check *C) (*VApp, *VM) {
 
 	return vapp, vm
 }
+
+func (vcd *TestVCD) Test_GetOvfEnvironment(check *C) {
+	if vcd.skipVappTests {
+		check.Skip("Skipping test because vapp was not successfully created at setup")
+	}
+	vapp := vcd.findFirstVapp()
+	existingVm, vmName := vcd.findFirstVm(vapp)
+	if vmName == "" {
+		check.Skip("skipping test because no VM is found")
+	}
+	vm, err := vcd.client.Client.GetVMByHref(existingVm.HREF)
+	check.Assert(err, IsNil)
+
+	vmStatus, err := vm.GetStatus()
+	check.Assert(err, IsNil)
+	if vmStatus != "POWERED_ON" {
+		task, err := vm.PowerOn()
+		check.Assert(err, IsNil)
+		err = task.WaitTaskCompletion()
+		check.Assert(err, IsNil)
+		check.Assert(task.Task.Status, Equals, "success")
+	}
+
+	// Read ovfenv when VM is started
+	ovfenv, err := vm.GetEnvironment()
+	check.Assert(err, IsNil)
+	check.Assert(strings.Contains(ovfenv.VCenterId, "vm-"), Equals, true)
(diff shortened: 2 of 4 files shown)


dataclouder:
Thanks for this contribution.
Please, expand the description of this PR with the following:
1. What is the problem that you are trying to solve and the current code doesn't solve? In other words, define what is the purpose of this proposal.
2. Add at least one test that proves this change is doing what it is supposed to do. If you are unable to write such test (the gocheck environment requires some learning) please suggest what such test should do to prove that the change you propose is effective.

govcd/vm.go
```diff
@@ -832,6 +832,21 @@ func (vm *VM) GetProductSectionList() (*types.ProductSectionList, error) {
 	return getProductSectionList(vm.client, vm.VM.HREF)
 }
 
+// GetEnvironment returns the OVF Environment. It's only available for poweredOn VM
+func (vm *VM) GetEnvironment() (*types.OVF_Environment, error) {
```
dataclouder, line 836: This function is basically doing a Refresh. So, why not calling VM.Refresh in the first place?
Moreover, I saw from the test that the environment is not available when the VM is not powered ON.
The way I understand it, the function should check if the VM is powered ON, return an error is it is not, and get the environment data when it is. No need to run a refresh, unless I am missing something.

govcd/vm_test.go
```diff
@@ -2281,3 +2281,55 @@ func createNsxtVAppAndVm(vcd *TestVCD, check *C) (*VApp, *VM) {
...
+		check.Assert(err, IsNil)
+		err = task.WaitTaskCompletion()
+		check.Assert(err, IsNil)
+		check.Assert(task.Task.Status, Equals, "success")
+	}
+
+	// Read ovfenv when VM is started
+	ovfenv, err := vm.GetEnvironment()
+	check.Assert(err, IsNil)
+	check.Assert(strings.Contains(ovfenv.VCenterId, "vm-"), Equals, true)
```
dataclouder, line 2310: This Assert will panic if the environment is nil

govcd/vm_test.go
```diff
@@ -2281,3 +2281,55 @@ func createNsxtVAppAndVm(vcd *TestVCD, check *C) (*VApp, *VM) {
...
+	check.Assert(ovfenv, NotNil)
+	check.Assert(ovfenv.PlatformSection, NotNil)
+	check.Assert(ovfenv.PlatformSection.Kind, Equals, "VMware ESXi")
```
dataclouder, line 2313: You are assuming that the VM is from VMware. What happens if it was a plain Ubuntu or CentOS?

govcd/vm_test.go
```diff
@@ -2281,3 +2281,55 @@ func createNsxtVAppAndVm(vcd *TestVCD, check *C) (*VApp, *VM) {
...
+			check.Assert(p.Value, Not(Equals), "")
+		}
+	}
+	check.Assert(ovfenv.EthernetAdapterSection, NotNil)
+	for _, p := range ovfenv.EthernetAdapterSection.AdapterList {
+		check.Assert(p.Mac, Not(Equals), "")
+	}
+
+	// PowerOff
+	task, err := vm.PowerOff()
```
dataclouder, line 2326: If the VM was powered on already, this test will change its state, which may affect other tests. 
This test needs to leave things as they were

types/v56/vm_types.go
```diff
@@ -161,3 +160,46 @@ type SourcedVmTemplateParams struct {
 	VmTemplateInstantiationParams *InstantiationParams `xml:"VmTemplateInstantiationParams,omitempty"` // Same as InstantiationParams used for VMs within a vApp
 	StorageProfile                *Reference           `xml:"StorageProfile,omitempty"`                // A reference to a storage profile to be used for the VM. The specified storage profile must exist in the organization vDC that contains the composed vApp. If not specified, the default storage profile for the vDC is used.
 }
+
+type OVF_Environment struct {
```
dataclouder, line 164: I would like to see descriptions of what the fields mean.
Where did you get the data structure? Could you share a link to the API element you have used?

govcd/vm_test.go
odraghi: PlatformSection show hypervisor informations.

So don't worry about Guest VM opertaing system.

odraghi: https://www.dmtf.org/standards/ovf

https://www.dmtf.org/sites/default/files/standards/documents/DSP0243_2.1.1.pdf
page 40


The **PlatformSection** element contains optional information provided by the deployment function. The
Kind, Version, and Vendor elements **describe the virtualization platform**. The Locale and TimeZone
elements describe the current locale and time zone.

govcd/vm_test.go
```diff
@@ -2281,3 +2281,69 @@ func createNsxtVAppAndVm(vcd *TestVCD, check *C) (*VApp, *VM) {
...
+		task, err := vm.PowerOn()
+		check.Assert(err, IsNil)
+		err = task.WaitTaskCompletion()
+		check.Assert(err, IsNil)
+		check.Assert(task.Task.Status, Equals, "success")
+	}
+
+	// Read ovfenv when VM is started
+	ovfenv, err := vm.GetEnvironment()
+	check.Assert(err, IsNil)
```
adambarreiro, line 2309:
```suggestion
	check.Assert(err, IsNil)
	check.Assert(ovfenv, NotNil)
```

types/v56/vm_types.go
```diff
@@ -161,3 +161,54 @@ type SourcedVmTemplateParams struct {
 	VmTemplateInstantiationParams *InstantiationParams `xml:"VmTemplateInstantiationParams,omitempty"` // Same as InstantiationParams used for VMs within a vApp
 	StorageProfile                *Reference           `xml:"StorageProfile,omitempty"`                // A reference to a storage profile to be used for the VM. The specified storage profile must exist in the organization vDC that contains the composed vApp. If not specified, the default storage profile for the vDC is used.
 }
+
+// The OVF environment enables the guest software to access information about the virtualization platform, such as
+// the user-specified values for the properties defined in the OVF descriptor.
+type OVF_Environment struct {
+	XMLName                xml.Name                `xml:"Environment"`
+	Ve                     string                  `xml:"ve,attr,omitempty"`				// Xml namespace
```
adambarreiro, line 169: I think this one needs a `go fmt` run.

types/v56/vm_types.go
```diff
@@ -161,3 +161,54 @@ type SourcedVmTemplateParams struct {
...
+	Locale  string   `xml:"Locale,omitempty"`  // Hypervisor locale
+}
+
+// Contains a list of key/value pairs corresponding to properties defined in the OVF descriptor
+// Operating system level configuration, such as host names, IP address, subnets, gateways, etc.
+// Application-level configuration such as DNS name of active directory server, databases and
+// other external services.
+type PropertySection struct {
+	XMLName      xml.Name     `xml:"PropertySection"`
+	PropertyList PropertyList `xml:"Property,omitempty"`
```
adambarreiro, line 192:
```suggestion
	PropertyList []*OVF_Property `xml:"Property,omitempty"`
```

Imho I'd leave it as `[]*OVF_Property` instead of `PropertyList` as the name could be a bit misleading (it is a slice and it sounds like a `container.list`). My same opinion goes for other cases in this file.

Not sure what do others think? cc: @dataclouder  @Didainius @lvirbalas

adambarreiro (changes requested):
Thanks for the contribution!
The new test passed for me in 10.4.1 and 10.3.0. 
Just a couple of comments/suggestions/changes:

govcd/vm_test.go
```diff
@@ -2281,3 +2281,70 @@ func createNsxtVAppAndVm(vcd *TestVCD, check *C) (*VApp, *VM) {
...
+
+	// PowerOff
+	task, err := vm.PowerOff()
+	check.Assert(err, IsNil)
+	err = task.WaitTaskCompletion()
+	check.Assert(err, IsNil)
+	check.Assert(task.Task.Status, Equals, "success")
+
+	ovfenv, err = vm.GetEnvironment()
+	check.Assert(strings.Contains(err.Error(), "ovf environment is only available when VM is poweredOn"), Equals, true)
```
lvirbalas, line 2339:
```suggestion
	check.Assert(strings.Contains(err.Error(), "OVF environment is only available when VM is powered on"), Equals, true)
```

govcd/vm.go
```diff
@@ -832,6 +832,20 @@ func (vm *VM) GetProductSectionList() (*types.ProductSectionList, error) {
...
+	vmStatus, err := vm.GetStatus()
+	if err != nil {
+		return nil, fmt.Errorf("unable to get ovf environment: %s", err)
```
lvirbalas, line 839:
```suggestion
		return nil, fmt.Errorf("unable to get OVF environment: %s", err)
```

govcd/vm.go
```diff
@@ -832,6 +832,20 @@ func (vm *VM) GetProductSectionList() (*types.ProductSectionList, error) {
...
+	}
+
+	if vmStatus != "POWERED_ON" {
+		return nil, fmt.Errorf("ovf environment is only available when VM is poweredOn")
```
lvirbalas, line 843:
```suggestion
		return nil, fmt.Errorf("OVF environment is only available when VM is powered on")
```

lvirbalas (changes requested):
(Tidying error messages).

types/v56/vm_types.go
```diff
@@ -161,3 +161,50 @@ type SourcedVmTemplateParams struct {
...
+// Contains a list of key/value pairs corresponding to properties defined in the OVF descriptor
+// Operating system level configuration, such as host names, IP address, subnets, gateways, etc.
+// Application-level configuration such as DNS name of active directory server, databases and
+// other external services.
+type PropertySection struct {
+	XMLName    xml.Name        `xml:"PropertySection"`
+	Properties []*OVF_Property `xml:"Property,omitempty"`
+}
+
+type OVF_Property struct {
```
lvirbalas, line 195: Please use `OvfProperty` and `OvfEnvironment` (above) format or the type names.

adambarreiro (approved):
Thanks for the contribution!

Didainius:
@odraghi ,
Thanks for all fixes. I am testing the SDK itself as well as will run our Terraform provider for VCD tests because the OVF properties belong to VM struct and it can have side effects

odraghi:
@Didainius you're welcome, let me know if you need me to change something

Didainius (approved):

Didainius:
I have tried to run tests in such combination:
* go-vcloud-director was on 10.4.0 (whole suite passed) and 10.3.0 (some tests failed, but they don't look related)
* Terraform provider tests ran in 10.4.1 (based on this PR https://github.com/vmware/terraform-provider-vcd/pull/1016). A few tests failed, but not related to anything about OVF customization

lvirbalas (approved):
Thank you for contributing!

dataclouder (approved):

Didainius:
Thanks @odraghi , this is in `main` now

PR #530, opened 2022-12-01
Fix multiple resource pools issue when fetching VM Groups

adambarreiro:
## Context

In [v2.17.0](https://github.com/vmware/go-vcloud-director/blob/main/CHANGELOG.md#2170-november-25-2022), more specifically in #504, we introduced the ability to fetch VM Groups from VCD. However, these methods were failing when the Provider VDC had more than one resource pool.

## RCA

The buggy functions are `GetVmGroupByNamedVmGroupIdAndProviderVdcUrn` and `GetVmGroupByNameAndProviderVdcUrn`. They use a private function called `getResourcePools` which is the culprit. This function is wrongly assuming that there can be only a single resource pool per Provider VDC.

## Fix

This PR changes:

* `getResourcePools` doesn't return error if there is more than one resource pool, as this the natural way of working: A Provider VDC can have multiple Resource pools.
* When multiple resource pools are found, the filter to search a VM Group just considers all of them (by constructing a filter using logical OR forms) like UI does. For example: `filter=(vmGroupName==testVmGroup;(clusterMoref==domain-c12,clusterMoref==domain-c12);(vcId==f583b76e-9e34-48e7-b90d-930653ee1616,vcId==f583b76e-9e34-48e7-b90d-930653ee1616))`
* This relies in the fact that one can't have multiple VM Groups in vSphere with same name on different resource pools.

## Test

To test this, one can create more Resource Pools in the vSphere that serves VCD. In the following image, we have **foo** and **STFTestRootResourcePool**:
Screenshot 2022-12-01 at 15 36 56

Once the Resource Pool is created, one can place some VMs there and create a VM Group there. In the following image, we have a VM Group called **foo1** with **ldap-server** VM (in STFTestRootResourcePool resource pool) and VM Group **asd** with **saddasd** VM (in foo resource pool)
Screenshot 2022-12-01 at 15 37 05
Screenshot 2022-12-01 at 15 37 10

Ideally, we should have something like the above: some VM Groups locating VMs to different VMs served by different Resource Pools.

After we have some VM Groups, we can configure the target Provider VDC to add the new Resource pool:
Screenshot 2022-12-01 at 15 37 22

Test `Test_VmGroupsCRUD` should pass with this patch and this configuration. Without this fix, it should fail with the error
```
could not get VM Group: more than one Resource Pool found for the pVDC
```


dataclouder (approved):

govcd/vdccomputepolicy_v2.go
```diff
@@ -182,28 +182,40 @@ func (vdcComputePolicy *VdcComputePolicyV2) Delete() error {
 // GetAllAssignedVdcComputePoliciesV2 retrieves all VDC assigned Compute Policies (V2) using OpenAPI endpoint. Query parameters can be supplied to perform additional
 // filtering
 func (vdc *AdminVdc) GetAllAssignedVdcComputePoliciesV2(queryParameters url.Values) ([]*VdcComputePolicyV2, error) {
+	return getAllAssignedVdcComputePoliciesV2(vdc.client, vdc.AdminVdc.ID, queryParameters)
+}
+
+// GetAllAssignedVdcComputePoliciesV2 retrieves all VDC assigned Compute Policies (V2) using OpenAPI endpoint. Query parameters can be supplied to perform additional
+// filtering
+func (vcdClient *VCDClient) GetAllAssignedVdcComputePoliciesV2(vdcId string, queryParameters url.Values) ([]*VdcComputePolicyV2, error) {
+	return getAllAssignedVdcComputePoliciesV2(&vcdClient.Client, vdcId, queryParameters)
```
Didainius, line 191: For clarity I think it is worth specifying `vdcId` is mandatory and possibly even fails here before calling `getAllAssignedVdcComputePoliciesV2`. It might return a very odd error because of malformed request url

adambarreiro: Done, thanks for spotting!

Didainius (approved):

lvirbalas (approved):

PR #533, opened 2022-12-09
Change position of BackLinkIdentifier in LDAP type

dataclouder:
The position of the field  was wrong, and prevented update when that field was set.

To test: 
```
go test -tags functional -check.f Test_LDAP -check.vv -timeout 0
```


Didainius (approved):

lvirbalas (approved):
Lovely!

adambarreiro (approved):

PR #540, opened 2023-01-10
Fix test Test_CatalogAccessAsOrgUsers

dataclouder:
The test was failing randomly, due to environmental issues. Depending on the VCD version, the deletion may be slower than expected. Adding a delay+timeout solves the issue.


Didainius (approved):

adambarreiro (approved):
Thanks!

PR #544, opened 2023-01-25
Add support to the Runtime Defined Entities (RDE) framework with RDEs (3/3) 

adambarreiro:
## Overview

This PR implements the foundations to onboard the RDE framework to the Terraform VCD Provider, by adding CRUD capabilities of **RDE instances**.

## Detailed description

This PR adds the following features:

### RDE types

* Adds the following methods to the existing `DefinedEntityType` type:
    * `GetAllRdes` to get all RDE instances from the given type, supports parameters like filtering.
    * `GetRdeByName` to get a unique RDE with given name. To make it unique it uses also the vendor+namespace+version from the given type, as one can have same name with this 3-tuple different combinations.
    * `GetRdeById` gets a unique RDE with given URN.
    * `CreateRde` to create a RDE instance. Creation is tricky, as it just returns a Task that remains at 1% progress in VCD. It will remain like that until it is "resolved" (see below method). So this method creates the RDE in an async way and recovers its refreshed state with an immediate get to return it.

### RDE instances
* Adds the following methods to the **new** `DefinedEntity` type:
    *  `Resolve` to resolve the entity. On creation, the RDE instance is just a Task that remains at 1% in VCD. It needs this method to be called to check the entity against the schema of the RDE type. If it's OK, the task is completed and RDE instance state is RESOLVED. Otherwise, it gets RESOLUTION_ERROR.
    *  `Update` to update the entity.  It is a bit smart to avoid errors if Name is not populated for update (ie: User doesn't want to update it, but it's mandatory). One use case for Update is to fix the JSON entity if we get a RESOLUTION_ERROR.
    * `Delete` to delete the entity. The entity needs to be resolved first.

### Client

* Added `OpenApiGetItemAndHeaders` and `OpenApiPostItemAndGetHeaders` to retrieve all the response headers when performing a call to an OpenAPI endpoint.


lvirbalas (approved):

.changes/v2.20.0/544-features.md
Didainius, line 1: I think a separate changelog record is needed to highlight new `OpenAPIX` functions

govcd/openapi.go
```diff
@@ -120,48 +120,58 @@ func (client *Client) OpenApiGetAllItems(apiVersion string, urlRef *url.URL, que
...
+	_, err := client.OpenApiGetItemAndHeader(apiVersion, urlRef, params, outType, additionalHeader, "")
+	return err
+}
+
+// OpenApiGetItemAndHeader is a low level OpenAPI client function to perform GET request for any item and return a given header.
+// The urlRef must point to ID of exact item (e.g. '/1.0.0/edgeGateways/{EDGE_ID}')
+// It responds with HTTP 403: Forbidden - If the user is not authorized or the entity does not exist. When HTTP 403 is
+// returned this function returns "ErrorEntityNotFound: API_ERROR" so that one can use ContainsNotFound(err) to
+// differentiate when an objects was not found from any other error.
+func (client *Client) OpenApiGetItemAndHeader(apiVersion string, urlRef *url.URL, params url.Values, outType interface{}, additionalHeader map[string]string, headerToGet string) (string, error) {
```
Didainius, line 132: As the `OpenApiX` are very low-level functions for all OpenAPI matters - should we make it more flexible by making a function that gets all headers in one go (we can later get specific headers for the ones who need it) to avoid people calling it multiple times if such a need arises?

govcd/openapi.go
```diff
@@ -251,19 +271,27 @@ func (client *Client) OpenApiPostItemAsync(apiVersion string, urlRef *url.URL, p
...
 // synchronous - it will track task until it is finished and pick reference to marshal outType.
 func (client *Client) OpenApiPostItem(apiVersion string, urlRef *url.URL, params url.Values, payload, outType interface{}, additionalHeader map[string]string) error {
+	_, err := client.OpenApiPostItemAndGetHeader(apiVersion, urlRef, params, payload, outType, additionalHeader, "")
+	return err
+}
+
+// OpenApiPostItemAndGetHeader is a low level OpenAPI client function to perform POST request for item supporting synchronous or
+// asynchronous requests, that returns also the given header value. The urlRef must point to POST endpoint (e.g. '/1.0.0/edgeGateways'). When a task is
+// synchronous - it will track task until it is finished and pick reference to marshal outType.
+func (client *Client) OpenApiPostItemAndGetHeader(apiVersion string, urlRef *url.URL, params url.Values, payload, outType interface{}, additionalHeader map[string]string, headerToRetrieve string) (string, error) {
```
Didainius, line 281: Similar comment to the other functions - could it return all headers so that it is more flexible for other user cases

Didainius (commented):
Looks good, I have one ask that is inline, but can also mention here for the `Client.OpenApiX` functions.

IMO they could all return _all_ headers instead of querying one specific. The reason is that these are our core client functions and people (or we) might find the need to fetch multiple headers in future and that would require multiple API calls instead of one

.changes/v2.20.0/544-features.md
adambarreiro: True! Added. Thanks!

govcd/openapi.go
adambarreiro: That's pretty right, I'm not sure why I didn't do that in the first place, but I've refactored them and now they return all headers for convenience.

govcd/openapi.go
adambarreiro: Same, refactored. Thanks for the suggestion, it is a good improvement indeed.

adezxc (approved):

dataclouder (approved):

PR #546, opened 2023-01-31
Create Test_RenameCatalog

adezxc:
## About
This PR introduces a new test: `Test_RenameCatalog` to check that the contents of the catalog have not changed after renaming it by following these steps:

1. Creates a catalog with the name `TestRenameCatalog`
2. Uploads a vApp template with the name `TestUploadOvf` to the catalog.
3. Retrieves `vAppTemplate1` with the name `TestUploadOvf` from the new catalog object.
4. Updates the catalog name to `TestRenameCatalog_updated`
5. Retrieves a new catalog object using the updated name
6. Retrieves `vAppTemplate2` with the name `TestUploadOvf` from the new catalog object.
7. Compares the vApp templates from `catalog` and `updatedCatalog`

### DIFF
--- a/.changes/v2.20.0/546-notes.md
+++ b/.changes/v2.20.0/546-notes.md
@@ -1,1 +1,1 @@
-* Created `Test_RenameCatalog` for making sure the contents of the Catalog don't change after rename  [GH-546]
+* Created `Test_RenameCatalog` for making sure the contents of the Catalog don't change after rename [GH-546]
--- a/.changes/v2.20.0/546-notes.md
+++ b/.changes/v2.20.0/546-notes.md
@@ -1,0 +1,1 @@
+* Created `Test_RenameCatalog` for making sure the contents of the Catalog don't change after rename  [GH-546]
--- a/govcd/api_vcd_test.go
+++ b/govcd/api_vcd_test.go
@@ -56,6 +56,7 @@ const (
 	TestRefreshOrgFullName        = "TestRefreshOrgFullName"
 	TestUpdateCatalog             = "TestUpdateCatalog"
 	TestDeleteCatalog             = "TestDeleteCatalog"
+	TestRenameCatalog             = "TestRenameCatalog"
 	TestRefreshOrg                = "TestRefreshOrg"
 	TestComposeVapp               = "TestComposeVapp"
 	TestComposeVappDesc           = "vApp created by tests"
--- a/govcd/api_vcd_test.go
+++ b/govcd/api_vcd_test.go
@@ -56,7 +56,6 @@ const (
 	TestRefreshOrgFullName        = "TestRefreshOrgFullName"
 	TestUpdateCatalog             = "TestUpdateCatalog"
 	TestDeleteCatalog             = "TestDeleteCatalog"
-	TestRenameCatalog             = "TestRenameCatalog"
 	TestRefreshOrg                = "TestRefreshOrg"
 	TestComposeVapp               = "TestComposeVapp"
 	TestComposeVappDesc           = "vApp created by tests"
--- a/govcd/catalog_test.go
+++ b/govcd/catalog_test.go
@@ -224,7 +224,7 @@ func (vcd *TestVCD) Test_RenameCatalog(check *C) {
 	check.Assert(err, IsNil)
 	check.Assert(adminCatalog, NotNil)
 	AddToCleanupList(testRenameCatalog, "catalog", vcd.config.VCD.Org, check.TestName())
-	
+
 	catalog, err := vcd.client.Client.GetCatalogByName(vcd.config.VCD.Org, testRenameCatalog)
 	check.Assert(err, IsNil)
 	check.Assert(catalog, NotNil)
--- a/govcd/catalog_test.go
+++ b/govcd/catalog_test.go
@@ -269,6 +269,9 @@ func (vcd *TestVCD) Test_RenameCatalog(check *C) {
 	// catalog and updatedCatalog
 	check.Assert(vAppTemplate1.VAppTemplate.HREF, Equals, vAppTemplate2.VAppTemplate.HREF)
 	check.Assert(mediaImage1.Media.HREF, Equals, mediaImage2.Media.HREF)
+
+	err = updatedCatalog.Delete(true, true)
+	check.Assert(err, IsNil)
 }
 
 // Tests System function UploadOvf by creating catalog and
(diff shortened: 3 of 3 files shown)


puskunalis (approved):

Hohlok (approved):
LGBT!

Didainius:
@puskunalis , @Hohlok ,
Although this project is an open one, please refrain from adding approvals without any input.

Thanks

govcd/api_vcd_test.go
```diff
@@ -56,6 +56,7 @@ const (
 	TestRefreshOrgFullName        = "TestRefreshOrgFullName"
 	TestUpdateCatalog             = "TestUpdateCatalog"
 	TestDeleteCatalog             = "TestDeleteCatalog"
+	TestRenameCatalog             = "TestRenameCatalog"
```
dataclouder, line 59: Instead of using a global constant, define a local variable using `check.Testname()` within the test

govcd/catalog_test.go
```diff
@@ -207,6 +207,55 @@ func doesCatalogExist(check *C, org *AdminOrg) {
...
+	check.Assert(err, IsNil)
+	check.Assert(vAppTemplate1, NotNil)
+
+	adminCatalog.AdminCatalog.Name = TestRenameCatalog + "_updated"
+	err = adminCatalog.Update()
+	check.Assert(err, IsNil)
+
+	// The catalog is added to the cleanup list only after the name update, as
+	// the old one can't be found.
+	AddToCleanupList(TestRenameCatalog+"_updated", "catalog", vcd.config.VCD.Org, check.TestName())
```
dataclouder, line 243: You should add to the cleanup list immediately after the creation, or else, if something fails between creation and update, the resource will not be removed. 
You can keep this second addition. The cleanup function behaves nicely when the resource was already removed.

govcd/catalog_test.go
```diff
@@ -207,6 +207,55 @@ func doesCatalogExist(check *C, org *AdminOrg) {
...
+
+	org, err := vcd.client.GetAdminOrgByName(vcd.config.VCD.Org)
+	check.Assert(err, IsNil)
+	check.Assert(org, NotNil)
+
+	adminCatalog, err := org.CreateCatalog(TestRenameCatalog, TestRenameCatalog)
+	check.Assert(err, IsNil)
+	check.Assert(adminCatalog, NotNil)
+
+	task, err := adminCatalog.UploadOvf(vcd.config.OVA.OvaPath, TestUploadOvf, TestUploadOvf, 1024)
```
dataclouder, line 224: 'TestUploadOvf` should be named after the test itself rather than using a global variable.

govcd/catalog_test.go
```diff
@@ -207,6 +207,55 @@ func doesCatalogExist(check *C, org *AdminOrg) {
...
+	check.Assert(err, IsNil)
+
+	catalog, err := vcd.client.Client.GetCatalogByName(vcd.config.VCD.Org, TestRenameCatalog)
+	check.Assert(err, IsNil)
+	check.Assert(catalog, NotNil)
+
+	vAppTemplate1, err := catalog.GetVAppTemplateByName(TestUploadOvf)
+	check.Assert(err, IsNil)
+	check.Assert(vAppTemplate1, NotNil)
+
```
dataclouder, line 236: Before renaming the catalog, you should also upload a catalog media.

dataclouder (changes requested):

govcd/catalog_test.go
```diff
@@ -207,6 +207,70 @@ func doesCatalogExist(check *C, org *AdminOrg) {
...
+	mediaImage2, err := updatedCatalog.GetMediaByName(testUploadMedia, false)
+	check.Assert(err, IsNil)
+	check.Assert(mediaImage2, NotNil)
+
+	// Check the HREFs of the vApp templates and media images that were retrieved from
+	// catalog and updatedCatalog
+	check.Assert(vAppTemplate1.VAppTemplate.HREF, Equals, vAppTemplate2.VAppTemplate.HREF)
+	check.Assert(mediaImage1.Media.HREF, Equals, mediaImage2.Media.HREF)
+}
```
Didainius, line 272:
```suggestion
	err = updatedCatalog.Delete(true, true)
	check.Assert(err, IsNil)
}
```

One important thing to do at the end of the test - is to cleanup created entities.

You have already added cleanup lines `AddToCleanupList` (and we always do that), but these run at the end of the test suite. However, a test suite runs hundreds of tests, and if we don't clean the items at the end of each test, they would persist until the final cleanup phase. This can impact other tests for various reasons (e.g. name clashes, failing some search mechanism query due to an incorrect expected number of results, etc).

Deleting a catalog with both `force` flags being true should be sufficient because the items you have uploaded (vApp template and media) are children of a `catalog`

govcd/api_vcd_test.go
adezxc: Fixed.

govcd/catalog_test.go
adezxc: Fixed.

govcd/catalog_test.go
adezxc: Fixed with `check.TestName()+"_ovf"|"_media"`

govcd/catalog_test.go
adezxc: Forgot about this method, fixed.

Didainius (approved):
Thanks!

The last note is to add the changelog record (which I see you have started). Because this is only a test addition (with no impact to SDK users), it would most probably best to classify under `notes` section (see example file `538-notes.md` in v2.19.0) instead of `improvements`

.changes/v2.20.0/546-notes.md
```diff
@@ -0,0 +1 @@
+* Created `Test_RenameCatalog` for making sure the contents of the Catalog don't change after rename  [GH-546]
```
adambarreiro, line 1: nit:
```suggestion
* Created `Test_RenameCatalog` for making sure the contents of the Catalog don't change after rename [GH-546]
```

adambarreiro (approved):
Thanks for adding this test!

dataclouder (approved):

.changes/v2.20.0/546-notes.md
adezxc: Fixed

lvirbalas (approved):
Neat and clean! Congratulations with the first PR!

PR #547, opened 2023-02-03
Fix API token URL generation

dataclouder:
The API token URL was created by removing "/api" from the client URL. This method fails when the URL host name starts with "api". ([Issue #994](https://github.com/vmware/terraform-provider-vcd/issues/994))

The new method uses a more robust approach, by combining the existing elements of the client URL.

To test that the API token works, set an API token in your VCD, and then:

```
export TEST_VCD_ORG=System 
export TEST_VCD_API_TOKEN=$YOUR_API_TOKEN 
go test -tags api -check.f TestVCDClient_GetBearerTokenFromApiToken -check.vv -timeout 0

unset TEST_VCD_ORG
unset TEST_VCD_API_TOKEN
```


Didainius (approved):
Thanks. Test worked.

adambarreiro (approved):
Nice fix, thanks!

lvirbalas (approved):

PR #561, opened 2023-03-31
NSX-T DHCP Binding support

Didainius:
This PR adds support for NSX-T DHCP Binding configuration in VCD 10.3.1+ versions. It will be used in PR https://github.com/vmware/terraform-provider-vcd/pull/1039

Tested in:
10.3.0, 10.3.1, 10.4.1


govcd/openapi_org_network_dhcp_binding.go
```diff
@@ -0,0 +1,277 @@
...
+	endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointOrgVdcNetworksDhcpBindings
+	apiVersion, err := client.getOpenApiHighestElevatedVersion(endpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	urlRef, err := client.OpenApiBuildEndpoint(fmt.Sprintf(endpoint, orgVdcNet.OpenApiOrgVdcNetwork.ID))
+	if err != nil {
+		return nil, err
+	}
```
lvirbalas, line 37: Just a thought for the future: maybe we should think how to encapsulate this block into a single call, as it is repeating in different places. On the other hand, being explicit has some benefits too.

lvirbalas (approved):

adezxc (approved):

govcd/openapi_org_network_dhcp_binding.go
```diff
@@ -0,0 +1,277 @@
...
+
+	typeResponses := []*types.OpenApiOrgVdcNetworkDhcpBinding{{}}
+	err = client.OpenApiGetAllItems(apiVersion, urlRef, queryParameters, &typeResponses, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// Wrap all typeResponses into OpenApiOrgVdcNetwork types with client
```
adambarreiro, line 95:
```suggestion
	// Wrap all typeResponses into OpenApiOrgVdcNetworkDhcpBinding types with client
```

adambarreiro (approved):

govcd/openapi_org_network_dhcp_binding.go
Didainius: I have some thoughts to experiment with generics for reusing a lot of it, but needs some experimentation. In some cases, we have to do some "custom" things (mainly handling some "one off" VCD API behaviors).

govcd/openapi_org_network_dhcp_binding.go
Didainius: Fixed.

PR #562, opened 2023-04-03
Add ability to authorize using a Service Account 

adezxc:
This PR is related to issue [terraform-provider-vcd#995](https://github.com/vmware/terraform-provider-vcd/issues/995)

Add support for authenticating using a [Service Account](https://blogs.vmware.com/cloudprovider/2022/07/cloud-director-service-accounts.html)

For now, the PR only enables using Service accounts that are in `Active Stage` so the user is expected to provide an file containing an api token `.json` file e.g
```json
{"refresh_token":"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"}
```

So the SDK will do the following steps:

1. Read File contents
2. Get bearer token and new `refresh_token` using `SetApiToken`
3. Save the new `refresh_token` to the same file


govcd/api_token.go
```diff
@@ -31,6 +33,48 @@ func (vcdClient *VCDClient) SetApiToken(org, apiToken string) (*types.ApiTokenRe
...
+	if err != nil {
+		return err
+	}
+
+	err = os.WriteFile(apiTokenFile, data, 0600)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
adambarreiro, line 75: Should we make errors more verbose/explanatory or is it not needed?

adezxc: I added more verbose errors upon which a user can act upon. Left others as is.

README.md
```diff
@@ -133,6 +133,11 @@ For the above two methods, you use:
 The file `scripts/get_token.sh` provides a handy method of extracting the token
 (`x-vcloud-authorization` value) for future use.
 
+* With a service account token
+```go
+	err := vcdClient.SetServiceAccountApiToken(Org, "tokenfile.json")
```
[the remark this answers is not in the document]
adezxc: Changed.

adambarreiro (approved):
Approving. Maybe in the near future we can check how to provide a testing Service account to create a test for `SetServiceAccountApiToken `.

Didainius (approved):

govcd/generic_functions.go
```diff
@@ -36,3 +42,34 @@ func oneOrError[T any](key, name string, entitySlice []*T) (*T, error) {
 
 	return entitySlice[0], nil
 }
+
+// readFileAndUnmarshalJSON reads a file and unmarshals it to the given variable
+func readFileAndUnmarshalJSON[T any](filename string, object T) error {
```
dataclouder, line 47: Why using generics here? The object is `any`, as it the parameter to `json.Unmarshal`, so we don't gain anything by using a definable type

govcd/generic_functions.go
```diff
@@ -36,3 +42,34 @@ func oneOrError[T any](key, name string, entitySlice []*T) (*T, error) {
...
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal file contents to the object: %s", err)
+	}
+
+	return nil
+}
+
+// marshalJSONAndWriteToFile marshalls the given object into JSON and writes
+// to a file with the given permissions in octal format (e.g 0600)
+func marshalJSONAndWriteToFile[T any](filename string, object *T, permissions int) error {
```
dataclouder, line 63: Same observation given for `readFileAndUnmarshalJSON`. I don't see the need for using generics

govcd/generic_functions.go
adezxc: True, the reason was just to extract the I/O functions so we could test them. I guess I can just make these into normal functions.

adezxc: I made the functions non-generic, but I don't know where I should move these utility functions, as they don't fit in generic_functions.go

dataclouder: I think these functions can stay in api_token.go, as they are private functions that deal with such matter

dataclouder (approved):

PR #565, opened 2023-04-13
Fix wrong assumption of System administrator in VCD Client properties

adambarreiro:
## Problem

Currently, to determine the `IsSysAdmin` property from the VCDClient struct, we just compare the current logged in user's Organization, if it's System, we say it is a System Administrator:

```
vcdClient.Client.IsSysAdmin = strings.EqualFold(org, "system")
```

This assumption is wrong, as one can have a user in System that doesn't have System Administrator role. Some examples:

- A "CSE Administrator" account belongs to System, but has a limited set of rights given by a new role "CSE Administrator" that was created in System.
- I create a "Not a System Admin" role by cloning "System Administrator" role. I remove one random right from the new role. A user with this role is no longer a System administrator.

## Solution

The solution is tricky. Let's review some discarded options:

- Comparing current Organization with "System", this is the current one and we saw above that **it's not valid**.
- Comparing user's role with "System Administrator". This is not valid, as one could clone this role to a new "System Administrator Clone", a user with this new role would be still a System Administrator as **it has all the rights**.

The implemented solution starts from this last idea: **it has all the rights**. We need to compare the rights of "System Administrator", which has every single one of them and is a read only role, with the role of the logged in user.

So in the solution we get role and rights from the user, then "System Administrator" role and rights, and compare their length. If it's the same, then is a SysAdmin.

### DIFF
--- a/.changes/v2.21.0/565-bug-fixes.md
+++ b/.changes/v2.21.0/565-bug-fixes.md
@@ -1,0 +1,2 @@
+* Fix a bug that wrongly calculated `IsSystem` flag on `Client` when the user is in System organization
+  but doesn't have enough rights [GH-565]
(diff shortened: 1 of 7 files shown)


adambarreiro:
Closing this PR as I think this was a misunderstanding from my side, as the affected flag is used more like "IsSystemOrg" rather than to detect that is the System Administrator with all its rights. If one searches throughout the code this behaviour can be seen, so this PR doesn't make any sense.

PR #532, opened 2022-12-09
Expand NSX-T Edge Gateway support for IP allocation management

Didainius:
### About

This PR introduced a few new functions for Edge Gateway IP allocation management:
* `NsxtEdgeGateway.Refresh`
* `NsxtEdgeGateway.GetUsedIpAddresses`
It uses direct VCD endpoint `edgeGateways/%s/usedIpAddresses`

* `NsxtEdgeGateway.GetUsedIpAddressSlice` is using the same endpoint as `NsxtEdgeGateway.GetUsedIpAddresses`, but returns values using Go's native `[]netip.Addr`
* `NsxtEdgeGateway.GetAllocatedIpCount` that traverses Edge Gateway structure and sums up amount of `TotalIPCount` in each subnet value
* `NsxtEdgeGateway.QuickDeallocateIpCount` and `NsxtEdgeGateway.DeallocateIpCount`
  functions to manually alter Edge Gateway body for IP deallocation (API does not officially support deallocation using negative numbers in `QuickAddAllocatedIPCount` even though it works in some versions)

* `NsxtEdgeGateway.GetUnusedExternalIPAddresses` - is a function which performs the following actions to get the result
1. Retrieves a complete list of IPs in Edge Gateway uplinks (returns error if none are found)
2. if 'optionalSubnet' was specified - filter IP addresses to only fall into that subnet
3. Retrieves all used IP addresses in Edge Gateway using dedicated API endpoint
4. Subtracts used IP addresses from available list of IPs in uplink (optionally filtered by optionalSubnet in step 2)
5. Checks if 'requiredIpCount' criteria is met, returns error otherwise
6. Returns required amount of unused IPs (as defined in 'requiredIpCount')

* `NsxtEdgeGateway.GetAllUnusedExternalIPAddresses` - almost the same as `NsxtEdgeGateway.GetUnusedExternalIPAddresses`, but without steps 2, 5 and 6:

1. Retrieves a complete list of IPs in Edge Gateway uplinks (returns error if none are found)
2. Retrieves all used IP addresses in Edge Gateway using dedicated API endpoint
3. Subtracts used IP addresses from available list of IPs in uplink (optionally filtered by optionalSubnet in step 2)

**Note.** Functions use Go's built-in `netip` package to handle IPs and prefixes. It provides a robust solution to support IPv4 and IPv6 addresses and ease dealing with IP ranges and comparisons which wouldn't be easy to deal when operating with strings (especially IPv6)

#### Not directly related changes
[...]

PR #567, opened 2023-04-17
Improve 'ipSliceDifference' function memory footprint

Didainius:
Methods `NsxtEdgeGateway.GetAllUnusedExternalIPAddresses` and `NsxtEdgeGateway.GetUnusedExternalIPAddresses` perform computations based on `netip.Addr` package.

The internal function `ipSliceDifference` is used to allocate a new slice to hold the final results. However, the operations become quite expensive memory wise having big subnets in Edge Gateway. Optimizing this function reduced memory consumption twice (see example below)

##### Example information

One example was allocation `/8` `IPv4` subnet (`16777213` IPs)

Before optimization 2 operations consumed such an amount of memory (`getAllUnusedExternalIPAddresses` - **7703.94MB**)
image

After optimization 2 operations of `getAllUnusedExternalIPAddresses` consume such amount (after **3853.80MB** vs before **7703.94MB**):
image

### DIFF
--- a/.changes/v2.20.0/532-features.md
+++ b/.changes/v2.20.0/532-features.md
@@ -4,9 +4,9 @@
 * Added `NsxtEdgeGateway.GetUsedIpAddressSlice` method to fetch used IP addresses in a slice
   [GH-532]
 * Added `NsxtEdgeGateway.GetUnusedExternalIPAddresses` method that can help to find an unused
-  IP address in an Edge Gateway by given constraints [GH-532]
+  IP address in an Edge Gateway by given constraints [GH-532,GH-567]
 * Added `NsxtEdgeGateway.GetAllUnusedExternalIPAddresses` method that can return all unused IP
-  addresses in an Edge Gateway [GH-532]
+  addresses in an Edge Gateway [GH-532,GH-567]
 * Added `NsxtEdgeGateway.GetAllocatedIpCount` method that sums up `TotalIPCount` fields in all
   subnets [GH-532]
 * Added `NsxtEdgeGateway.QuickDeallocateIpCount` and `NsxtEdgeGateway.DeallocateIpCount`
--- a/govcd/nsxt_edgegateway.go
+++ b/govcd/nsxt_edgegateway.go
@@ -749,57 +749,6 @@ func flattenEdgeGatewayUplinkToIpSlice(uplinks []types.EdgeGatewayUplinks) ([]ne
 	return assignedIpSlice, nil
 }
 
-// ipSliceDifferenceOld performs mathematical subtraction for two slices of IPs
-// The formula is (minuend − subtrahend = difference)
-//
-// Special behavior:
-// * Passing nil minuend results in nil
-// * Passing nil subtrahend will return minuendSlice
-func ipSliceDifferenceOld(minuendSlice, subtrahendSlice []netip.Addr) []netip.Addr {
-	if minuendSlice == nil {
-		return nil
-	}
-
-	if subtrahendSlice == nil {
-		return minuendSlice
-	}
-
-	// Removal of elements from an empty slice results in an empty slice
-	if len(minuendSlice) == 0 {
-		return []netip.Addr{}
-	}
-	// Having an empty subtrahendSlice results in minuendSlice
-	if len(subtrahendSlice) == 0 {
-		return minuendSlice
-	}
-
-	var difference []netip.Addr
-
-	// Loop over minuend IPs
-	for _, minuendIp := range minuendSlice {
-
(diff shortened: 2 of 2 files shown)


govcd/nsxt_edgegateway.go
```diff
@@ -755,6 +755,9 @@ func flattenEdgeGatewayUplinkToIpSlice(uplinks []types.EdgeGatewayUplinks) ([]ne
 // Special behavior:
 // * Passing nil minuend results in nil
 // * Passing nil subtrahend will return minuendSlice
+//
+// NOTE. This function will mutate minuendSlice to save memory and avoid having a copy of all values
+// which can become expensive if there are a lot of items
 func ipSliceDifference(minuendSlice, subtrahendSlice []netip.Addr) []netip.Addr {
```
adambarreiro, line 761: Just wondering whether this function should return nothing at all to state that is mutating `minuendSlice`?

.changes/v2.20.0/532-features.md
```diff
@@ -4,9 +4,9 @@
 * Added `NsxtEdgeGateway.GetUsedIpAddressSlice` method to fetch used IP addresses in a slice
   [GH-532]
 * Added `NsxtEdgeGateway.GetUnusedExternalIPAddresses` method that can help to find an unused
-  IP address in an Edge Gateway by given constraints [GH-532]
+  IP address in an Edge Gateway by given constraints [GH-532,GH-567]
 * Added `NsxtEdgeGateway.GetAllUnusedExternalIPAddresses` method that can return all unused IP
-  addresses in an Edge Gateway [GH-532]
+  addresses in an Edge Gateway [GH-532,GH-567]
```
adambarreiro, line 9: I have a doubt here, as I used a different syntax in my PR:
https://github.com/vmware/go-vcloud-director/pull/566
Although I tried to search and it seems both are used in changelog 😄 
Which one should we use?

Didainius: The answer is in `scripts/changelog-links.sh`

Such message
```
* Added `NsxtEdgeGateway.GetAllUnusedExternalIPAddresses` method that can return all unused IP addresses in an Edge Gateway [GH-532,GH-567]
```

Renders: 

* Added `NsxtEdgeGateway.GetAllUnusedExternalIPAddresses` method that can return all unused IP addresses in an Edge Gateway ([#532](https://github.com/vmware/go-vcloud-director/pull/532),[#567](https://github.com/vmware/go-vcloud-director/pull/567))


While such message

```
* Added support for Runtime Defined Entity Interfaces with client methods
  `VCDClient.CreateDefinedInterface`, `VCDClient.GetAllDefinedInterfaces`,
  `VCDClient.GetDefinedInterface`, `VCDClient.GetDefinedInterfaceById` and methods to manipulate
  them `DefinedInterface.Update`, `DefinedInterface.Delete` [GH-527], [GH-566]
```

Renders:
* Added support for Runtime Defined Entity Interfaces with client methods
  `VCDClient.CreateDefinedInterface`, `VCDClient.GetAllDefinedInterfaces`,
  `VCDClient.GetDefinedInterface`, `VCDClient.GetDefinedInterfaceById` and methods to manipulate
  them `DefinedInterface.Update`, `DefinedInterface.Delete` ([#527](https://github.com/vmware/go-vcloud-director/pull/527)], [[#566](https://github.com/vmware/go-vcloud-director/pull/566))

govcd/nsxt_edgegateway.go
Didainius: Apparently, I can't do exactly that (similar to why built-in `append` has similar mechanics).

https://forum.golangbridge.org/t/slice-pass-as-value-or-pointer/2866

> the contents of a slice argument can be modified by a function, but its header cannot. The length stored in the slice variable is not modified by the call to the function, since the function is passed a copy of the slice header, not the original. Thus if we want to write a function that modifies the header, we must return it as a result parameter, just as we have done here.


https://go.dev/blog/slices -> **Passing slices to functions**.

The way around it could be `*[]netip.Addr`, but this misses the semantics of `[]netip.Addr` everywhere.

adambarreiro: Very interesting read, thanks for sharing! So this is all good then 😄

.changes/v2.20.0/532-features.md
adambarreiro: Thanks! This clarifies it! Time to change #566 👍

adambarreiro (approved):
LGTM! Thanks for answering the doubts, it helped a lot.

adezxc (approved):

dataclouder (approved):

PR #572, opened 2023-05-01
Fix disabled org creations in 10.4.2 tests

dataclouder:
Due to a VCD bug in 10.4.2, creating a disabled Org is not possible. As a workaround, we change the enabled status of the Org being created for this specific VCD version, during Org tests.

Additionally, we raise the test timeout, which can sometimes be insufficient.


.changes/v2.21.0/572-bug-fixes.md
```diff
@@ -0,0 +1,2 @@
+* Changed Org enablement status during tests for VCD 10.4.2, to circumvent a VCD bug that prevents creation of disabled Orgs [GH-572]
```
Didainius, line 1: That is a test only patch - should we keep it under "notes"?

Didainius (approved):

.changes/v2.21.0/572-bug-fixes.md
dataclouder: Changed

adezxc (approved):

lvirbalas (approved):
I like how you narrow down the change of the test behavior only for the affected VCD versions, thanks!

adambarreiro (approved):

PR #574, opened 2023-05-08
Skip policy clean error test in 10.4.2

dataclouder:
The bug that the test checks for is fixed in 10.4.2, so we can skip the test safely.


adambarreiro (approved):
Thanks!

lvirbalas (approved):

adezxc (approved):

PR #577, opened 2023-05-26
Support managing API tokens and Service Accounts

adezxc:
Add methods for managing API tokens and Service Accounts

New public methods for token management:
`*VCDClient.CreateToken`,`*VCDClient.GetTokenById`,`*VCDClient.GetAllTokens`,`*VCDClient.GetTokenByNameAndUsername`, `*VCDClient.RegisterToken` , `*Token.GetInitialApiToken`, `*Token.Delete`

`*Client.GetApiToken` - get the structure containing the access token and the bearer token using the Service Account/API token

Public methods for service account management:
`*Org.GetServiceAccountById`, `*Org.GetAllServiceAccounts`, `*Org.GetServiceAccountByName`

`*VCDClient.CreateServiceAccount`

`*ServiceAccount.Update`, `*ServiceAccount.Authorize`, `*ServiceAccount.Grant`, `*ServiceAccount.Refresh`, `*ServiceAccount.Revoke`, `*ServiceAccount.Delete`, `*ServiceAccount.GetInitialApiToken`

Public OpenAPI function for sending a POST request in `x-www-form-urlencoded` format: `client.OpenApiPostUrlEncoded`


govcd/service_account.go
```diff
@@ -0,0 +1,359 @@
...
+		return nil, fmt.Errorf("error updating Service Account: %s", err)
+	}
+
+	return returnServiceAccount, nil
+}
+
+// AuthorizeServiceAccount authorizes a service account and returns a DeviceID and UserCode which will be used while granting
+// the request, and sets the Service Account in `Requested` status
+func (sa *ServiceAccount) Authorize() error {
+	client := sa.org.client
```
dataclouder, line 192: Do we need to check the service account current status?
If not, does the API return a clear error message if we do the wrong thing?

govcd/service_account.go
```diff
@@ -0,0 +1,359 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// GrantServiceAccount Grants access to the Service Account and sets it in `Granted` status
+func (sa *ServiceAccount) Grant() error {
+	client := sa.org.client
```
dataclouder, line 220: Do we need to check the service account current status?
If not, does the API return a clear error message if we do the wrong thing?

govcd/service_account.go
```diff
@@ -0,0 +1,359 @@
...
+
+	err = client.OpenApiPostItem(apiVersion, urlRef, nil, nil, nil, nil)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// DeleteServiceAccountByID deletes a Service Account
```
dataclouder, line 305:
```suggestion
// Delete deletes a Service Account
```

govcd/service_account_test.go
```diff
@@ -0,0 +1,140 @@
+//go:build api || functional || ALL
+
+/*
+ * Copyright 2021 VMware, Inc.  All rights reserved.  Licensed under the Apache v2 License.
```
dataclouder, line 4:
```suggestion
 * Copyright 2023 VMware, Inc.  All rights reserved.  Licensed under the Apache v2 License.
```

.changes/v2.21.0/577-features.md
```diff
@@ -0,0 +1 @@
+* Added support for Service Account and API token management [GH-577]
```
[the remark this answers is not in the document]
adezxc: Updated

govcd/api_token_test.go
```diff
@@ -51,3 +52,72 @@ func (vcd *TestVCD) TestVCDClient_GetBearerTokenFromApiToken(check *C) {
 	check.Assert(tokenInfo.ExpiresIn, Not(Equals), 0)
 	check.Assert(tokenInfo.TokenType, Equals, "Bearer")
 }
+
+func (vcd *TestVCD) Test_ApiToken(check *C) {
```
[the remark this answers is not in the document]
adezxc: Changed.

govcd/service_account.go
adezxc: Authorize returns a reasonable error if an account is already granted or active, if not, it will work every time, so we don't need to check the status.

govcd/service_account.go
adezxc: Updated. Added a check if the userCode is set, if yes, then it's atleast authorized. If it's already granted it will say that it failed to grant the SA as the usercode is invalid.

govcd/service_account.go
adezxc: Changed.

govcd/service_account_test.go
adezxc: Changed

types/v56/types.go
```diff
@@ -3247,6 +3247,25 @@ type UpdateVdcStorageProfiles struct {
 	RemoveStorageProfile *Reference                      `xml:"RemoveStorageProfile,omitempty"`
 }
 
+type Token struct {
```
Didainius, line 3250: Not necessarily for all fields, but at least a short information what this struct is used for in general would be good. Same for `ServiceAccount` and `ServiceAccountAuthParams`

govcd/api_token.go
```diff
@@ -21,122 +21,213 @@ import (
 	"github.com/vmware/go-vcloud-director/v2/util"
 )
 
-// SetApiToken behaves similarly to SetToken, with the difference that it will
-// return full information about the bearer token, so that the caller can make decisions about token expiration
-func (vcdClient *VCDClient) SetApiToken(org, apiToken string) (*types.ApiTokenRefresh, error) {
-	tokenRefresh, err := vcdClient.GetBearerTokenFromApiToken(org, apiToken)
+// TODO Have distinct names for API and Refresh tokens
```
Didainius, line 24: is this TODO for now, or for future?

Didainius (commented):
A quick scan with a few notes. I will get back for a more in depth review.

govcd/api_token.go
adezxc: For future, right now we can't change the public functions and as we have API token/Service Account overlapping RefreshToken structure, it would be nice to have a distinction in code between API tokens and Refresh tokens.

types/v56/types.go
adezxc: Updated.

govcd/service_account.go
```diff
@@ -0,0 +1,371 @@
...
+		return fmt.Errorf("minimum API version for Service Accounts is 37.0 - Version detected: %s", client.APIVersion)
+	}
+
+	endpoint := types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointServiceAccounts
+	urlRef, err := client.OpenApiBuildEndpoint(endpoint, sa.ServiceAccount.ID)
+	if err != nil {
+		return err
+	}
+
+	err = client.OpenApiDeleteItem("37.0", urlRef, nil, nil)
```
Didainius, line 335: You could leverage `endpointMinApiVersions` structure and use `getOpenApiHighestElevatedVersion` - this will avoid having hardcoded API version in a separate place and also do the version check automatically that will make this part unneeded https://github.com/vmware/go-vcloud-director/pull/577/files#diff-3743a8e278e471adff2ca7110df5cc98b9600165b09face7e71dd9bfef78ea54R319-R327

Just like in the other place you have done - https://github.com/vmware/go-vcloud-director/pull/577/files#diff-3743a8e278e471adff2ca7110df5cc98b9600165b09face7e71dd9bfef78ea54R299

govcd/service_account.go
```diff
@@ -0,0 +1,371 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return token, nil
+}
+
+// Refresh updates the Service Account object
+func (sa *ServiceAccount) Refresh() error {
+	if sa.ServiceAccount == nil || sa.org.client == nil || sa.ServiceAccount.ID == "" {
+		return fmt.Errorf("cannot refresh Edge Gateway without ID")
```
Didainius, line 282:
```suggestion
		return fmt.Errorf("cannot refresh Service Account without ID")
```

govcd/service_account.go
```diff
@@ -0,0 +1,371 @@
...
+	if err != nil {
+		return fmt.Errorf("error granting service account: %s", err)
+	}
+
+	tenantContext, err := sa.org.getTenantContext()
+	if err != nil {
+		return fmt.Errorf("error granting service account: %s", err)
+	}
+
+	err = client.OpenApiPostItem("37.0", urlRef, nil, userCode, nil, getTenantContextHeader(tenantContext))
```
Didainius, line 250: Similar comment as in other place - using `getOpenApiHighestElevatedVersion` would help you get API version in a centralised way and avoid hardcoding API version

govcd/api_token.go
```diff
@@ -155,7 +241,163 @@ func (vcdClient *VCDClient) GetBearerTokenFromApiToken(org, token string) (*type
...
+	req.Header.Add("Accept", fmt.Sprintf("application/*;version=%s", apiVersion))
+	req.Header.Add("Content-Type", contentType)
+
+	resp, err := client.Http.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("error getting request URL from %s : %s", reqHref.String(), err)
+	}
+
+	return checkRespWithErrType(types.BodyTypeJSON, resp, err, &types.OpenApiError{})
+}
```
Didainius, line 274: Can this also be done by some generic function in `openapi.go`?

govcd/api_token.go
```diff
@@ -21,122 +21,208 @@ import (
...
+	if err != nil {
+		return nil, err
 	}
 
-	// The default response data to show in the logs is a string of asterisks
 	responseData := "[" + strings.Repeat("*", 10) + "]"
 	// If users request to see sensitive data, we pass the unchanged response body
 	if util.LogPasswords {
 		responseData = string(body)
 	}
```
Didainius, line 216: There is a centralised function `hideSensitive` that captures sensitive data and filters it

govcd/api_token.go
```diff
@@ -21,122 +21,208 @@ import (
...
 	if err != nil {
-		return nil, fmt.Errorf("error getting request URL from %s : %s", reqUrl, err)
+		return nil, fmt.Errorf("error getting request URL from %s : %s", urlRef.String(), err)
 	}
 
-	data := bytes.NewBufferString(fmt.Sprintf("grant_type=refresh_token&refresh_token=%s", token))
-	req := vcdClient.Client.NewRequest(nil, http.MethodPost, *reqHref, data)
-	req.Header.Add("Accept", "application/*;version=36.1")
+	newTokenParams := &types.ApiTokenParams{}
+	err = client.OpenApiPostItemSync("36.1", urlRef, nil, tokenParams, newTokenParams)
```
Didainius, line 188: Could we leverage `getOpenApiHighestElevatedVersion` here based on endpoint to avoid having hardcoded API version

govcd/api_token.go
```diff
@@ -21,122 +21,208 @@ import (
...
+	client := vcdClient.Client
+	if client.APIVCDMaxVersionIs("< 36.1") {
+		version, err := client.GetVcdFullVersion()
 		if err == nil {
 			return nil, fmt.Errorf("minimum version for API token is 10.3.1 - Version detected: %s", version.Version)
 		}
 		// If we can't get the VCD version, we return API version info
-		return nil, fmt.Errorf("minimum API version for API token is 36.1 - Version detected: %s", vcdClient.Client.APIVersion)
+		return nil, fmt.Errorf("minimum API version for API token is 36.1 - Version detected: %s", client.APIVersion)
+	}
```
Didainius, line 153: `GetAllTokens` will do version check within `getOpenApiHighestElevatedVersion` https://github.com/vmware/go-vcloud-director/pull/577/files#diff-d988e622a91bb1b92dd39fd3e26c321c99e2f78f21633147b3604f5a44a572ceR116

govcd/api_token.go
```diff
@@ -21,122 +21,208 @@ import (
...
+func (vcdClient *VCDClient) GetTokenById(tokenId string) (*Token, error) {
+	client := vcdClient.Client
+	if client.APIVCDMaxVersionIs("< 36.1") {
+		version, err := client.GetVcdFullVersion()
+		if err == nil {
+			return nil, fmt.Errorf("minimum version for API token is 10.3.1 - Version detected: %s", version.Version)
+		}
+		// If we can't get the VCD version, we return API version info
+		return nil, fmt.Errorf("minimum API version for API token is 36.1 - Version detected: %s", client.APIVersion)
+	}
```
Didainius, line 78: `getOpenApiHighestElevatedVersion` a few lines below will do version check

govcd/api_token.go
```diff
@@ -21,122 +21,208 @@ import (
...
+func (vcdClient *VCDClient) CreateToken(org, tokenName string) (*Token, error) {
+	client := vcdClient.Client
+	if client.APIVCDMaxVersionIs("< 36.1") {
+		version, err := client.GetVcdFullVersion()
+		if err == nil {
+			return nil, fmt.Errorf("minimum version for API token is 10.3.1 - Version detected: %s", version.Version)
+		}
+		// If we can't get the VCD version, we return API version info
+		return nil, fmt.Errorf("minimum API version for API token is 36.1 - Version detected: %s", client.APIVersion)
+	}
```
Didainius, line 44: The first API call this function does (`GetTokenById`) will perform version validation

Didainius (changes requested):
Thanks for it - I have left improvement comments inline

govcd/api_token.go
```diff
@@ -21,122 +21,213 @@ import (
...
+
+	resp, err := client.doTokenRequest(org, "register", apiVersion, "application/json", bytes.NewBuffer(data))
+	if resp != nil {
+		defer resp.Body.Close()
+	}
+
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
adezxc: Removed this altogether and moved to OpenApiPostUrlEncoded

govcd/api_token.go
```diff
@@ -21,122 +21,213 @@ import (
...
+		return nil, err
 	}
 
-	data := bytes.NewBufferString(fmt.Sprintf("grant_type=refresh_token&refresh_token=%s", token))
-	req := vcdClient.Client.NewRequest(nil, http.MethodPost, *reqHref, data)
-	req.Header.Add("Accept", "application/*;version=36.1")
+	return token, nil
+}
+
+func (vcdClient *VCDClient) RegisterToken(org, apiVersion string, tokenParams *types.ApiTokenParams) (*types.ApiTokenParams, error) {
```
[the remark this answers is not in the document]
adezxc: Updated to not ask for apiVersion.

govcd/api_token.go
```diff
@@ -21,122 +21,213 @@ import (
...
+	err = json.Unmarshal(body, newTokenParams)
+	if err != nil {
+		return nil, err
+	}
+
+	return newTokenParams, nil
+}
+
+// GetAccessToken gets the access token structure containing the bearer token
+func (client *Client) GetAccessToken(org, apiVersion, funcName string, data *bytes.Buffer) (*types.ApiTokenRefresh, error) {
```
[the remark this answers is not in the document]
adezxc: Made the function private and removed apiVersion argument

govcd/api_token.go
```diff
@@ -21,122 +21,208 @@ import (
...
+	if client.APIVCDMaxVersionIs("< 36.1") {
+		version, err := client.GetVcdFullVersion()
 		if err == nil {
-			return fmt.Errorf("minimum version for Service Account authentication is 10.4 - Version detected: %s", version.Version)
+			return nil, fmt.Errorf("minimum version for API token is 10.3.1 - Version detected: %s", version.Version)
 		}
 		// If we can't get the VCD version, we return API version info
-		return fmt.Errorf("minimum API version for Service Account authentication is 37.0 - Version detected: %s", vcdClient.Client.APIVersion)
+		return nil, fmt.Errorf("minimum API version for API token is 36.1 - Version detected: %s", client.APIVersion)
 	}
```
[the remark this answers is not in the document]
adezxc: Fixed.

govcd/service_account.go
adezxc: Fixed.

govcd/service_account.go
adezxc: Fixed, thanks for spotting.

govcd/service_account.go
adezxc: Fixed.

govcd/api_token.go
adezxc: Created a function, `OpenApiPostUrlEncoded` to handle this on a lower level.

govcd/api_token.go
adezxc: Updated `hideSensitive` to handle the new cases for the refresh token and access token from `getAccessToken`

govcd/api_token.go
adezxc: It isn't a CloudAPI endpoint, so there's no minimal version. Kept the checks as they are.

govcd/api_token.go
adezxc: Fixed.

govcd/api_token.go
adezxc: Fixed.

govcd/api_token.go
adezxc: Fixed.

govcd/openapi.go
```diff
@@ -334,6 +334,55 @@ func (client *Client) OpenApiPostItemAndGetHeaders(apiVersion string, urlRef *ur
...
+	body := strings.NewReader(urlValues.Encode())
+
+	// Create the header map if it's nil
+	if additionalHeaders == nil {
+		additionalHeaders = make(map[string]string)
+	}
+	// Overwrite the Content-Type header as this is a method only usable for x-www-form-urlencoded
+	additionalHeaders["Content-Type"] = "application/x-www-form-urlencoded"
+
+	req := client.newOpenApiRequest("", params, http.MethodPost, urlRef, body, additionalHeaders)
```
Didainius, line 358: Is the `apiVersion` not missing?

govcd/openapi.go
```diff
@@ -334,6 +334,55 @@ func (client *Client) OpenApiPostItemAndGetHeaders(apiVersion string, urlRef *ur
 	return resp.Header, nil
 }
 
+// OpenApiPostUrlEncoded is a non-standard function used to send a POST request with `x-www-form-urlencoded` format.
+// Accepts a map in format of key:value, marshals the response body in JSON format to outType.
+func (client *Client) OpenApiPostUrlEncoded(urlRef *url.URL, params url.Values, payloadMap map[string]string, outType interface{}, additionalHeaders map[string]string) error {
```
Didainius, line 339: I am still thinking if we could actually use existing `OpenApiPostItem` as it looks very similar.

IIRC, the main difference here is the `Content-Type` which by default we assume is `types.JSONMime (application/json)`.

We already have  the patch https://github.com/vmware/go-vcloud-director/pull/577/files#diff-585772f0a826713495acbb0951db5ff9d9f252426fca85e2e112581e5d90a29dR758-R761 that provides mechanism to override `additionalHeaders["Content-Type"] = "application/x-www-form-urlencoded"`

If the only problem remaining now is that `openApiPerformPostPut` tries to "beautify" JSON indentation - we could check if we are overriding `Content-Type` to `application/x-www-form-urlencoded` via `additionalHeader` parameter and skip it if so

Didainius (approved):
Thank you for all these fixes. I am actually good now - the only remaining comment from is to see if we could use `OpenApiPostItem` instead of `OpenApiPostUrlEncoded` as they look to be very very similar (see my comment inline). Not insisting on that if it doesn't feel like a good approach.

adambarreiro (approved):
Great work, approved.

govcd/openapi.go
adezxc: Well, I think the main point is that without the new function, we would expect the caller function to provide properly encoded string for `x-www-form-urlencoded` instead of a map, like it is now which is IMO more intuitive. I guess we could avoid the new function by creating an if statement, but we should agree on the type we should expect from the user.

govcd/service_account.go
```diff
@@ -0,0 +1,339 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	tenantContext, err := org.getTenantContext()
+	if err != nil {
+		return nil, err
+	}
+
+	// VCD has a pageSize limit on this specifi endpoint
```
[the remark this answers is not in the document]
adezxc: Fixed, thanks for spotting.

govcd/service_account.go
```diff
@@ -0,0 +1,339 @@
...
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+)
+
+type ServiceAccount struct {
+	ServiceAccount *types.ServiceAccount
+	authParams     *types.ServiceAccountAuthParams
+	org            *Org
+}
+
+func (org *Org) GetServiceAccountById(serviceAccountId string) (*ServiceAccount, error) {
```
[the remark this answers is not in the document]
adezxc: Definitely, updated now.

govcd/api_token_test.go
```diff
@@ -51,3 +52,118 @@ func (vcd *TestVCD) TestVCDClient_GetBearerTokenFromApiToken(check *C) {
...
+	check.Assert(newToken, IsNil)
+}
+
+func (vcd *TestVCD) Test_GetFilteredTokensOrg(check *C) {
+	isApiTokenEnabled, err := vcd.client.Client.VersionEqualOrGreater("10.3.1", 3)
+	check.Assert(err, IsNil)
+	if !isApiTokenEnabled {
+		check.Skip("This test requires VCD 10.3.1 or greater")
+	}
+	orgName := vcd.config.Tenants[0].SysOrg
```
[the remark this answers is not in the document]
adezxc: Thanks for the suggestion, updated.

govcd/api_token.go
```diff
@@ -1,161 +1,302 @@
...
-// authentication request using SetApiToken and overwrites the old file.
-func (vcdClient *VCDClient) SetServiceAccountApiToken(org, apiTokenFile string) error {
-	if vcdClient.Client.APIVCDMaxVersionIs("< 37.0") {
-		version, err := vcdClient.Client.GetVcdFullVersion()
-		if err == nil {
-			return fmt.Errorf("minimum version for Service Account authentication is 10.4 - Version detected: %s", version.Version)
-		}
-		// If we can't get the VCD version, we return API version info
-		return fmt.Errorf("minimum API version for Service Account authentication is 37.0 - Version detected: %s", vcdClient.Client.APIVersion)
+func (vcdClient *VCDClient) GetAllTokens(queryParameters url.Values) ([]*Token, error) {
```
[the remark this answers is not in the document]
adezxc: Updated.

govcd/api_token.go
```diff
@@ -1,161 +1,302 @@
...
+			Token:  typeResponses[sliceIndex],
+			client: &client,
+		}
 	}
-	err = marshalJSONAndWriteToFile(apiTokenFile, saApiToken, 0600)
+
+	return results, nil
+}
+
+// GetTokenByNameForCurrentUser retrieves a Token by name and username
```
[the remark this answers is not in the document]
adezxc: Thanks for spotting, updated.

.changes/v2.21.0/577-features.md
```diff
@@ -0,0 +1,7 @@
+* Added Service Account CRUD support via `ServiceAccount` and `types.ServiceAccount`: `*VCDClient.CreateServiceAccount`,
+  `*Org.GetServiceAccountById`, `*Org.GetAllServiceAccounts`, `*Org.GetServiceAccountByName`, 
+  `*ServiceAccount.Update`, `*ServiceAccount.Authorize`, `*ServiceAccount.Grant`, `*ServiceAccount.Refresh`, 
+  `*ServiceAccount.Revoke`, `*ServiceAccount.Delete`, `*ServiceAccount.GetInitialApiToken` [GH-577]
+* Added API Token CRUD support via `Token` and `types.Token`: `*VCDClient.CreateToken`,`*VCDClient.GetTokenById`,
+`*VCDClient.GetAllTokens`,`*VCDClient.GetTokenByNameAndUsername`, `*VCDClient.RegisterToken` , `*Token.GetInitialApiToken`, `*Token.Delete`, `Client.GetApiToken` [GH-577]
+
```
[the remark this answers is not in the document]
adezxc: Updated.

govcd/openapi.go
adezxc: Fixed.

govcd/openapi.go
Didainius: Ok, agreed. If you could just put a note in that if one provides `Content-Type` in the `additionalHeaders` parameter for `OpenApiPostUrlEncoded` - it will be overriden to `application/x-www-form-urlencoded"` as the behavior is sort of "masked"

adezxc: Added one more line to the function comment, it's also described in the function.

Didainius: Thanks!

> it's also described in the function

Yep, but as an SDK user (and this is a public function) one does not want to read internal function docs unless there is some real trouble

dataclouder (approved):

lvirbalas (approved):

PR #181, opened 2019-04-16
Introduce race condition tests for VM and vApp

Didainius:
Adding concurrent tests to prove #180.

They are guarded by `concurrent` build tag and will not be executed by default, but should be useful in future once we start to improve on structures.

To run them with Go race detector one can do:
```bash
cd govcd && go test -race -tags "concurrent" -check.vv -check.f Test_VMRefreshConcurrent .
cd govcd && go test -race -tags "concurrent" -check.vv -check.f Test_VAPPRefreshConcurrent .
```

Similar output is expected:
```
START: vapp_concurrent_test.go:17: TestVCD.Test_VAPPRefreshConcurrent
==================
WARNING: DATA RACE
Write at 0x00c0000c2d00 by goroutine 37:
  github.com/vmware/go-vcloud-director/v2/govcd.(*VApp).Refresh()
      /Users/dserplis/Documents/vmware/go-vcloud-director/govcd/vapp.go:96 +0x54f
  github.com/vmware/go-vcloud-director/v2/govcd.(*TestVCD).Test_VAPPRefreshConcurrent.func1()
      /Users/dserplis/Documents/vmware/go-vcloud-director/govcd/vapp_concurrent_test.go:29 +0x38

Previous write at 0x00c0000c2d00 by goroutine 34:
  [failed to restore the stack]

Goroutine 37 (running) created at:
  github.com/vmware/go-vcloud-director/v2/govcd.(*TestVCD).Test_VAPPRefreshConcurrent()
      /Users/dserplis/Documents/vmware/go-vcloud-director/govcd/vapp_concurrent_test.go:28 +0x248
  runtime.call32()
      /usr/local/Cellar/go/1.12.4/libexec/src/runtime/asm_amd64.s:519 +0x3a
  reflect.Value.Call()
      /usr/local/Cellar/go/1.12.4/libexec/src/reflect/value.go:308 +0xc0
  gopkg.in/check%2ev1.(*suiteRunner).forkTest.func1()
      /Users/dserplis/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20180628173108-788fd7840127/check.go:781 +0x978
  gopkg.in/check%2ev1.(*suiteRunner).forkCall.func1()
      /Users/dserplis/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20180628173108-788fd7840127/check.go:675 +0xb7

Goroutine 34 (running) created at:
[...]

Issue #180, opened 2019-04-16
vCD object data structures are not thread-safe


Didainius:
vCD object data structures (like vApp, VM, etc) are not thread-safe and one can have data race problems if operations are run concurrently (in goroutines). At the moment one has to avoid using goroutines or any other concurrent way to access structures.

This needs some bigger thoughts and wider discussion, but from the top of my head there are a few potential solutions:
* Use mutex locks to avoid multiple actions to update data structure at the same time. (And maybe consider to make these objects as singletons)
* Rethinking structure and maybe use value semantics instead of pointer semantics (doesn't seem to be the nicest thing, but noting for further thoughts).

lvirbalas:
Please provide more details on:
- Which data structures?
- What are possible solution?

Didainius:
Sorry. I've added more details to title and issue description rather than getting it lost in the thread.

lvirbalas:
Awesome!
CC @vbauzysvmware

vbauzysvmware:
At moment my mind is main approach would be create singleton for vapp, edge gtw and other required resources and be gatekeeper that only one object could be created and then on funtion level locks could be used to sync multi access.

vbauzysvmware:
I think we don't have this issue anymore after redoing rationalization? @Didainius

Didainius:
This issue is no longer valid as we have decided that the user of this library must understand of what can use parallelism, and what can not. https://github.com/terraform-providers/terraform-provider-vcd is one example how one can deal with parallelism.

PR #587, opened 2023-07-03
Add support for standalone Distributed Firewall Rule management

Didainius:
This PR improves NSX-T Distributed Firewall Rule management.
VCD APIs do not provide a mechanism to create a single firewall rule, only all of them at once. This PR introduces types and methods for managing Distributed Firewall Rules one by one.

The main thing to note is that `VdcGroup.CreateDistributedFirewallRule` still leverages the API endpoint for "update all", but does the work of merging rules, preserving their order, using `json.RawMessage` instead of exact types to prevent altering rules in any way now and in future.


govcd/nsxt_distributed_firewall.go
```diff
@@ -99,3 +109,329 @@ func (firewall *DistributedFirewall) DeleteAllRules() error {
...
+// which handles all distributed firewall (dfw) rules at once. While there is no real endpoint to
+// create single firewall rule, it is a requirements for some cases (e.g. using in Terraform)
+// The code works by doing the following steps:
+//
+// 1. Getting all Distributed Firewall Rules and storing them in private intermediate
+// type`distributedFirewallRulesRaw` which holds a []json.RawMessage (text) instead of exact types.
+// This will prevent altering existing rules in any way (for example if a new field appears in
+// schema in future VCD versions)
+//
+// 2. Converting the give `rule` into json.RawMessage so that it is provided in the same format as
```
lvirbalas, line 182: Typo?

```suggestion
// 2. Converting the given `rule` into json.RawMessage so that it is provided in the same format as
```

govcd/nsxt_distributed_firewall.go
```diff
@@ -99,3 +109,329 @@ func (firewall *DistributedFirewall) DeleteAllRules() error {
...
+	return vdcGroup.GetDistributedFirewallRuleById(oneByName.ID)
+}
+
+// CreateDistributedFirewallRule is a wrapper around "vdcGroups/%s/dfwPolicies/%s/rules" endpoint
+// which handles all distributed firewall (dfw) rules at once. While there is no real endpoint to
+// create single firewall rule, it is a requirements for some cases (e.g. using in Terraform)
+// The code works by doing the following steps:
+//
+// 1. Getting all Distributed Firewall Rules and storing them in private intermediate
+// type`distributedFirewallRulesRaw` which holds a []json.RawMessage (text) instead of exact types.
```
lvirbalas, line 178: An awesome idea!

govcd/nsxt_distributed_firewall.go
```diff
@@ -99,3 +109,329 @@ func (firewall *DistributedFirewall) DeleteAllRules() error {
...
+// other already retrieved rules
+//
+// 3. Creating a new structure of []json.RawMessage which puts the new rule into one of places:
+// 3.1. to the end of []json.RawMessage - bottom of the list
+// 3.2. if `optionalAboveRuleId` argument is specified - identifying the position and placing new
+// rule above it
+// 4. Perform a PUT (update) call to the "vdcGroups/%s/dfwPolicies/%s/rules" endpoint using the
+// newly constructed payload
+//
+// Note. Running this function concurrently will corrupt firewall rules as it uses an endpoint that
```
lvirbalas, line 192: I was waiting to read something like this :) I believe it is very risky to leave concurrency handling to the caller of this function. The cost is too high in such case, as it would lead to data corruption, which we cannot afford. What about synchronizing this function right here, so the callers don't have to?

FYI: @dataclouder @adambarreiro

govcd/nsxt_distributed_firewall_test.go
```diff
@@ -283,3 +283,130 @@ func dumpDistributedFirewallRulesToScreen(rules []*types.DistributedFirewallRule
...
+	testDistributedFirewallRuleSequence(vcd, check, randomizedFwRuleSubSet, vdcGroup, true)
+}
+
+// testDistributedFirewallRuleSequence tests the following:
+// * create firewall rules one one by one
+// * check that the order of firewall rules is the same as requested (or exactly reverse if
+// reverseOrder=true)
+// * check that all IDs of created firewall rules persisted during further updates (means that no
+// firewall rules were recreated during addition of new ones)
+func testDistributedFirewallRuleSequence(vcd *TestVCD, check *C, randomizedFwRuleSubSet []*types.DistributedFirewallRule, vdcGroup *VdcGroup, reverseOrder bool) {
```
lvirbalas, line 362: Would it be possible to add a test to hit parallel creation case with the goal to corrupt data? This ask is coming from my comment above: if we do add lock to the public function, then this test should never corrupt data. In a sense, it's a TDD: if it's doable at all, first you could write the test, hit the data corruption and afterwards add the lock and see that there's no more data corruption happening.

CC: @dataclouder

lvirbalas (changes requested):
This PR is close to magick. For some reason, you have two like that this week already (first was the partial matching of IP ranges in #1062) :) Hence, sharing some thoughts in-line.

govcd/nsxt_distributed_firewall.go
Didainius: I am not sure we want to do an exception here. There are plenty of situations where one can hit such problems (https://github.com/vmware/go-vcloud-director/pull/181)
But back then we agreed to offload it to consumer (Terraform provider in our case)
https://github.com/vmware/go-vcloud-director/issues/180#issuecomment-547390216


OTOH, from the technical side I don't see a big problem making a locking mechanism, but it is just not something that we have done in the SDK (yet)

govcd/nsxt_distributed_firewall.go
```diff
@@ -99,3 +109,329 @@ func (firewall *DistributedFirewall) DeleteAllRules() error {
...
+	oneByName, err := oneOrError("name", name, filteredByName)
+	if err != nil {
+		return nil, err
+	}
+
+	return vdcGroup.GetDistributedFirewallRuleById(oneByName.ID)
+}
+
+// CreateDistributedFirewallRule is a wrapper around "vdcGroups/%s/dfwPolicies/%s/rules" endpoint
+// which handles all distributed firewall (dfw) rules at once. While there is no real endpoint to
```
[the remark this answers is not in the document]
Didainius: Fixed.

govcd/nsxt_distributed_firewall.go
Didainius: Yep, fixed.

govcd/nsxt_distributed_firewall.go
Didainius: Thanks. To be fair this is not a new mechanism - we use the same mechanism for the core OpenAPI support functionality `OpenApiGetAllItems` https://github.com/vmware/go-vcloud-director/blob/main/govcd/openapi.go#L98-L112

govcd/nsxt_distributed_firewall_test.go
Didainius: I can add a test that replicates a problem - the outcome wouldn't be exactly predictable (some rules could make it, others wouldn't - even depending on latency), but the fact is that it wouldn't be as predicted. But OTOH - I am not sure if we want to perform locking itself - https://github.com/vmware/go-vcloud-director/pull/587#discussion_r1252894679

govcd/nsxt_distributed_firewall.go
lvirbalas: After leaving my comment, I started thinking about that as well...

Would you share a link where is the locking for this rule management implemented currently? For some reason, I'm having a hard time quickly locating it. Thanks!

Didainius: Both `vcd_distributed_firewall` and `vcd_distributed_firewall_rule` are locking on parent VDC Group. Only one operation can happen at a time at that particular VDC Group. (There can be only one Distributed Firewall per VDC Group {at least for now} )

* https://github.com/vmware/terraform-provider-vcd/pull/1076/files#diff-fb43a12d1b2bebd9d74f4c764d71fa5d5886715209c589c5e385d1ce12c8ac07R143-R144
* https://github.com/vmware/terraform-provider-vcd/blob/main/vcd/resource_vcd_nsxt_distributed_firewall.go#L154-L156

adambarreiro: I think the note is good enough. Possible suggestions could be

```
// CreateDistributedFirewallRule is a non-thread safe wrapper ...
```

or

```
// WARNING: Running this function ...
```

Just to make the "heads up" more visible.

adambarreiro (approved):
Just some really minor things. Other than that, great PR. The wrapper is really smart.

govcd/nsxt_distributed_firewall.go
lvirbalas: @adambarreiro once you're happy with the note above, please feel free to resolve this conversation, as my questions have been answered here :)

govcd/nsxt_distributed_firewall_test.go
```diff
@@ -283,3 +283,130 @@ func dumpDistributedFirewallRulesToScreen(rules []*types.DistributedFirewallRule
...
+
+	vdc, vdcGroup := test_CreateVdcGroup(check, adminOrg, vcd)
+	check.Assert(vdc, NotNil)
+	check.Assert(vdcGroup, NotNil)
+
+	fmt.Println("# Running Distributed Firewall tests for single Rule")
+	test_NsxtDistributedFirewallRule(vcd, check, vdcGroup.VdcGroup.Id, vcd.client, vdc)
+
+	// Cleanup
+	err = vdcGroup.Delete()
```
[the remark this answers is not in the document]
Didainius: Fixed.

govcd/nsxt_distributed_firewall_test.go
```diff
@@ -283,3 +283,130 @@ func dumpDistributedFirewallRulesToScreen(rules []*types.DistributedFirewallRule
 		util.Logger.Printf("Error while dumping Distributed Firewall rules to screen: %s", err)
 	}
 }
+
+// Test_NsxtDistributedFirewallRule tests the capability of managing FIrewall Rules one by one using
```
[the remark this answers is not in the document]
Didainius: fixed.

govcd/nsxt_distributed_firewall.go
Didainius: > PS: I'd choose the first one, you can get the "warning" that is not thread-safe in a glimpse.

Is it the first one, or second with WARNING?

adambarreiro: I'd go with `// CreateDistributedFirewallRule is a non-thread safe wrapper `

Didainius: > I'd go with // CreateDistributedFirewallRule is a non-thread safe wrapper

Made it so

lvirbalas (approved):

adezxc (approved):
Great implementation!

dataclouder (approved):

PR #590, opened 2023-07-18
Fix catalog creation

dataclouder:
This PR makes sure that the created catalog runs its creation tasks completely, and it is ready to perform operations, such as updating or adding metadata.
This change will be completed in `terraform-provider-vcd` by removing the function `waitForMetadataReadiness`, which will become redundant.


adambarreiro (approved):
Nice! This feels more natural than the workaround in the Terraform provider.

adezxc (approved):

Didainius (approved):

Didainius:
Missing changelog entry

lvirbalas (approved):
I like there are no sleeps in this PR :)

PR #592, opened 2023-07-20
Fix shell scripts

dataclouder:
Make shell scripts shebang consistent


adambarreiro (approved):

adezxc (approved):

Didainius (approved):

scripts/changelog-links.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
lvirbalas, line 1: I'm on a fence with switching to `env` "middle-man". Though it should be more portable, but do we have a case where current approach actually failed? My concern is that there are some articles suggesting `env` has security implications:

https://www.baeldung.com/linux/bash-shebang-lines
https://www.brianstorti.com/rethinking-your-shebang/

dataclouder: We have seen several cases where `/bin/bash` did not exist, or it was a very old version of Bash (which is a bigger security problem than using `env`)

lvirbalas (approved):

PR #593, opened 2023-07-20
Improve logs

dataclouder:
Change the logging ability of detecting binary data, so that the payload sent when we update a vApp template (which is not human-readable) doesn't get logged


adambarreiro (approved):

adezxc (approved):

Didainius (approved):

lvirbalas (approved):
Perfect!

PR #602, opened 2023-07-31
Fix Test_*OrgCreateCatalog

dataclouder:
The creation task may not be available after a recent fix to catalog creation (PR #590). Accessing the task directly may lead to a panic. The check is replaced with a more resilient function call (ResourceComplete)

Also, remove chance of infinite loop from `Test_VmShutdown`


adambarreiro (approved):
Thanks for fixing it!

adezxc (approved):

PR #608, opened 2023-08-17
Improve Test_InsertOrEjectMedia, address gosec issues

Didainius:
This PR improves `Test_InsertOrEjectMedia` which sometimes fails.
The reason is - in rare cases VCD lags to update VM structure after media is ejected. The hardware item still persists. In my experimentation, this could happen once in 6-8hours when the test is run in a loop.

Additionally:
* Addresses new `gosec` reported issues
* Bumps GitHub actions to use Go 1.21 by default


govcd/vm_test.go
```diff
@@ -450,10 +450,32 @@ func (vcd *TestVCD) Test_InsertOrEjectMedia(check *C) {
...
+			err = vm.Refresh()
+			if err != nil {
+				fmt.Printf("error refreshing VM: %s\n", err)
+			}
+			loopisMediaNotEjected := isMediaInjected(vm.VM.VirtualHardwareSection.Item)
+			if loopisMediaNotEjected == false {
+				fmt.Printf("media error recovered at %d refresh attempt\n", a)
+				check.SucceedNow()
+			}
+			time.Sleep(1 * time.Second)
```
dataclouder, line 475:
```suggestion
			time.Sleep(time.Second)
```

govcd/vm_test.go
```diff
@@ -450,10 +450,32 @@ func (vcd *TestVCD) Test_InsertOrEjectMedia(check *C) {
...
+	if firstIsMediaNotEjected {
+		// Run a few more attempts every second to see if the VM finally shows no media being
+		// present in VirtualHardwareItem structure
+		for a := 0; a < retryCount; a++ {
+			fmt.Printf("attempt %d\n", a)
+			err = vm.Refresh()
+			if err != nil {
+				fmt.Printf("error refreshing VM: %s\n", err)
+			}
+			loopisMediaNotEjected := isMediaInjected(vm.VM.VirtualHardwareSection.Item)
```
dataclouder, line 470: `lopisMediaNotEjected` is a confusing name, because it contains `Not` and we compare it with `false` in the next line. 
Perhaps `loopIsMediaInjected` (and `if !loopIsMediaInjected` in the next line) ?

govcd/vm_test.go

govcd/vm_test.go
Didainius: Good point. Made an attempt to change names

lvirbalas (approved):

PR #614, opened 2023-09-04
Use verbose retry function for ALB test cleanup

Didainius:
~This PR attempts to improve on test cleanup for some cases by providing an experimental solution.~

~These cases for now are related to ALB, which sometimes (although very rarely) can fail in an unexpected place. These failures happen in unexpected places and it is not always possible to catch all of them.~

~I have looked at 2 ways of improving cleanup so that other tests are not impacted:~
~* build explicit cleanup code~
~* Reuse already existing cleanup list. This looks to be more universal and has less overhead.~

~The general idea is to build a similar function to `removeLeftoverEntities` which is being triggered at the end of test (unless prevented). A new function `removeLeftoverEntitiesByTestName` is made. The difference is that it accepts `testName` so that all items in the cleanup list can be filtered by test name (put into `createdBy` in `AddToCleanupListOpenApi` function). That way we can be sure that cleanup is performed only on the entities, that were created by that particular test.~

The above approach was tried in [2ba1109](https://github.com/vmware/go-vcloud-director/pull/614/commits/2ba11096d6cd73dd44902cf440e4a490723b253e).

Latest approach is to hit the problem at exact place were it occurred by trying to perform a verbose retry for operation using new function `retryOnError` which is used for testing.

### DIFF
--- a/govcd/api_vcd_test.go
+++ b/govcd/api_vcd_test.go
@@ -747,6 +747,28 @@ func (vcd *TestVCD) getAdminOrgAndVdcFromCleanupEntity(entity CleanupEntity) (or
 	return org, vdc, nil
 }
 
+func (vcd *TestVCD) removeLeftoverEntitiesByTestName(check *C, testName string) {
+	// Gets the persistent cleanup list from file, if exists.
+	cleanupList, err := readCleanupList()
+	if err != nil {
+		check.Logf("failed to read cleanup list: %s", err)
+	}
+	if len(cleanupList) > 0 && err == nil {
+		fmt.Printf("*** Attempting to cleanup entities, created by this test '%s' in file '%s'\n",
+			testName, makePersistentCleanupFileName())
+
+		for i, cleanupEntity := range cleanupList {
+
+			// Ignoring entities, that are not created in the test with a given name
+			if cleanupEntity.CreatedBy != testName {
+				continue
+			}
+			fmt.Printf("# %d ", i+1)
+			vcd.removeLeftoverEntities(cleanupEntity)
+		}
+	}
+}
+
 // Removes leftover entities that may still exist after failed tests
 // or the ones that were explicitly created for several tests and
 // were relying on this procedure to clean up at the end.
--- a/govcd/api_vcd_test.go
+++ b/govcd/api_vcd_test.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"sync"
 	"testing"
+	"time"
 
 	. "gopkg.in/check.v1"
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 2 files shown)


govcd/nsxt_alb_virtual_service_test.go
```diff
@@ -17,6 +17,13 @@ func (vcd *TestVCD) Test_AlbVirtualService(check *C) {
 	skipNoNsxtAlbConfiguration(vcd, check)
 	skipOpenApiEndpointTest(vcd, check, types.OpenApiPathVersion1_0_0+types.OpenApiEndpointAlbEdgeGateway)
 
+	failedAlbCleanup := true
+	defer func() {
+		if failedAlbCleanup {
+			vcd.removeLeftoverEntitiesByTestName(check, check.TestName())
+		}
+	}()
```
Didainius, line 25: By default a run of cleanup for this test is deferred, unless it is disabled at the end of the test

govcd/nsxt_alb_virtual_service_test.go
```diff
@@ -78,6 +85,8 @@ func (vcd *TestVCD) Test_AlbVirtualService(check *C) {
 	// cleanup Org user
 	err = orgUser.Delete(true)
 	check.Assert(err, IsNil)
+
+	failedAlbCleanup = false
```
Didainius, line 89: If the test got here without failures - there is no need to run deferred cleanup

govcd/nsxt_alb_virtual_service_test.go
```diff
@@ -498,6 +507,8 @@ func testAlbVirtualServiceConfig(check *C, vcd *TestVCD, name string, setupConfi
 		check.Assert(updatedPool.NsxtAlbVirtualService.GatewayRef.ID, NotNil)
 	}
 
+	check.Errorf("Testing failure")
```
Didainius, line 510: For the purpose of testing cleanup - a failure is injected

govcd/api_vcd_test.go
```diff
@@ -747,6 +747,28 @@ func (vcd *TestVCD) getAdminOrgAndVdcFromCleanupEntity(entity CleanupEntity) (or
 	return org, vdc, nil
 }
 
+func (vcd *TestVCD) removeLeftoverEntitiesByTestName(check *C, testName string) {
```
Didainius, line 750: Very similar to `removeLeftoverEntities`, except it filters cleanup entities by `testName`

dataclouder (approved):

lvirbalas (approved):
Cool!

adambarreiro (approved):

adezxc (approved):

PR #615, opened 2023-09-08
Improve RDE related methods

adambarreiro:
This PR modifies the following:

- Makes `DefinedEntityType.SetBehaviorAccessControls` more robust to avoid NullPointerException errors in VCD side when the input is a nil slice.


lvirbalas (approved):

adezxc (approved):

dataclouder (changes requested):
Can we add a test to demonstrate the assumption? i.e. getting an array of nil as input

dataclouder (approved):

Didainius (approved):

Issue #552, opened 2023-02-28
Option eject_force do not eject media without an human intervention


dmicheneau:
### Describe the bug

When we try to destroy the resource vcd_inserted_media we need to do a human intervention to answer Yes to force the ejection.
But on Terraform side the intervention seems to be ok
```
vcd_inserted_media.example: Refreshing state... [id=vapp_test3_TestRomain_debian-9.9.0-amd64-netinst.iso]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  - destroy

Terraform will perform the following actions:

  # vcd_inserted_media.example will be destroyed
  - resource "vcd_inserted_media" "example" {
      - catalog     = "my-catalog_romain" -> null
      - eject_force = true -> null
      - id          = "vapp_test3_TestRomain_debian-9.9.0-amd64-netinst.iso" -> null
      - name        = "debian-9.9.0-amd64-netinst.iso" -> null
      - vapp_name   = "vapp_test3" -> null
      - vm_name     = "TestRomain" -> null
    }

Plan: 0 to add, 0 to change, 1 to destroy.

Do you really want to destroy all resources?
  Terraform will destroy all your managed infrastructure, as shown above.
  There is no undo. Only 'yes' will be accepted to confirm.

  Enter a value: yes

vcd_inserted_media.example: Destroying... [id=vapp_test3_TestRomain_debian-9.9.0-amd64-netinst.iso]
vcd_inserted_media.example: Still destroying... [id=vapp_test3_TestRomain_debian-9.9.0-amd64-netinst.iso, 10s elapsed]
vcd_inserted_media.example: Destruction complete after 11s

Destroy complete! Resources: 1 destroyed.
```  
However side UI vCloud Director, you need to make an action:

![image](https://user-images.githubusercontent.com/47741512/221866412-7d72c2b6-490e-4aa6-8732-1854a1e4b05d.png)

And give a answer normally manage by the option eject_force.

### Reproduction steps

1. Make a terraform apply with resource "vcd_inserted_media" "example"
2. Destroy terraform resource "vcd_inserted_media"
3. check result on vCloud Director UI

### Expected behavior

The attribute eject_force (to true by default) bypass the human intervention.

### Additional context

```
terraform version
Terraform v1.3.9
on linux_amd64
+ provider registry.terraform.io/vmware/vcd v3.8.2
```
On Api vcloud director 37.0

Didainius:
[...]

PR #620, opened 2023-09-29
add "expectingAction" status to make sure - in case of a locked media - the vmquestion is recognized and can be "forcefully" automatically answered again.

schories:
## Issue
Simple fix for 1 additional, missing status - causing the automatic answering of vmquestions - which arise in case of "locked" media in vm - and therefore the media eject task to fail.

Very likely also fixes https://github.com/vmware/go-vcloud-director/issues/552

## Description

The "expectingAction" status is also missing from official "task type - status" documentation:
https://developer.vmware.com/apis/1260/doc/types/TaskType.html

Yet, it represents the status of an media eject task 
- in case of a locked virtual drive
- therefore while the vm is waiting for the vmquestion to be answered (by human or WaitTaskCompletion function), whether media shall be forcefully removed anyway

This PR 
- adds the "expectingAction" status 
- to make sure - in case of a locked media - the vmquestion is recognized and can be "forcefully" automatically answered again.

### DIFF
--- a/govcd/ejecttask.go
+++ b/govcd/ejecttask.go
@@ -51,7 +51,7 @@ func (ejectTask *EjectTask) WaitInspectTaskCompletion(isAnswerYes bool, delay ti
 		}
 
 		// If task is not in a waiting status we're done, check if there's an error and return it.
-		if ejectTask.Task.Task.Status != "queued" && ejectTask.Task.Task.Status != "preRunning" && ejectTask.Task.Task.Status != "running" {
+		if ejectTask.Task.Task.Status != "queued" && ejectTask.Task.Task.Status != "preRunning" && ejectTask.Task.Task.Status != "running" && ejectTask.Task.Task.Status != "expectingAction" {
 			if ejectTask.Task.Task.Status == "error" {
 				return fmt.Errorf("task did not complete succesfully: %s", ejectTask.Task.Task.Error.Message)
 			}


dataclouder:
Hi @schories 
Thanks for this contribution.
While the change is small enough, we are always looking to avoid side effects in other operations.
Before accepting it, we will run the full test suite to see whether something else breaks.
Still, the ideal contribution should include a test that demonstrates the need for such a change.
The best solution would be for you to provide a test that reproduces the case. I understand that this is a lot to ask for a one-line change. If you are unable to do so, please at least provide a code snippet that reproduces the issue.

schories:
Hi,

after reading https://github.com/vmware/go-vcloud-director/blob/main/TESTING.md I understood that a VCD is required, to run the existing tests. 

For ejecttask.go currently no ejecttask_test.go is existing. So I would start building those test from scratch - but:

My customer is a tenant of a commercial VCD offering. The tenant VCD access is production level. Obviously I won't run the test suite against that VCD.

What options, besides buying/renting a VCD instance myself on a hyperscaler, do I have?

While I can support to a certain degree and would appreciate releasing additions and fixes to the public, I obviously won't spent serious money on top of that.

Thanks,

Alexander

dataclouder:
@schories 
if you can make a test, we will run it and eventually suggest changes.

PR #622, opened 2023-10-04
Media item upload any file and download

dataclouder:
Replaces PR #621 

Allows the upload of any file as media item.
Allows the download of any media item as a byte stream


govcd/media.go
```diff
@@ -747,3 +747,90 @@ func (vdc *Vdc) QueryAllMedia(mediaName string) ([]*MediaRecord, error) {
...
+		return nil, err
+	}
+
+	downloadUrl, err := url.ParseRequestURI(downloadHref)
+	if err != nil {
+		return nil, fmt.Errorf("error getting download URL: %s", err)
+	}
+
+	request := media.client.NewRequest(map[string]string{}, http.MethodGet, *downloadUrl, nil)
+	resp, err := media.client.Http.Do(request)
```
adezxc, line 822: Should we `defer resp.Body.Close()` here?

govcd/media_test.go
```diff
@@ -57,6 +59,73 @@ func (vcd *TestVCD) Test_DeleteMedia(check *C) {
...
+	check.Assert(err, IsNil)
+	check.Assert(media, NotNil)
+	check.Assert(media.Media.Name, Equals, itemName)
+
+	// Repeat the download a few times. Make sure that a repeated download works as well as the first one
+	for i := 0; i < 2; i++ {
+		// Download the media item from VCD as a byte slice
+		contents, err := media.Download()
+		check.Assert(err, IsNil)
+		check.Assert(contents, Not(Equals), "")
```
adezxc, line 105:
```suggestion
		check.Assert(contents, Not(Equals), []byte{})
```

govcd/media.go
```diff
@@ -747,3 +747,90 @@ func (vdc *Vdc) QueryAllMedia(mediaName string) ([]*MediaRecord, error) {
 	util.Logger.Printf("[TRACE] Found media records by name: %#v \n", mediaResults)
 	return newMediaRecords, nil
 }
+
+// enableDownload prepares a media item for download and returns a download link
+func (media *Media) enableDownload() (string, error) {
```
adezxc, line 752: Is it a toggle switch or an idempotent function? If it's a switch, maybe it could be changed into something like `EnableDownload(enable bool)`? If not, maybe we could name it like `getDownloadLink` or something similar??

dataclouder: It's idempotent. As explained in line 757, when we request a download, a task operation starts, and a file is saved somewhere. There is an ID pointing to that file added to the Media structure.

If we repeat the operation some time later, the same ID is returned.
The best name would probably be "prepareDownload". I have left as it is for two reasons:
1. it is a private function, and won't confuse users;
2. if we decide to implement downloads for vApps and vApp templates, it will be easier to find this code and use it to create something similar. All three use similar calls to prepare/enable the download

govcd/media.go
dataclouder: Indeed we should. Thanks. Added

govcd/media.go
adezxc: Thanks for the explanation!

Didainius (approved):

govcd/media.go
```diff
@@ -747,3 +747,97 @@ func (vdc *Vdc) QueryAllMedia(mediaName string) ([]*MediaRecord, error) {
...
+
+	if downloadHref == "" {
+		return "", fmt.Errorf("no download URL found")
+	}
+
+	return downloadHref, nil
+}
+
+// Download gets the contents of a media item as a byte stream
+func (media *Media) Download() ([]byte, error) {
```
lvirbalas, line 809: Shall we add a note about download size limit here?

adezxc (approved):

govcd/media.go
dataclouder: I've added a note

dataclouder: The limit is that the download sends the file in one go. The server doesn't provide a send-in-chunks header. Thus, the user will require as much memory as the size of the object.

lvirbalas (approved):

adambarreiro (approved):
LGTM! Test passed on 10.5.0

PR #629, opened 2023-11-14
Amend metadata tests to be compatible with latest VCD

adambarreiro:
In the latest VCD version, certain items such as vApps and VMs that are created via a vApp Template come with autogenerated metadata that needs to be considered.

In the metadata tests, there was an assert that expected a hardcoded value of 2, but this number may vary due to the reason stated above, or due to any other feature that may come in the future. In order to make the test more robust and flexible, instead of hardcoding we use a real calculated value that varies depending on the case.

This also fixes the cleanup function, that was removing all metadata, instead of respecting the already existent metadata.

- [x] 10.5.0 `go test -tags metadata -check.vv -timeout 0`
- [x] 10.5.1 `go test -tags metadata -check.vv -timeout 0`

### DIFF
--- a/govcd/metadata_v2_test.go
+++ b/govcd/metadata_v2_test.go
@@ -388,11 +388,11 @@ func (vcd *TestVCD) testMetadataIgnore(resource metadataCompatible, objectType, 
 			check.Assert(singleMetadata.TypedValue.Value, Equals, "bar")
 		}
 
-		// We add a new entry that won't be filtered out
+		// Add a new entry that won't be filtered out
 		err = resource.AddMetadataEntryWithVisibility("test", "bar2", types.MetadataStringValue, types.MetadataReadWriteVisibility, false)
 		check.Assert(err, IsNil)
 
-		// We retrieve all metadata
+		// Retrieve all metadata
 		allMetadata, err := resource.GetMetadata()
 		check.Assert(err, IsNil)
 		check.Assert(allMetadata, NotNil)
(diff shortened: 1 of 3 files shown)


lvirbalas (approved):
Approving this one up-front.

govcd/metadata_v2_test.go
```diff
@@ -364,27 +392,29 @@ func (vcd *TestVCD) testMetadataIgnore(resource metadataCompatible, objectType,
...
+			// If metadata is ignored, there should be an offset of 1 entry (with key "test")
+			check.Assert(len(allMetadata.MetadataEntry), Equals, len(existingMetadata.MetadataEntry)+1)
+			for _, entry := range allMetadata.MetadataEntry {
+				if tt.metadataIsIgnored {
+					check.Assert(entry.Key, Not(Equals), "foo")
+					check.Assert(entry.TypedValue.Value, Not(Equals), "bar")
+				}
+			}
+		} else {
+			// If metadata is NOT ignored, there should be an offset of 2 entry (with key "foo" and "test")
```
dataclouder, line 414:
```suggestion
			// If metadata is NOT ignored, there should be an offset of 2 entries (with key "foo" and "test")
```

dataclouder (approved):

govcd/metadata_v2_test.go
adambarreiro: Oops, thanks for spotting! Fixed.

Didainius (approved):

adezxc (approved):

PR #645, opened 2024-01-26
Add methods to support CRUD operations on CSE Kubernetes clusters

adambarreiro:
## Context

While the generic Runtime Defined Entity approach to create, read, update and delete Kubernetes clusters is completely valid and usable, it causes some challenges to the final users, as the process can be cumbersome at certain stages, and prone to errors.

The idea of this PR is to provide a set of abstractions that simplify the mentioned operations. These abstractions are accomplished by introducing several new public types and methods on these types.

## Description

This PR adds some new types:
- `CseKubernetesCluster` to manage Container Service Extension Kubernetes clusters. 
- `CseClusterSettings`, `CseControlPlaneSettings`, `CseWorkerPoolSettings`, `CseDefaultStorageClassSettings` to configure a Container Service Extension Kubernetes cluster.
- `CseClusterEvent` to define an event inside `CseKubernetesCluster`
- `CseClusterUpdateInput`, `CseControlPlaneUpdateInput`, `CseWorkerPoolUpdateInput` to update an existing Container Service Extension Kubernetes cluster.
- Some other internal/private auxiliary types.

This PR adds some new methods:

- The `Org` type has two new methods `CseCreateKubernetesCluster` and `CseCreateKubernetesClusterAsync` to create Kubernetes clusters in a VCD appliance with Container Service Extension installed.
- The `VCDClient` type has a new method `CseGetKubernetesClusterById` and `Org` has a new method `CseGetKubernetesClustersByName`, both used to retrieve a Container Service Extension Kubernetes cluster. The first one returns a unique cluster, and the `Org` one returns several clusters, as there can be multiple clusters with the same name.
- The new type `CseKubernetesCluster` has several methods:
  - A method `GetKubeconfig` to retrieve the *kubeconfig* of a provisioned Container Service
  Extension Kubernetes cluster
  - A method `Refresh` to refresh the information and properties of an existing Container Service Extension Kubernetes cluster.
  - Methods to update a Container Service Extension Kubernetes cluster: `UpdateWorkerPools`,
  `AddWorkerPools`, `UpdateControlPlane`, `UpgradeCluster`,
  `SetNodeHealthCheck` and `SetAutoRepairOnErrors`. These are like "syntactic sugar" of the more generic/wide `Update` method that can do several updates in one place.
  - A method `GetSupportedUpgrades` to retrieve all the valid TKGm OVAs that a given Container
  Service Extension Kubernetes cluster can use to be upgraded
  - A method `Delete` to delete a cluster

## File layout

- `cse.go` contains the main public CRUD operations described above. The goal of this file is to be concise, to have a clear documentation and provide an instant view of the available operations to the consumers of the SDK.
- `cse_util.go` contains private methods used by the CRUD operations, that may be too complex or reused several times. These are all private and users should not worry about them.
- `cse_internal.go` contains private methods that transform the "public facing types" (described above) to "internal private ones", and manage/consume Go templates with them.
- `cse_yaml.go` contains private methods that manipulate the CAPI YAML that describes a cluster. Essentially, they are used on updates, as this operation is the one that manipulates YAML.
- `cse_type.go` contains all the types used by all the CSE methods, public and private ones.

## Extra changes

- Improves the previous logic to retrieve a created RDE. When you create a RDE, the response is empty and there is a new Task in the headers. The previous one didn't take into consideration this Task, and queried all RDEs to filter them by name, which can be error-prone, to recover the created RDE and return it to the user. The improved way recovers this Task, and with the ID that is set inside it recovers the unique RDE represented by it. This way is way more precise and correct.
- Refactors several methods that were only public and took a public entity such as `VCDClient`, to have also a friendly private version that can take any generic `Client` object.
- Adds missing fields to existing types, that were needed.

### Testing CSE methods

To run the CSE tests, one must set the `TEST_VCD_CSE` environment variable to any non-empty value. The reason is that CSE requires a very unique environment, and the tests take a long time to be executed.

The tests also require a special `govcd_test_config.yaml` file that is almost empty and contains the new `cse` section:
```
provider:
  user: administrator
  password: *****
  url: https://my-vcd.com/api
  sysOrg: System
vcd:
  org: tenant_org
  nsxt:
    vdc: tenant_vdc
logging:
  enabled: true
  logFileName: go-vcloud-director.log
  verboseCleanup: true
cse:
  version: "4.2.0"
  solutionsOrg: "solutions_org"
  storageProfile: "*"
  tenantOrg: "tenant_org"
  tenantVdc: "tenant_vdc"
  routedNetwork: "tenant_net_routed"
  edgeGateway: "tenant_edgegateway"
  ovaCatalog: "tkgm_catalog"
  ovaName: "ubuntu-2004-kube-v1.25.7+vmware.2-tkg.1-8a74b9f12e488c54605b3537acb683bc"
```

The one can run:
```
GOVCD_SKIP_VAPP_CREATION=1;TEST_VCD_CSE=1 go test -tags cse -check.f 'Test_Cse' -check.vv -timeout=0 | tee $HOME/test-cse.log
```

### Testing other minor changes

To test other minor changes, like refactors and RDE minor improvements, the usual procedure can be used.


go.mod
```diff
@@ -9,11 +9,11 @@ require (
 	github.com/kr/pretty v0.2.1
 	github.com/peterhellberg/link v1.1.0
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127
-	gopkg.in/yaml.v2 v2.4.0
+	gopkg.in/yaml.v2 v2.2.2
```
dataclouder, line 12: This requirement can be removed, and replaced everywhere by `sigs.k8s.io/yaml`

.changes/v2.23.0/645-features.md
```diff
@@ -0,0 +1,19 @@
+* Added a new type `CseKubernetesCluster` to manage Container Service Extension Kubernetes clusters [GH-645]
```
dataclouder, line 1:
```suggestion
* Added type `CseKubernetesCluster` to manage Container Service Extension Kubernetes clusters [GH-645]
```
I think we can remove "new " or "a new" from all the statements in this entry

govcd/cse.go
```diff
@@ -0,0 +1,379 @@
...
+// CseCreateKubernetesCluster creates a Kubernetes cluster with the data given as input (CseClusterSettings). If the given
+// timeout is 0, it waits forever for the cluster creation.
+//
+// If the timeout is reached and the cluster is not available (in "provisioned" state), it will return a non-nil CseKubernetesCluster
+// with only the cluster ID and an error. This means that the cluster will be left in VCD in any state, and it can be retrieved afterward
+// with Org.CseGetKubernetesClusterById and the returned ID.
+//
+// If the cluster is created correctly, returns all the available data in CseKubernetesCluster or an error if some of the fields
+// of the created cluster cannot be calculated or retrieved.
+func (org *Org) CseCreateKubernetesCluster(clusterData CseClusterSettings, timeoutMinutes time.Duration) (*CseKubernetesCluster, error) {
```
dataclouder, line 22:
```suggestion
func (org *Org) CseCreateKubernetesCluster(clusterData CseClusterSettings, timeout time.Duration) (*CseKubernetesCluster, error) {
```
If the timeout is of type `time.Duration`, there is no need to indicate the unit of measurement.

govcd/cse.go
```diff
@@ -0,0 +1,379 @@
...
+	}
+
+	if !updated {
+		return fmt.Errorf("could not update the Kubernetes cluster '%s' after %d retries, due to an ETag lock blocking the operations", cluster.ID, maxRetries)
+	}
+
+	return cluster.Refresh()
+}
+
+// Delete deletes a CSE Kubernetes cluster, waiting the specified amount of minutes. If the timeout is reached, this method
```
dataclouder, line 337:
```suggestion
// Delete deletes a CSE Kubernetes cluster, waiting the specified amount of time. If the timeout is reached, this method
```

govcd/cse.go
```diff
@@ -0,0 +1,379 @@
...
+// returns an error, even if the cluster is already marked for deletion.
+func (cluster *CseKubernetesCluster) Delete(timeoutMinutes time.Duration) error {
```
dataclouder, line 339:
```suggestion
func (cluster *CseKubernetesCluster) Delete(timeout time.Duration) error {
```

govcd/cse.go
```diff
@@ -0,0 +1,379 @@
...
+	var elapsed time.Duration
+	start := time.Now()
+	markForDelete := false
+	forceDelete := false
+	for elapsed <= timeoutMinutes*time.Minute || timeoutMinutes == 0 { // If the user specifies timeoutMinutes=0, we wait forever
```
dataclouder, line 344:
```suggestion
	for elapsed <= timeout || timeout == 0 { // If the user specifies timeout=0, we wait forever
```

govcd/cse.go
```diff
@@ -0,0 +1,379 @@
...
+		}
+
+		util.Logger.Printf("[DEBUG] Cluster '%s' is still not deleted, will check again in 10 seconds", cluster.ID)
+		time.Sleep(10 * time.Second)
+		elapsed = time.Since(start)
+	}
+
+	// We give a hint to the user about the deletion process result
+	if markForDelete && forceDelete {
+		return fmt.Errorf("timeout of %v minutes reached, the cluster was successfully marked for deletion but was not removed in time", timeoutMinutes)
```
dataclouder, line 376:
```suggestion
		return fmt.Errorf("timeout of %s reached, the cluster was successfully marked for deletion but was not removed in time", timeout)
```
instances of `time.Deletion` convert to string in a readable way

govcd/cse.go
```diff
@@ -0,0 +1,379 @@
...
+	}
+	return fmt.Errorf("timeout of %v minutes reached, the cluster was not marked for deletion, please try again", timeoutMinutes)
```
dataclouder, line 378:
```suggestion
	return fmt.Errorf("timeout of %s  reached, the cluster was not marked for deletion, please try again", timeout)
```

govcd/cse_type.go
```diff
@@ -0,0 +1,196 @@
+package govcd
+
+import (
+	semver "github.com/hashicorp/go-version"
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+	"time"
+)
+
+// CseKubernetesCluster is a type for managing an existing Kubernetes cluster created by the Container Service Extension (CSE)
```
dataclouder, line 9: Why these types are not in `./types`?
I would expect most of them to be in `./types`, and only  the usual "Outer" type to be in './govcd`

dataclouder (commented):
First pass

govcd/cse_type.go
adambarreiro: The types from this file are "opinionated" and public, for the SDK users to consume them, whereas the `Capvcd` type from `./types` is the one owned/specified/managed by CSE developers (or in other words, it is "in VCD"), so the idea that I had is to put the ones that _we "own"_ in `./govcd`, and the one that _"VCD/CSE owns"_ in `/types`.

Or in other words, most of the types defined here would be "outer" types, but the difference is that here we don't do the typical wrap of "inner type + client", but instead we provide a set of computed properties that resemble what users see in UI.

I also chose to make the inner type private this time (`capvcdType *types.Capvcd`) as there is a lot of information there, hard to get/understand and sometimes outdated/doesn't reflect the actual cluster state (it is reflected accurately in YAML instead). The public properties reflect what users see in UI, they're picked from the actual YAML state in most cases and they should be enough, but we could discuss as maybe making `capvcdType` public would make sense to someone.

cc: @Didainius @lvirbalas

govcd/cse.go
adambarreiro: Good catch, this started as an `timeoutMinutes int` and the name survived among refactors 😄

govcd/cse.go
adambarreiro: Fixed

govcd/cse.go
adambarreiro: Fixed

govcd/cse.go
adambarreiro: Fixed

govcd/cse.go
adambarreiro: Fixed

.changes/v2.23.0/645-features.md
adambarreiro: Fixed

go.mod
adambarreiro: Done, thanks for pointing out

govcd/cse/4.2.0/capiyaml_cluster.tmpl
```diff
@@ -0,0 +1,153 @@
...
+      controllerManager:
+        extraArgs:
+          enable-hostpath-provisioner: "true"
+      dns:
+        imageRepository: "{{.ContainerRegistryUrl}}"
+        imageTag: "{{.DnsVersion}}"
+      etcd:
+        local:
+          imageRepository: "{{.ContainerRegistryUrl}}"
+          imageTag: "{{.EtcdVersion}}"
```
sahithi, line 106: Is Terraform going to provide a map of etcd and coredns versions for a given Kubernetes version? How does the user know these values? Are we storing any map internally that users can refer to?

govcd/cse/4.2.0/rde.tmpl
```diff
@@ -0,0 +1,32 @@
...
+      {{- if .DefaultStorageClassName }}
+      "defaultStorageClassOptions": {
+          "filesystem": "{{.DefaultStorageClassFileSystem}}",
+          "k8sStorageClassName": "{{.DefaultStorageClassName}}",
+          "vcdStorageProfileName": "{{.DefaultStorageClassStorageProfile}}",
+          "useDeleteReclaimPolicy": {{.DefaultStorageClassUseDeleteReclaimPolicy}}
+      },
+      {{- end }}
+      "secure": {
+        "apiToken": "{{.ApiToken}}"
```
sahithi, line 26: If rde template has spec.vcdKe.secure.apiToken, remove secret object from the capiYaml template

govcd/cse_internal_unit_test.go
```diff
@@ -0,0 +1,235 @@
...
+import (
+	semver "github.com/hashicorp/go-version"
+	"os"
+	"reflect"
+	"strings"
+	"testing"
+)
+
+// Test_cseClusterSettingsInternal_generateCapiYamlAsJsonString
+func Test_cseClusterSettingsInternal_generateCapiYamlAsJsonString(t *testing.T) {
```
lvirbalas, line 14: Awesome test! Please add a comment describing more about it.

govcd/cse_util.go
```diff
@@ -0,0 +1,919 @@
...
+		}, nil
+	}
+	return nil, fmt.Errorf("the Container Service Extension version '%s' is not supported", cseVersion.String())
+}
+
+// cseConvertToCseKubernetesClusterType takes a generic RDE that must represent an existing CSE Kubernetes cluster,
+// and transforms it to an equivalent CseKubernetesCluster object that represents the same cluster, but
+// it is easy to explore and consume. If the input RDE is not a CSE Kubernetes cluster, this method
+// will obviously return an error.
+//
```
lvirbalas, line 44: Given how large the function is and the warning below, would be nice to have a bigger description here on of how it does it.

lvirbalas (approved):
Huge! Looking forward to the Terraform part!

govcd/cse.go
```diff
@@ -0,0 +1,379 @@
...
+
+// CseGetKubernetesClustersByName retrieves all the CSE Kubernetes clusters from VCD with the given name that belong to the receiver Organization.
+// Note: The clusters retrieved won't have a valid ETag to perform operations on them. Use VCDClient.CseGetKubernetesClusterById for that instead.
+func (org *Org) CseGetKubernetesClustersByName(cseVersion semver.Version, name string) ([]*CseKubernetesCluster, error) {
+	cseSubcomponents, err := getCseComponentsVersions(cseVersion)
+	if err != nil {
+		return nil, err
+	}
+
+	rdes, err := getRdesByName(org.client, "vmware", "capvcdCluster", cseSubcomponents.CapvcdRdeTypeVersion, name)
```
Didainius, line 88: Just a word of caution. Maybe these should be constants extracted to the top of file.

govcd/cse.go
```diff
@@ -0,0 +1,379 @@
...
+}
+
+// Refresh gets the latest information about the receiver CSE Kubernetes cluster and updates its properties.
+// All cached fields such as the supported OVAs list (from CseKubernetesCluster.GetSupportedUpgrades) are also cleared.
+func (cluster *CseKubernetesCluster) Refresh() error {
+	refreshed, err := getCseKubernetesClusterById(cluster.client, cluster.ID)
+	if err != nil {
+		return fmt.Errorf("failed refreshing the CSE Kubernetes Cluster: %s", err)
+	}
+	*cluster = *refreshed
```
Didainius, line 121: Any particular reason for this notation? IIUC, `cluster = refreshed` should work fine as well

govcd/cse.go
```diff
@@ -0,0 +1,379 @@
...
+		return fmt.Errorf("can't update a Kubernetes cluster that is not in 'provisioned' state, as it is in '%s'", cluster.capvcdType.Status.VcdKe.State)
+	}
+
+	if input.AutoRepairOnErrors != nil {
+		// Since CSE 4.1.1, the AutoRepairOnError toggle can't be modified and is turned off
+		// automatically by the CSE Server.
+		v411, err := semver.NewVersion("4.1.1")
+		if err != nil {
+			return fmt.Errorf("can't update the  'Auto Repair on Errors' flag: %s", err)
+		}
```
Didainius, line 280: This error looks to incorrect. Is it simply an error that would happen if `semver.NewVersion("4.1.1")` couldn't be parsed?

govcd/cse.go
```diff
@@ -0,0 +1,379 @@
...
+		return err
+	}
+	entityContent := map[string]interface{}{}
+	err = json.Unmarshal(marshaledPayload, &entityContent)
+	if err != nil {
+		return err
+	}
+
+	// We do this loop to increase the chances that the Kubernetes cluster is successfully updated, as this method will be
+	// "fighting" with the CSE Server
```
Didainius, line 304: To expand on the comment - so the point is that CSE server itself will perform updates and `etag` keeps on changing, while we must also squeeze in our update, right?

govcd/cse_internal.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	prettyYaml += fmt.Sprintf("%s\n---\n%s", nodePoolYaml, buf.String())
+
+	// We don't use a standard json.Marshal() as the YAML contains special characters that are not encoded properly, such as '<'.
+	buf.Reset()
+	enc := json.NewEncoder(buf)
+	enc.SetEscapeHTML(false)
+	err = enc.Encode(prettyYaml)
+	if err != nil {
+		return "", fmt.Errorf("could not encode the CAPI YAML into a JSON string: %s", err)
```
Didainius, line 143: Does the `sigs.k8s.io/yaml` not help here?

govcd/cse/4.2.0/capiyaml_cluster.tmpl
adambarreiro: The map will be stored in this repo, these versions are stored in the JSON file **govcd/cse/tkg_versions.json**.

When the user specifies a given OVA, we check in code that it is indeed a TKG OVA, and we inspect the internals to retrieve the `VERSION` property from the Product Section data. With this version we inspect the map and get the corresponding versions (if the version is not in the JSON map, we return an error).

That way the user doesn't need to provide these details, they are retrieved automatically from the OVA, as long it is a TKG OVA.

This logic is in the function [getTkgVersionBundleFromVAppTemplate](https://github.com/adambarreiro/go-vcloud-director/blob/new-methods-cse/govcd/cse_util.go#L742)

govcd/cse/4.2.0/capiyaml_cluster.tmpl
```diff
@@ -0,0 +1,153 @@
...
+    namespace: "{{.TargetNamespace}}"
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: capi-user-credentials
+  namespace: {{.TargetNamespace}}
+type: Opaque
+data:
+  username: "{{.UsernameB64}}"
```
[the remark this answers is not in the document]
adambarreiro: Good catch! I think I was putting the API Token twice, one in the RDE JSON and another one in the CAPVCD payload. I'll remove the YAML one.

govcd/cse/4.2.0/rde.tmpl
adambarreiro: Good catch, I will.

govcd/cse/tkg_versions.json
```diff
@@ -0,0 +1,92 @@
+{
```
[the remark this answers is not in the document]
adambarreiro: Users won't need this map (unless they want to check for curiosity), as the map key is retrieved from the input TKG OVA that the user sets as part of cluster creation settings, automatically. Then, with the Key we retrieve the subcomponents versions (etcd, coreDNS, etc).

The users in this case only need to provide a correct TKG OVA (otherwise they get an error on creation)

govcd/cse_yaml.go
```diff
@@ -0,0 +1,412 @@
...
+		if err != nil {
+			return nil, err
+		}
+		result = append(result, mhc)
+	} else {
+		// There is a MachineHealthCheck block
+		if machineHealthCheckEnabled {
+			// We want it, but it is already there, so nothing to do
+			// TODO: What happens in UI if the VCDKEConfig MHC values are changed, does it get reflected in the cluster?
+			//       If that's the case, we might need to update this value always
```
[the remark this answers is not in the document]
adambarreiro: Oops, forgot about it 😅

govcd/system.go
```diff
@@ -1153,6 +1173,38 @@ func QueryAdminOrgVdcStorageProfileByID(vcdCli *VCDClient, id string) (*types.Qu
...
+		return nil, err
+	}
+	return results.Results.AdminOrgVdcStorageProfileRecord, nil
+}
+
+// queryOrgVdcStorageProfilesByVdcId finds all Storage Profiles of a VDC
+func queryOrgVdcStorageProfilesByVdcId(client *Client, vdcId string) ([]*types.QueryResultOrgVdcStorageProfileRecordType, error) {
+	if client.IsSysAdmin {
+		return nil, errors.New("can't query type QueryResultOrgVdcStorageProfileRecordType as System administrator")
+	}
```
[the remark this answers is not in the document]
adambarreiro: I've expanded the `Test_QueryOrgVdcStorageProfileByID` test with this function, I'll play around with it and remove the check if it's too redundant.

There are some similar checks in other similar functions, if I don't get myself into a deep hole I'll also try to simplify.

govcd/cse.go
adambarreiro: To be honest I thought the same but the IDE warned me it wouldn't work. Doing a bit of research the reason looks like this:

- Doing `cluster = refreshed` overrides the "cluster" address with the "refreshed" address, but "cluster" is scoped to the method, so outside of it does nothing, the address goes back to the previous value. I think the reason is, "cluster" is a copy of the address, so changing it will do nothing.

- On the other hand, doing `*cluster = *refreshed` overrides the "cluster" pointed contents with the "cluster" pointed contents, so as the address remains the same (due to reason above), it works.

[Here](https://stackoverflow.com/questions/38105543/reassigning-in-pointer-method-receiver) probably it's explained a bit better than I did 😅

govcd/cse.go
adambarreiro: Yeah, in reality it is a bit dumb, as the version is hardcoded and should never fail, but I added the error for completeness. Will change it to
```
if err != nil {
    return err
}
```

govcd/cse.go
```diff
@@ -0,0 +1,379 @@
...
+	}
+
+	if input.AutoRepairOnErrors != nil {
+		// Since CSE 4.1.1, the AutoRepairOnError toggle can't be modified and is turned off
+		// automatically by the CSE Server.
+		v411, err := semver.NewVersion("4.1.1")
+		if err != nil {
+			return fmt.Errorf("can't update the  'Auto Repair on Errors' flag: %s", err)
+		}
+		if cluster.CseVersion.GreaterThanOrEqual(v411) {
```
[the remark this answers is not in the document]
adambarreiro: Good point. I added a new case to the unit test `Test_getCseComponentsVersions`
```
		{
			name:       "CSE 4.1.1a is equivalent to 4.1.1",
			cseVersion: "4.1.1a",
			want: &cseComponentsVersions{
				VcdKeConfigRdeTypeVersion: "1.1.0",
				CapvcdRdeTypeVersion:      "1.2.0",
				CseInterfaceVersion:       "1.0.0",
			},
			wantErr: false,
		},
```

govcd/cse.go
adambarreiro: Improved the comment a bit.

govcd/cse.go
```diff
@@ -0,0 +1,379 @@
...
+func (cluster *CseKubernetesCluster) GetKubeconfig(refresh bool) (string, error) {
+	if refresh {
+		err := cluster.Refresh()
+		if err != nil {
+			return "", err
+		}
+	}
+
+	if cluster.State != "provisioned" {
+		return "", fmt.Errorf("cannot get a Kubeconfig of a Kubernetes cluster that is not in 'provisioned' state")
```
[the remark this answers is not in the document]
adambarreiro: Done, thanks!

govcd/cse.go
adambarreiro: Done

govcd/cse_util.go
adambarreiro: Done, thanks

govcd/cse_internal_unit_test.go
adambarreiro: Done, thanks

govcd/cse_yaml.go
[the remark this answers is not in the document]
adambarreiro: Removed the comment. Just for the record, what happens when VCDKEConfig Health values are updated is, eventually the cluster YAML gets patched by the CSE Server.
So I removed this TODO without taking any action, as there's the `refresh` option that should pick these changes in the YAML, so better not to be "too smart here" doing the same thing as the CSE Server does.

govcd/cse_internal.go
dataclouder: Dainius is probably thinking of `yaml.YAMLToJSON`

adambarreiro: Yeah, that would be the way to go if we aimed to transform the YAML into JSON and set the (hypothetical) `capiJson` object with it, something like:
```
{
  "apiVersion": "capvcd.vmware.com/v1.1",
  "kind": "CAPVCDCluster",
  "name": "my-cluster",
  "capiJson": {
     -- The nicely JSON object here
  }
}
```
But here is a bit different, as we are "embedding" a one-liner-YAML as a string value for a JSON field:

```
{
  "apiVersion": "capvcd.vmware.com/v1.1",
  "kind": "CAPVCDCluster",
  "name": "my-cluster",
  "capiYaml": "The big YAML chunk with special characters"
}
```

So in our case we must omit that it's a YAML and consider it as a regular string with odd characters that needs to be inserted in a JSON.

govcd/system.go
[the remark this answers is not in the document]
adambarreiro: I did some tests and it won't work. If a System admin tries to query `QtOrgVdcStorageProfile`, the records are nil (not only `OrgVdcStorageProfileRecord` but also `AdminOrgVdcStorageProfileRecord`). But if I query for `QtAdminOrgVdcStorageProfile` I get the desired records (only `AdminOrgVdcStorageProfileRecord` obviously).

A bit odd if you ask me, but it seems this is how it works it seems... (makes sense as other functions like `QueryOrgVdcStorageProfileByID` have the same check)

govcd/cse_type.go
Didainius: > The types from this file are "opinionated" and public, for the SDK users to consume them, whereas the `Capvcd` type from `./types` is the one owned/specified/managed by CSE developers (or in other words, it is "in VCD"), so the idea that I had is to put the ones that _we "own"_ in `./govcd`, and the one that _"VCD/CSE owns"_ in `/types`.

I am good with this approach

govcd/system.go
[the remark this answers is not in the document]
Didainius: Ah, thanks!

govcd/cse.go
```diff
@@ -0,0 +1,392 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	rde, err := createRdeAndGetFromTask(org.client, cseKubernetesClusterVendor, cseKubernetesClusterNamespace, cseSubcomponents.CapvcdRdeTypeVersion,
+		types.DefinedEntity{
+			EntityType: internalSettings.RdeType.ID,
+			Name:       internalSettings.Name,
+			Entity:     payload,
+		}, &TenantContext{
```
Didainius, line 67: No need, but there is `org.getTenantContext()` :)

govcd/cse_internal.go
```diff
@@ -0,0 +1,237 @@
...
+
+// This collection of files contains all the Go Templates and resources required for the Container Service Extension (CSE) methods
+// to work.
+//
+//go:embed cse
+var cseFiles embed.FS
+
+// getUnmarshaledRdePayload gets the unmarshaled JSON payload to create the Runtime Defined Entity that represents
+// a CSE Kubernetes cluster, by using the receiver information. This method uses all the Go Templates stored in cseFiles
+func (clusterSettings *cseClusterSettingsInternal) getUnmarshaledRdePayload() (map[string]interface{}, error) {
```
Didainius, line 21:
```suggestion
func (clusterSettings *cseClusterSettingsInternal) getUnmarshalledRdePayload() (map[string]interface{}, error) {
```

govcd/cse_yaml.go
```diff
@@ -0,0 +1,410 @@
...
+	"strings"
+)
+
+// updateCapiYaml takes a YAML and modifies its Kubernetes Template OVA, its Control plane, its Worker pools
+// and its Node Health Check capabilities, by using the new values provided as input.
+// If some of the values of the input is not provided, it doesn't change them.
+// If none of the values is provided, it just returns the same untouched YAML.
+func (cluster *CseKubernetesCluster) updateCapiYaml(input CseClusterUpdateInput) (string, error) {
+	if cluster == nil {
+		return cluster.capvcdType.Spec.CapiYaml, fmt.Errorf("receiver cluster is nil")
```
Didainius, line 17: Not me, but linter complains - any risk of `capvcdType` being `nil`?

adambarreiro: Good catch, indeed it would cause a really dumb panic

govcd/cse_internal.go
adambarreiro: Fixed

govcd/cse.go
adambarreiro: Changed, more elegant now.

Didainius (approved):

govcd/cse_util.go
```diff
@@ -0,0 +1,949 @@
...
+	capvcd := &types.Capvcd{}
+	err = json.Unmarshal(entityBytes, &capvcd)
+	if err != nil {
+		return nil, fmt.Errorf("could not unmarshal the RDE contents to create a Capvcd instance: %s", err)
+	}
+
+	result := &CseKubernetesCluster{
+		CseClusterSettings: CseClusterSettings{
+			Name:               rde.DefinedEntity.Name,
+			ApiToken:           "******", // We must not return this one, we return the "standard" 6-asterisk value
```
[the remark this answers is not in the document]
adambarreiro: Good point. Added. Now it logs:
```
        ...
        "markForDelete": false,
        "secure": {
          "apiToken": *******"
        }
```

govcd/cse_yaml.go
```diff
@@ -0,0 +1,411 @@
...
+
+	if input.ControlPlane == nil && input.WorkerPools == nil && input.NodeHealthCheck == nil && input.KubernetesTemplateOvaId == nil && input.NewWorkerPools == nil {
+		return cluster.capvcdType.Spec.CapiYaml, nil
+	}
+
+	// The YAML contains multiple documents, so we cannot use a simple yaml.Unmarshal() as this one just gets the first
+	// document it finds.
+	yamlDocs, err := unmarshalMultipleYamlDocuments(cluster.capvcdType.Spec.CapiYaml)
+	if err != nil {
+		return cluster.capvcdType.Spec.CapiYaml, fmt.Errorf("error unmarshaling YAML: %s", err)
```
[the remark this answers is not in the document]
adambarreiro: Fixed (checked other files as well for the same typo)

dataclouder (approved):

PR #647, opened 2024-01-30
add UDF file format reader to support solutions add-on installation

vmanasiev:
## Short Description
An internal package to support installation of  VCD Solution Add-ons

## Description
In this package a reader is implemented that supports extracting files from solutions add-on ISO without unpacking/mounting it advance. 
## Motivation
A new feature is being implemented where the vcd client and terraform provider should support installation of the  VCD solution add-ons. Solution add-ons are packaged as an ISO media. To support larger files and avoid other issues with simple CDROM file format, the .iso file is packaged as  UDF filesystem (specification: http://www.osta.org/specs/pdf/udf260.pdf ).
udf internal package implements partially that specification and enables read. Specifically, the code allows only reading of top level objects within a .ISO file.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,8 +13,11 @@ require (
 )
 
 require (
+	github.com/howeyc/crc16 v0.0.0-20171223171357-2b2a61e366a6
 	github.com/kr/text v0.1.0 // indirect
 	github.com/stretchr/testify v1.5.1 // indirect
+	golang.org/x/exp v0.0.0-20240119083558-1b970713d09a
+	golang.org/x/text v0.14.0
 )
 
 replace (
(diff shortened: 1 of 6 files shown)


dataclouder (approved):

Didainius (approved):

lvirbalas (approved):
Thank you for your contribution. Approving in advance with a few requests below:

@vmanasiev or @dataclouder please add a few things:
1. Amend the PR description to state that this code is extracting only files from the top of the ISO tree.
2. Add top-level comment/description to the .go files (which are missing it) to explain the purpose and expectations of the code of a particular file.

dataclouder:
> Thank you for your contribution. Approving in advance with a few requests below:
> 
> @vmanasiev or @dataclouder please add a few things:
> 
>     1. Amend the PR description to state that this code is extracting only files from the top of the ISO tree.
> 
>     2. Add top-level comment/description to the .go files (which are missing it) to explain the purpose and expectations of the code of a particular file.

I will take care of the code description after merging.

PR #648, opened 2024-02-01
Add IP Space floating IP suggestion method

Didainius:
This PR adds `*VCDClient.GetAllIpSpaceFloatingIpSuggestions` method as well as relevant type `IpSpaceFloatingIpSuggestion`

### DIFF
--- a/govcd/ip_space_allocation.go
+++ b/govcd/ip_space_allocation.go
@@ -283,11 +283,10 @@ func getAllIpSpaceAllocations(client *Client, ipSpaceId string, org *Org, queryP
 }
 
 // GetAllIpSpaceFloatingIpSuggestions suggests IP addresses to use for networking services on Edge
-// Gateway or Provider Gateway. 'gatewayId' filter is required. Based on the specified Gateway, VCD
-// will query all the applicable IP Spaces and suggest some IP addresses which can be utilized to
-// configure the network services on the Gateway. IP Space IP addresses which are allocated but not
-// currently used for any network services are considered. Results can also be filtered by IPV4 or
-// IPV6 IP address types.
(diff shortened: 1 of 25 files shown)


rocknes (approved):
:lgtm:

govcd/ip_space_allocation.go
```diff
@@ -279,3 +281,30 @@ func getAllIpSpaceAllocations(client *Client, ipSpaceId string, org *Org, queryP
 
 	return results, nil
 }
+
+// GetAllIpSpaceFloatingIpSuggestions suggests IP addresses to use for networking services on Edge
+// Gateway or Provider Gateway. 'gatewayId' is mandatory. Based on the specified Gateway, VCD
+// will query all the applicable IP Spaces and suggest some IP addresses which can be utilized to
+// configure the network services on the Gateway. IP Space IP addresses which are are allocated but
+// not currently used for any network services are considered. Results can also be filtered by IPV4
```
lvirbalas, line 289: A typo. I'm also thinking how to make this description feel less "floaty" (i.e. make it more determined) :)

Something like this:

```suggestion
// configure the network services on the Gateway. Allocated IP Space's IP addresses, but
// not currently used for any network services are returned. Results can also be filtered by IPV4
```

lvirbalas (approved):
One ask and looks great!

dataclouder (approved):

govcd/ip_space_allocation.go
Didainius: I have set it to be such. This comes from API docs, but indeed might sound confusing.

adambarreiro (approved):
Thanks!

PR #657, opened 2024-03-22
Vapp network lists

dataclouder:
Add methods to list vApp networks

## Implementation remarks

1. The API to retrieve vApp networks doesn't expose information about Org and VDC. The only ancestors mentioned in the returned data structure are the vApp and the vCenter.
2. Consequently, we can retrieve the networks associated with a vApp, but not the ones belonging to a VDC.
3. Association with an Org is obtained indirectly by running the request as tenant.
4. When running as Org user, the API populates a field named `linkNetworkName`, which points to the network linked to. But when running as System administrator, the API fills instead a field `realNetworkName` , which points to the name of the **network in vCenter**.


lvirbalas (approved):
Thanks!

Didainius (approved):

adambarreiro (approved):

PR #639, opened 2023-12-19
Remove conditional API calls with outdated API version

adambarreiro:
This PR is the same as #636. This bump is needed to avoid failures in the near future versions of VCD, as 35.0 is fairly outdated.

PR #636, opened 2023-12-04
Remove conditional API calls with outdated API version

Didainius:
This PR removes conditional API version elevation for unsupported API versions (`< 37.0`). This SDK support only `37.0` going forward.

Old API calls were identified using `grep -Rin "GetSpecificApiVersionOnCondition(" *`. Only the ones that have versions `< 37.0` are patched to simply use default API versions `client.APIVersion` which is `37.0` now.

```
# grep -Rin "GetSpecificApiVersionOnCondition(" *                               
api_vcd_versions.go:237:func (client *Client) GetSpecificApiVersionOnCondition(vcdApiVersionCondition, wantedApiVersion string) string {
disk.go:88:             vdc.client.GetSpecificApiVersionOnCondition(">= 36.0", "36.0"))
disk.go:170:            disk.client.GetSpecificApiVersionOnCondition(">= 36.0", "36.0"))
disk.go:231:            disk.client.GetSpecificApiVersionOnCondition(">= 36.0", "36.0"))
disk.go:327:            vdc.client.GetSpecificApiVersionOnCondition(">= 36.0", "36.0"))
disk.go:362:            vdc.client.GetSpecificApiVersionOnCondition(">= 36.0", "36.0"))
disk.go:384:            vdc.client.GetSpecificApiVersionOnCondition(">= 36.0", "36.0"))
system.go:742:          "", "error retrieving storage profile: %s", nil, vdcStorageProfile, client.GetSpecificApiVersionOnCondition(">= 35.0", "35.0"))
vapp.go:164:            vAppComposition, vapp.client.GetSpecificApiVersionOnCondition(">=37.1", "37.1"))
vdc.go:1044:            vdc.client.GetSpecificApiVersionOnCondition(">=37.1", "37.1"))
vm.go:74:       _, err := vm.client.ExecuteRequestWithApiVersion(refreshUrl, http.MethodGet, "", "error refreshing VM: %s", nil, vm.VM, vm.client.GetSpecificApiVersionOnCondition(">=37.1", "37.1"))
vm.go:1507:             vm.client.GetSpecificApiVersionOnCondition(">=37.1", "37.1"))
vm.go:1794:             vapp.client.GetSpecificApiVersionOnCondition(">=37.1", "37.1"))
vm.go:1842:             "", "error retrieving vm: %s", nil, newVm.VM, client.GetSpecificApiVersionOnCondition(">=37.1", "37.1"))
[...]

PR #658, opened 2024-03-25
Patch GetSpecificApiVersionOnCondition to respect minimum API version (client.APIVersion)

Didainius:
This is a continuation of #639.
The function `GetSpecificApiVersionOnCondition` did not respect minimum API version. This PR fixes the function so that we don't have errors waiting to happen like it was in #636 case. Also it simplifies API calls that used `GetSpecificApiVersionOnCondition` that are not relevant anymore.

### Note

When `client.APIVersion` is bumped, this function may use newer API version than the constraints defined therefore one must test the code.

### Testing

Tests passed on 10.5.1.1 with only known anomalies:
```
FAIL: org_saml_test.go:112: TestVCD.TestClient_RetrieveRemoteDoc
FAIL: vm_test.go:2047: TestVCD.Test_GetOvfEnvironment
FAIL: nsxt_l2_vpn_tunnel_test.go:16: TestVCD.Test_NsxtL2VpnTunnel
```


adambarreiro (approved):
Thanks, LGTM

dataclouder (approved):

govcd/api_vcd_versions.go
```diff
@@ -231,13 +231,23 @@ func (client *Client) validateAPIVersion() error {
...
-// described in vcdApiVersionCondition
-// f.e. values ">= 32.0", "32.0" returns 32.0 if vCD version is above or 9.7
-func (client *Client) GetSpecificApiVersionOnCondition(vcdApiVersionCondition, wantedApiVersion string) string {
+// GetSpecificApiVersionOnCondition returns default version or wantedApiVersion if it is connected
+// to version described in vcdApiVersionCondition e.g. values ">= 32.0", "32.0" returns 32.0 if vCD
+// version is above or 9.7
+// Note. This function will always respect minimum supported API version which is defined in
+// client.APIVersion. If the wantedApiVersionOrMinimumRequired is less than minimum supported
+// version, this function will return the minimum supported version
+func (client *Client) GetSpecificApiVersionOnCondition(vcdApiVersionCondition, wantedApiVersionOrMinimumRequired string) string {
```
lvirbalas, line 240: Would be useful to have a paragraph explaining the risk involved here (connecting to a newer API version than originally developed against).

Didainius: Added

Didainius:
> Thank you. Please add a note for reference to PR description too (about the risk involved in using this function without retesting on newer VCD versions).

Added

PR #659, opened 2024-03-26
Fix Test_NsxtL2VpnTunnel for VCD 10.5.1.1+

Didainius:
VCD 10.5.1.23400185 returns 6 asterisks for presharedkey and that causes check to fail in `Test_NsxtL2VpnTunnel`. This PR patches test to match these values.
Note Terraform is not affected as it does not set the PreSharedKey value.

Test `Test_NsxtL2VpnTunnel` passed on both 10.5.1.1 and 10.5.1


.changes/v2.24.0/659-noted.md
```diff
@@ -0,0 +1 @@
+* Patched `Test_NsxtL2VpnTunnel` to match PresharedKey of VCD 10.5.1.1+ [GH-659]
```
adambarreiro, line 1: Typo in file name

adambarreiro (approved):
LGTM, thanks for the patch!

.changes/v2.24.0/659-noted.md
Didainius: Fixed.

dataclouder (approved):

lvirbalas (approved):
Thanks!! If you could also add a bit more description to the PR on the context, would be helpful.

Didainius:
> Thanks!! If you could also add a bit more description to the PR on the context, would be helpful.

Improved.

PR #661, opened 2024-03-26
Add new field ActionValue to types. NsxtFirewallRule to support REJECT action 

Didainius:
`types. NsxtFirewallRule` has new field `ActionValue` instead of deprecated `Action` and it additionally supports `REJECT` value in addition to `ALLOW` and `DROP`.


.changes/v2.24.0/661-improvements.md
```diff
@@ -0,0 +1 @@
+* `types.NsxtFirewallRule` adds field `ActionValue` instead of `Action` that is deprecated VCD API. It allows users to use `REJECT` option [GH-661]
```
adambarreiro, line 1: To match past changelog entries:
```suggestion
* Added field `ActionValue` to `types.NsxtFirewallRule` instead of `Action` that is deprecated in VCD API. It allows users to use `REJECT` option [GH-661]
```

adambarreiro (approved):

.changes/v2.24.0/661-improvements.md
Didainius: Fixed.

lvirbalas (approved):
Looks great!

PR #665, opened 2024-04-11
Add code for Solution Landing Zone management [1]

Didainius:
This PR adds support for managing Solution Landing Zones on VCD 10.4.1+

PR #670, opened 2024-04-24
Solution Add-On upload [2]

Didainius:
This PR continues on top of #665 and adds management of Solution Add-Ons within a Solution Landing Zone.

To create a Solution Add-on, one must have a Solution Add-On image. Due to the architecture of Solution Add-Ons it must be present both - locally and and also uploaded to the catalog. 
A local copy is needed to extract the image and pick required metadata - such as icon, EULA, certificate and Add-On type, while a copy in the VCD catalog will be used to create Solution Add-On instances later.

### Testing 

Testing VCD environment is expected to have an image uploaded already to `nsxtBackedCatalogName` and name of it must be specified in `nsxtCatalogAddonDse` variable. The `Test_SolutionAddOn` test will automatically check if the image is present locally within `/test-resources/cache` subdirectory (which is also added to `.gitignore`) - if it is not found - it will download the file from VCD catalog as defined in `nsxtCatalogAddonDse` variable

Tested on 10.4.1 and 10.5.1


govcd/upload.go
```diff
@@ -159,8 +159,9 @@ func newFileUploadRequest(client *Client, requestUrl string, filePart []byte, of
 	rangeExpression := "bytes " + strconv.FormatInt(int64(offset), 10) + "-" + strconv.FormatInt(int64(offset+filePartSize-1), 10) + "/" + strconv.FormatInt(int64(fileSizeToUpload), 10)
 	uploadReq.Header.Set("Content-Range", rangeExpression)
 
-	for key, value := range uploadReq.Header {
-		util.Logger.Printf("[TRACE] Header: %s :%s \n", key, value)
+	sanitizedHeader := util.SanitizedHeader(uploadReq.Header)
```
Didainius, line 162: This prevents dumping access token in plain text

govcd/certificate_management.go
```diff
@@ -295,3 +326,36 @@ func (certificate *Certificate) Delete() error {
...
+	text := reText.FindStringSubmatch(pem)
+	if len(text) < 2 {
+		return "", fmt.Errorf("start marker 'BEGIN CERTIFICATE' or end marker 'END CERTIFICATE' not found in input certificate")
+	}
+	return text[1], nil
+}
+
+// Same returns true if the certificate text matches the text of the provided PEM
+// (without the BEGIN CERTIFICATE and END CERTIFICATE markers)
+func (certificate *Certificate) Same(pem string) (bool, error) {
```
lvirbalas, line 351: For you consideration:

```suggestion
func (certificate *Certificate) IsSame(pem string) (bool, error) {
```

lvirbalas (approved):
One possible tuning in a comment suggestion.

govcd/landing_zone_add_on.go
```diff
@@ -0,0 +1,492 @@
...
+	}
+
+	queryParams := url.Values{}
+	queryParams.Add("filter", fmt.Sprintf("name==%s", name))
+	results, err := vcdClient.GetAllSolutionAddons(queryParams)
+	if err != nil {
+		return nil, fmt.Errorf("error retrieving Solution Add-Ons: %s", err)
+	}
+
+	return oneOrError("name", name, results)
```
adambarreiro, line 243: Dumb question, can names be repeated? (It happened with K8s clusters)
And in that case, shall we return a slice?

adambarreiro (commented):
First scan, still need to deep dive into the PR

govcd/certificate_management.go
```diff
@@ -165,6 +166,36 @@ func (client *Client) GetAllCertificatesFromLibrary(queryParameters url.Values)
 	return getAllCertificateFromLibrary(client, queryParameters, nil)
 }
 
+// FoundCertificateInLibrary searches among all certificates and return the number of certificates
+// with the text that matches the given PEM
+func (client *Client) FoundCertificateInLibrary(pem string) (int, error) {
```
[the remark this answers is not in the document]
Didainius: Hm, can it be `CountMatchingCertificates`?

govcd/landing_zone_add_on.go
Didainius: Not really. The `name` itself is quite stupid in this case and depending on Add-On version (`vmware.ds-1.4.0-23376809`). One can't upload two Add-Ons of the same version as far as I've experimented

govcd/certificate_management.go
Didainius: I like `SameAs` - reads well

lvirbalas: Sounds great!

govcd/certificate_management.go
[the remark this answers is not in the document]
adambarreiro: Sounds good!

govcd/certificate_management.go
Didainius: Changed.

govcd/landing_zone_add_on_test.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func fetchCacheFile(catalog *Catalog, fileName string, check *C) (string, error) {
+	pwd, err := os.Getwd()
+	check.Assert(err, IsNil)
+	cacheDirPath := pwd + "/test-resources/cache"
+	cacheFilePath := cacheDirPath + "/" + fileName
+	printVerbose("# Using '%s' file to cache Solution Add-On\n", cacheFilePath)
+
+	if _, err := os.Stat(cacheFilePath); errors.Is(err, os.ErrNotExist) {
```
dataclouder, line 102: This check will fail if `test-resources/cache` exists, but it is not a directory.

Didainius: Now it checks if it is a file, removes it and creates a directory instead.

govcd/certificate_management.go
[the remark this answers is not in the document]
Didainius: Changed.

dataclouder (approved):

govcd/landing_zone_add_on.go
```diff
@@ -0,0 +1,492 @@
...
+}
+
+func (s *SolutionAddOn) Update(saoCfg *types.SolutionAddOn) (*SolutionAddOn, error) {
+	unmarshalledRdeEntityJson, err := convertAnyToRdeEntity(saoCfg)
+	if err != nil {
+		return nil, err
+	}
+
+	s.DefinedEntity.DefinedEntity.Entity = unmarshalledRdeEntityJson
+	err = s.DefinedEntity.Update(*s.DefinedEntity.DefinedEntity)
```
adambarreiro, line 253: I have one doubt, should this method return just `error` and mutate the receiver `s *SolutionAddOn`?
I'm asking because if I understood correctly, we're doing a mix of mutate receiver + return a brand new `*SolutionAddOn`: This line mutates the receiver (which makes it "dirty" let's say), but at the end we ignore it and return a new object.

Or in other words, the caller of this method would end with a half-mutated receiver and a new object.
Unless I'm missing something, that is 🙂

adambarreiro (approved):
LGTM, just a doubt

govcd/landing_zone_add_on.go
Didainius: Improved

PR #678, opened 2024-05-16
Add Autoscaler to CSE

adambarreiro:
## Overview

This PR automates the steps to deploy an Autoscaler into a Kubernetes cluster stated in this document:
https://www.vmware.com/content/dam/digitalmarketing/vmware/en/pdf/docs/vmw-whitepaper-cluster-auto-scaler.pdf

This procedure affects cluster creation and cluster update operations.

## Description

A new `autoscaler.tmpl` file has been added to the collection. A new configuration block `Autoscaler` has been added to both the `CseWorkerPoolSettings` (settings to create a cluster) and `CseWorkerPoolUpdateInput` (input to update an existing cluster).

When the `Autoscaler` is set during cluster creation on any of the Worker Pools settings, we add the `autoscaler.tmpl` rendered YAML to the final result, and the Worker Pool deployments will have special `metadata` entries, and won't render the`replicas` field (as it conflicts with the autoscaler).

When none of the Worker Pools settings don't have the `Autoscaler` set, we follow the existing workflow: Nothing new is done.

The cluster is then created with or without autoscaling elements present in the cluster, and also the new metadata entries in each Worker Pool.

What about updating a cluster?

In this case, a user can enable or disable the `Autoscaler` in the same way, using `CseWorkerPoolUpdateInput` or creating new Worker Pools with `Autoscaler` set.

In this case, we search for the Autoscaler YAML document, to check whether it is present in the cluster or not. If we need autoscaling capabilities and the YAML is not there (that would mean that the cluster was created without autoscaling), we add the YAML and the required metadata to the worker pools.

If we don't need autoscaling, then we need to check whether it is enabled in the cluster, and disable its replicas if so.

## Tests

### DIFF
--- a/.changes/v2.25.0/674-bug-fixes.md
+++ b/.changes/v2.25.0/674-bug-fixes.md
@@ -1,0 +1,2 @@
+* Fixed a bug that caused CSE Kubernetes cluster creation to fail when the configured Organization VDC Network belongs to
+  a VDC Group [GH-674]
--- a/govcd/cse/4.1/capiyaml_workerpool.tmpl
+++ b/govcd/cse/4.1/capiyaml_workerpool.tmpl
@@ -19,9 +19,15 @@ kind: MachineDeployment
 metadata:
   name: "{{.NodePoolName}}"
   namespace: "{{.TargetNamespace}}"
+  {{- if and (gt .AutoscalerMaxSize -1) (gt .AutoscalerMinSize -1) }}
+  cluster.x-k8s.io/cluster-api-Autoscaler-node-group-max-size: "{{.AutoscalerMaxSize}}"
+  cluster.x-k8s.io/cluster-api-Autoscaler-node-group-min-size: "{{.AutoscalerMinSize}}"
+  {{- end}}
 spec:
   clusterName: "{{.ClusterName}}"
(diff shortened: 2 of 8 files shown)


govcd/cse_test.go
```diff
@@ -222,22 +261,170 @@ func (vcd *TestVCD) Test_Cse(check *C) {
...
 			check.Assert(pool.MachineCount, Equals, 0)
+			check.Assert(pool.Autoscaler, IsNil)
 		} else {
 			check.Assert(pool.MachineCount, Equals, 1)
 		}
 	}
 }
 
+// Test_CseWithAutoscaler tests the autoscaling capabilities in CSE clusters
+func (vcd *TestVCD) Test_CseWithAutoscaler(check *C) {
```
lvirbalas, line 288: As this feature is not trivial, suggesting to extend/add a test to cover mis-configurations, etc. (the "non-happy" path).

adambarreiro: Added `Test_CseValidationErrors`. Actually I found+fixed a tiny validation bug thanks to this new test 🙂

lvirbalas (approved):

dataclouder (approved):

Didainius (approved):
Tests passed

PR #679, opened 2024-05-23
Solution Add-On instance and publishing [3]

Didainius:
This PR is the last in the set of 3 that manages Solution Landing Zones, Add-Ons and their instances.
It adds a layer of types and methods that will help to create Solution Add-Ons, their instances and also allows publish/unpublish instances. 
It is a generalised solution that should be capable of accommodating different Solution Add-Ons, although DSE (Data Solutions Extension) was mainly tested here with a slight check of OSE (Object Storage Extension).

#### Additional things
* File names of previous solution landing zone PRs are shuffled a bit as seeing the whole picture sparked additional thoughts.
* A new convenience method `DefinedEntity.State()` to save on boilerplate in code when checking for `nil` parent structures
In short this allows for such check:
```go
if DefinedEntity.State() != "RESOLVED"
```
  instead of 
```go
if rde != nil || rde.DefinedEntity != nil && rde.DefinedEntity.State != nil && *rde.DefinedEntity.State != "RESOLVED {
...
```

#### Testing
Testing this part of code requires a build that has CSE installed


govcd/solution_addon_instance_publishing.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+)
+
+var addOnInstancePublishBehaviorId = "urn:vcloud:behavior-interface:invoke:vmware:solutions_add_on_instance:1.0.0"
+
+// Publishing manages publish and Unpublish operations are managed in the same API call
```
dataclouder, line 16:
```suggestion
// Publishing manages publish and unpublish operations, which are handled in the same API call.
```

govcd/solution_addon_instance_publishing.go
```diff
@@ -0,0 +1,40 @@
...
+// To unpublish the `scopeAll` has to be `false and `scope` must be empty
```
dataclouder, line 17:
```suggestion
// To unpublish, the `scopeAll` has to be `false and `scope` must be empty
```

dataclouder (approved):

govcd/solution_addon_instance_publishing.go
Didainius: fixed.

govcd/solution_addon_instance_publishing.go
Didainius: fixed.

govcd/solution_addon_instance.go
```diff
@@ -0,0 +1,445 @@
+/*
```
adambarreiro, line 1: nit: There is `solution_add_on` and `solution_addon` in filenames 😇

adambarreiro (approved):

govcd/solution_addon_instance.go
```diff
@@ -0,0 +1,445 @@
...
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"maps"
+	"net/url"
+	"reflect"
+	"strconv"
+	"strings"
+
+	"github.com/davecgh/go-spew/spew"
```
lvirbalas, line 17: I see we have two occurrences of this import and add here the third. Would it be possible to part without this lib? Any 3rd party lib adds a toll on SRP.

CC: @dataclouder @adambarreiro

adambarreiro: Maybe we could simplify logging (I believe it's only used there, for tracing purposes).

Didainius: Simplified logging and used builtin `%#v` instead of additional package. Removed other occurrences and this dependency is no longer here.

lvirbalas (approved):

govcd/solution_addon_instance.go
dataclouder: we have several occurrence of using `pretty.Formatter` in this repo. In some cases also in the main code, not only in testing.

PR #683, opened 2024-06-12
Fix getOpenApiHighestElevatedVersion bug for unsupported API version

Didainius:
There is a bug in core OpenAPI version picking function `getOpenApiHighestElevatedVersion` that could choose an unsupported API version when automatic elevation mechanism `endpointElevatedApiVersions` is used.

This bug can start occurring when API v36.0 is no longer supported.

The bug occurs in OpenAPI handling function `getOpenApiHighestElevatedVersion` that picks client API versions for OpenAPI endpoints. It occurs when both of the following conditions are met:

* Endpoint in question uses automatic elevation mechanism (has at least single entry in `endpointElevatedApiVersions` map)
* Endpoint has one or more elevation records in `endpointElevatedApiVersions`, but all of these entries only elevate to API versions older than  default minimum supported API version (37.0 at this moment)

#### Example 

So an example when the bug occurs is: We have an endpoint `1.0.0/firewallGroups/`. This endpoint leverages automatic elevation mechanism and only has a single record: 
`"36.0", // Adds support for Dynamic Security Groups by deprecating Type field in favor of TypeValue in endpointElevatedApiVersions.`

The function `getOpenApiHighestElevatedVersion` has a bug when evaluating such conditions and it picks the version that is in the `endpointElevatedApiVersions` which is `36.0` in a given example. This is not good and will fail when minimum supported version is `37.0`.

There are many, but example test that surfaces this bug is `Test_NsxtFirewall`

#### Affected tests:
* Test_AlbVirtualService
* Test_NsxtDynamicSecurityGroup
* Test_NsxtEdgeVdcGroup
* Test_NsxtFirewall
* Test_NsxtIpSet
* Test_NsxtNatDnat
* Test_NsxtNatDnatExternalPortPort
* Test_NsxtNatDnatFirewallMatchPriority
* Test_NsxtNatNoDnat
* Test_NsxtNatNoSnat
* Test_NsxtNatPriorityAndFirewallMatch
* Test_NsxtNatReflexive
* Test_NsxtNatSnat
* Test_NsxtOrgVdcNetworkImportedNsxtLogicalSwitch
* Test_NsxtOrgVdcNetworkIsolated
* Test_NsxtOrgVdcNetworkRouted
* Test_NsxtSegmentProfileTemplate
* Test_NsxtVdcGroupForceDelete
* Test_VdcNetworkProfile

### DIFF
--- a/govcd/api_vcd.go
+++ b/govcd/api_vcd.go
@@ -29,6 +29,8 @@ func init() {
 	requestCounter = &counter
 }
 
+var minVcdApiVersion = "37.0" // supported by 10.4+
+
 // VCDClientOption defines signature for customizing VCDClient using
 // functional options pattern.
 type VCDClientOption func(*VCDClient) error
@@ -136,7 +138,6 @@ func (vcdClient *VCDClient) vcdCloudApiAuthorize(user, pass, org string) (*http.
 // NewVCDClient initializes VMware VMware Cloud Director client with reasonable defaults.
 // It accepts functions of type VCDClientOption for adjusting defaults.
 func NewVCDClient(vcdEndpoint url.URL, insecure bool, options ...VCDClientOption) *VCDClient {
-	minVcdApiVersion := "37.0" // supported by 10.4+
 	userDefinedApiVersion := os.Getenv("GOVCD_API_VERSION")
 	if userDefinedApiVersion != "" {
 		_, err := semver.NewVersion(userDefinedApiVersion)
--- a/govcd/openapi_endpoints.go
+++ b/govcd/openapi_endpoints.go
@@ -305,6 +305,9 @@ func (client *Client) getOpenApiHighestElevatedVersion(endpoint string) (string,
 			// highest version found - store it and abort the loop
 			supportedElevatedVersion = elevatedVersion.Original()
 			break
+		} else {
+			util.Logger.Printf("[DEBUG] Skipped Elevated version '%s' for endpoint '%s', Default minimum version '%s'",
+				elevatedVersion.Original(), endpoint, client.APIVersion)
 		}
 
 		util.Logger.Printf("[DEBUG] API version '%s' is not supported by VCD instance for endpoint '%s'",
--- a/govcd/openapi_endpoints.go
+++ b/govcd/openapi_endpoints.go
@@ -298,9 +298,8 @@ func (client *Client) getOpenApiHighestElevatedVersion(endpoint string) (string,
 			elevatedVersion.Original(), endpoint)
 		// Check if maximum VCD API version supported is greater or equal to elevated version
 
-		// if client.APIVCDMaxVersionIs(fmt.Sprintf(">= %s", elevatedVersion.Original())) &&
-		// 	!client.APIClientVersionIs(fmt.Sprintf("> %s", elevatedVersion.Original())) {
-		if client.APIVCDMaxVersionIs(fmt.Sprintf(">= %s", elevatedVersion.Original())) {
+		if client.APIVCDMaxVersionIs(fmt.Sprintf(">= %s", elevatedVersion.Original())) &&
+			!client.APIClientVersionIs(fmt.Sprintf("> %s", elevatedVersion.Original())) {
 			util.Logger.Printf("[DEBUG] Elevated version '%s' is supported by VCD instance for endpoint '%s'",
 				elevatedVersion.Original(), endpoint)
 			// highest version found - store it and abort the loop
--- a/govcd/openapi_endpoints.go
+++ b/govcd/openapi_endpoints.go
@@ -297,6 +297,9 @@ func (client *Client) getOpenApiHighestElevatedVersion(endpoint string) (string,
 		util.Logger.Printf("[DEBUG] Checking if elevated version '%s' is supported by VCD instance for endpoint '%s'",
 			elevatedVersion.Original(), endpoint)
 		// Check if maximum VCD API version supported is greater or equal to elevated version
+
+		// if client.APIVCDMaxVersionIs(fmt.Sprintf(">= %s", elevatedVersion.Original())) &&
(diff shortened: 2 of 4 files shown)


govcd/api_vcd.go
```diff
@@ -29,6 +29,8 @@ func init() {
 	requestCounter = &counter
 }
 
+var minVcdApiVersion = "37.0" // supported by 10.4+
```
Didainius, line 32: Extracting default minimum supported API version definition into a global variable so that it is shared between client initialization `NewVCDClient`, but can also be used in unit tests

govcd/openapi_endpoints.go
```diff
@@ -297,13 +297,16 @@ func (client *Client) getOpenApiHighestElevatedVersion(endpoint string) (string,
 		util.Logger.Printf("[DEBUG] Checking if elevated version '%s' is supported by VCD instance for endpoint '%s'",
 			elevatedVersion.Original(), endpoint)
 		// Check if maximum VCD API version supported is greater or equal to elevated version
-		if client.APIVCDMaxVersionIs(fmt.Sprintf(">= %s", elevatedVersion.Original())) {
+
+		if client.APIVCDMaxVersionIs(fmt.Sprintf(">= %s", elevatedVersion.Original())) &&
+			!client.APIClientVersionIs(fmt.Sprintf("> %s", elevatedVersion.Original())) {
```
Didainius, line 302: The bug itself - there was no check that elevated version candidate is not lower than default minimum client version.

govcd/openapi_endpoints_unit_test.go
```diff
@@ -34,40 +44,60 @@ func TestClient_getOpenApiHighestElevatedVersion(t *testing.T) {
...
+			wantVersion:                 "34.0",
+			wantErr:                     false, // NAT minimum requirement is version 34.0
+			overRideclientMinApiVersion: "34.0",
+		},
+		{
+			name:              "VCD_version_higher_than_elevated_version_entries",
+			supportedVersions: renderSupportedVersions([]string{"37.0", "37.1", "37.2", "37.3", "38.0", "38.1", "39.0"}),
+			endpoint:          types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointFirewallGroups,
+			wantVersion:       minVcdApiVersion,
+			wantErr:           false,
```
Didainius, line 68: An exact case that replicate the bug - `types.OpenApiEndpointFirewallGroups` only has one elevation directive in `endpointElevatedApiVersions`:

```go
types.OpenApiPathVersion1_0_0 + types.OpenApiEndpointFirewallGroups: {
		//"34.0", // Basic minimum required version
		"36.0", // Adds support for Dynamic Security Groups by deprecating `Type` field in favor of `TypeValue`
	},
```

govcd/openapi_endpoints_unit_test.go
```diff
@@ -89,23 +119,35 @@ func TestClient_getOpenApiHighestElevatedVersion(t *testing.T) {
...
+
+			semverWantVersion, err := semver.NewSemver(wantVersion)
+			if err != nil {
+				t.Fatalf("error parsing 'singleElevatedApiVersion': %s", err)
+			}
+
+			if semverWantVersion.LessThan(semverMinVcdApiVersion) {
+				semverMinVcdApiVersionSegments := semverMinVcdApiVersion.Segments()
+				wantVersion = fmt.Sprintf("%d.%d", semverMinVcdApiVersionSegments[0], semverMinVcdApiVersionSegments[1])
+			}
```
Didainius, line 141: Expecting generated test cases to respect `minVcdApiVersion`

govcd/openapi_endpoints_unit_test.go
```diff
@@ -16,12 +18,20 @@ import (
...
+
 	type testCase struct {
 		name              string
 		supportedVersions SupportedVersions
 		endpoint          string
 		wantVersion       string
 		wantErr           bool
+		// overRideclientMinApiVersion is an option to override default expected version that is
+		// defined in global variable`minVcdApiVersion`
+		overRideclientMinApiVersion string
```
adambarreiro, line 34: Small nit
```suggestion
		// overrideClientMinApiVersion is an option to override default expected version that is
		// defined in global variable`minVcdApiVersion`
		overrideClientMinApiVersion string
```

adambarreiro (approved):
LGTM

dataclouder (approved):

govcd/openapi_endpoints_unit_test.go
Didainius: Thanks. Fixed. LOL, that was a failure that made me laugh now :D

govcd/openapi_endpoints_unit_test.go
```diff
@@ -34,40 +44,60 @@ func TestClient_getOpenApiHighestElevatedVersion(t *testing.T) {
...
 
+		wantVersion := minimumRequiredVersion
+		semverWantVersion, err := semver.NewSemver(wantVersion)
+		if err != nil {
+			t.Fatalf("error parsing 'singleElevatedApiVersion': %s", err)
+		}
+
+		if semverWantVersion.LessThan(semverMinVcdApiVersion) {
+			semverMinVcdApiVersionSegments := semverMinVcdApiVersion.Segments()
+			wantVersion = fmt.Sprintf("%d.%d", semverMinVcdApiVersionSegments[0], semverMinVcdApiVersionSegments[1])
```
lvirbalas, line 90: Suggestion for a log message when these exception conditions occur here and below.

lvirbalas (approved):
A suggestion inline to add more logs for this case.

govcd/openapi_endpoints_unit_test.go
Didainius: Good point. This is a test so I added a few additional messages for those who run tests with verbose mode. But I also added extra logging message to the core function as well.

Didainius: Done.

PR #685, opened 2024-06-13
Patch last remaining VDC creation with bug

Didainius:
Continuing #684 and there is one last remaining ocurence of invalid VDC configuration

### DIFF
--- a/govcd/adminvdc_nsxt_test.go
+++ b/govcd/adminvdc_nsxt_test.go
@@ -91,6 +91,7 @@ func (vcd *TestVCD) Test_CreateNsxtOrgVdc(check *C) {
 		if allocationModel == "Flex" {
 			vdcConfiguration.IsElastic = &trueValue
 			vdcConfiguration.IncludeMemoryOverhead = &trueValue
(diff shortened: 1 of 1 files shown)


govcd/adminvdc_nsxt_test.go
```diff
@@ -91,6 +91,7 @@ func (vcd *TestVCD) Test_CreateNsxtOrgVdc(check *C) {
 		if allocationModel == "Flex" {
 			vdcConfiguration.IsElastic = &trueValue
 			vdcConfiguration.IncludeMemoryOverhead = &trueValue
+			vdcConfiguration.ResourceGuaranteedMemory = addrOf(1.00)
```
lvirbalas, line 94: Would be nice to have a comment here why/in which API version it became needed :)

lvirbalas (approved):

dataclouder (approved):

adambarreiro (approved):

govcd/adminvdc_nsxt_test.go
Didainius: Added

PR #686, opened 2024-06-13
Add VDC Template support

adambarreiro:
## Overview

This PR adds support for VDC Templates with CRUD methods, Access level methods and an instantiation method.

## Description

A VDC Template is a System administrator feature that can make VDC creation easier, by defining how the VDCs should look like in terms of configuration. The templates can then be published to tenants, or remain in the System administrator realm if not published. The published tenants can then be instantiated by either Organizations (inside of it, in Libraries) or in the System org.

## Methods

The new file `vcd_template.go` contains several new methods:

* CRUD methods: `VCDClient.CreateVdcTemplate`, `VdcTemplate.Update`, `VCDClient.GetVdcTemplateById`, `VCDClient.GetVdcTemplateByName`, `VdcTemplate.Delete`
* Access level methods: `VdcTemplate.SetAccessControl`, `VdcTemplate.GetAccessControl`
* Instantiation method: `VdcTemplate.InstantiateVdc`
* Query methods: `VDCClient.QueryAdminVdcTemplates`, `Org.QueryVdcTemplates`

## Binding IDs

There's this odd concept from `VMWVdcTemplate.[]ProviderVdcReference.[]Binding` that is worth mentioning, specially when the new methods are consumed from the Terraform Provider

A Binding looks like:

```
type VMWVdcTemplateBinding struct {
	Name  string
	Value *Reference
}
```

To understand it easily, a `Binding.Name` represents a "field" (or an "option") in the Provider VDC configuration of the VDC Template, for example "External network", "Gateway Edge cluster" or "Service Edge cluster". In UI, these options are visible drop-down menus for the users to select. Then, the `Binding.Value` represents the value of that field. In UI, this would be the selected value.

The `Binding.Name` must be generated on our own, and looks like `urn:vcloud:binding:....`.

With this information, we can understand that, given many Provider VDCs in the VDC Template, each one of them share the same `Binding.Name` (the same sections, eg: "External network"), but differ in values.

**Summary:**

`Binding.Name` is an self-generated UUID that replaces a human readable field (eg: "External network") with a generic URN (`urn:vcloud:binding:....`). `Binding.Value` is the value for that field/option.

## Quirks

The XML flavor of the `/admin/extension/vdcTemplate` doesn't work, as it fails to marshal the `VMWVdcTemplate` structure, hence JSON is used. For that reason, all relevant types and fields now have a mapping to JSON.


types/v56/types.go
```diff
@@ -171,17 +171,18 @@ type IPRanges struct {
...
-	AllocatedIPAddresses *IPAddresses    `xml:"AllocatedIpAddresses,omitempty"` // Read-only list of allocated IP addresses in the network.
-	SubAllocations       *SubAllocations `xml:"SubAllocations,omitempty"`       // Read-only list of IP addresses that are sub allocated to edge gateways.
+	IsInherited           bool            `xml:"IsInherited" json:"isInherited,omitempty"`                             // True if the IP scope is inherit from parent network.
+	Gateway               string          `xml:"Gateway,omitempty" json:"gateway,omitempty"`                           // Gateway of the network.
+	Netmask               string          `xml:"Netmask,omitempty" json:"netmask,omitempty"`                           // Network mask.
+	SubnetPrefixLength    string          `xml:"SubnetPrefixLength,omitempty"`                                         // Prefix length (as an string, used everywhere).
+	SubnetPrefixLengthInt *int            `json:"subnetPrefixLength,omitempty"`                                        // Prefix length (as an int, used in VDC Templates).
```
adambarreiro, line 178: Heads up for this ugly workaround. Until now, it seems all endpoints expected `SubnetPrefixLength` to be a `string`. However, VDC Template endpoint expects an `int`

govcd/vdc_template.go
```diff
@@ -0,0 +1,219 @@
...
+	href := fmt.Sprintf("%s/controlAccess", vdcTemplate.VdcTemplate.HREF)
+	_, err := vdcTemplate.client.ExecuteRequest(href, http.MethodGet, types.AnyXMLMime, "error getting access of VDC Template: %s", nil, result)
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
+}
+
+// Instantiate creates a new VDC from the template and returns its ID if the operation finishes successfully.
+func (vdcTemplate *VdcTemplate) Instantiate(vdcName, description, organizationId string) (string, error) {
```
dataclouder, line 185: I have a few concerns about this method:
1. The name suggests that the template is doing something, but not exactly what. Only the comment reveals that it is creating a VDC. Why not calling it `CreateVdcFromTemplate`? Or, since it is a method of `VdcTemplate`, perhaps `CreateVdc`?
2. the method returns a string and error. This is unusual for our creation methods. Either return a task+error ( like many "Async*" creation methods do) or a pointer to the created object. If there are problems with this strategy, they should be explained in a comment.

govcd/vdc_template.go
```diff
@@ -0,0 +1,238 @@
...
+func (vdcTemplate *VdcTemplate) GetAccess() (*types.ControlAccessParams, error) {
+	if !vdcTemplate.client.IsSysAdmin {
+		return nil, fmt.Errorf("functionality requires System Administrator privileges")
+	}
+	if vdcTemplate.VdcTemplate.HREF == "" {
+		return nil, fmt.Errorf("cannot get the Access list for the VDC Template, its HREF is empty")
+	}
+	result := &types.ControlAccessParams{}
+	href := fmt.Sprintf("%s/controlAccess", vdcTemplate.VdcTemplate.HREF)
+	_, err := vdcTemplate.client.ExecuteRequest(href, http.MethodGet, types.AnyXMLMime, "error getting access of VDC Template: %s", nil, result)
```
lvirbalas, line 177: > error getting access of VDC Template

This reads strange. Is there a mistake in phrasing this sentence?

types/v56/types.go
```diff
@@ -657,6 +658,176 @@ type VdcConfiguration struct {
...
+	// NetworkPoolReference must be empty if this element appears in the request.
+	// AutomaticNetworkPoolReference *Reference `xml:"vmext:AutomaticNetworkPoolReference,omitempty"`
+
+	// Defines a gateway and NAT Routed organization VDC network to be created.
+	GatewayConfiguration *VdcTemplateSpecificationGatewayConfiguration `json:"gatewayConfiguration,omitempty"`
+
+	// A set of name of Storage Profiles, with corresponding limit value, that all Provider VDCs must have, and that are selected at the time of VDC Template instantiation.
+	StorageProfile []*VdcStorageProfile `json:"storageProfile,omitempty"`
+
+	// Set to true to indicate if the FLEX vDC is to be elastic. This field can only be set on input for FLEX VDC templates
```
lvirbalas, line 744: Everywhere it's "VCD", but some comments have "vDC" in this file.

lvirbalas (approved):
Couple inline comments and looks good!

govcd/vdc_template.go
adambarreiro: Yeah, I think that log entry and the methods needed a better naming.

Changed to `Access control` and `AccessControl` instead of just `Access` everywhere.

dataclouder (approved):

Didainius (approved):
LGTM tests passed.

types/v56/types.go
```diff
@@ -657,6 +658,176 @@ type VdcConfiguration struct {
...
+	// The quota of VMs that can be created in this VDC. Includes VMs in both vApps and vApp templates, deployed, or otherwise.
+	// Defaults to 0, which specifies an unlimited number.
+	VmQuota int `json:"vmQuota"`
+
+	// Maximum number of network objects that can be deployed in this VDC. Defaults to 0, which means no networks can be deployed.
+	ProvisionedNetworkQuota int `json:"provisionedNetworkQuota"`
+
+	// Indicates that the Provider VDC's automatically-created VXLAN network pool should be used.
+	// NetworkPoolReference must be empty if this element appears in the request.
+	// AutomaticNetworkPoolReference *Reference `xml:"vmext:AutomaticNetworkPoolReference,omitempty"`
```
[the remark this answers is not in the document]
adambarreiro: Indeed, good catch

govcd/vdc_template.go
```diff
@@ -0,0 +1,263 @@
...
+		return nil, fmt.Errorf("functionality requires System Administrator privileges")
+	}
+
+	result := &types.VMWVdcTemplate{}
+
+	resp, err := client.executeJsonRequest(href.String(), method, input, "error when performing a "+method+" for VDC Template: %s")
+	defer closeBody(resp)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
adambarreiro: Fixed

PR #694, opened 2024-07-10
VM IPv6 support

Didainius:
This PR adds support for VM Secondary IP configuration. This secondary configuration applies to IPv6.

#### Testing
* 10.4.0
* 10.5.1
* 10.6.0


govcd/vm_test.go
```diff
@@ -2350,3 +2350,118 @@ func testVmExtraConfig(vcd *TestVCD, label string, vm *VM, check *C, wantPowerOn
...
+	// Populate OrgVDCNetwork
+	net, err := vdc.GetOrgVdcNetworkByName(orgNetworkName, false)
+	if err != nil {
+		return nil, fmt.Errorf("error finding network : %s", err)
+	}
+
+	// Create empty vApp
+	vapp, err := vdc.CreateRawVApp(vappName, "description")
+	if err != nil {
+		return nil, fmt.Errorf("error creating vapp: %s", err)
```
lvirbalas, line 2453: Nit (noticed as a below error uses "vApp"):

```suggestion
		return nil, fmt.Errorf("error creating vApp: %s", err)
```

lvirbalas (approved):
One comment inline.

govcd/vm_test.go

adambarreiro (approved):

PR #696, opened 2024-07-15
Add External Endpoint and API Filter CRUD methods

adambarreiro:
## Description

This PR adds support for External Endpoints and API Filters with the addition of several new methods:

- CRUD operations for External Endpoints: `VCDClient.CreateExternalEndpoint`, `VCDClient.GetAllExternalEndpoints`, `VCDClient.GetExternalEndpoint`, `VCDClient.GetExternalEndpointById`, `ExternalEndpoint.Update` and `ExternalEndpoint.Delete`
- CRUD operations for API Filters: `VCDClient.CreateApiFilter`, `VCDClient.GetAllApiFilters`, `VCDClient.GetApiFilterById`, `ApiFilter.Update` and `ApiFilter.Delete`

PR #697, opened 2024-07-30
Fix tests

adambarreiro:
## Fixes

* Fixes `Test_ApiFilter` introduced in #696 
* Fixes an XML error when updating a VM spec section:
```
error updating VM spec section: API Error: 400: [ 34132-2024-07-30-10-32-54-071--c3eec3d0-5b83-42e1-8740-d276d94841a5 ] HTTP 400 Bad Request
         - Undeclared namespace prefix "ns5" (for attribute "ipAddressingMode")
         at [row,col {unknown-source}]: [34,114]
```
* Modifies `Test_CreateDeleteEdgeGatewayAdvanced` to avoid an odd test failure as VCD returns `Connected: true` when retrieving the Edge Gateway data.


Didainius (approved):
Needs a changelog note

PR #637, opened 2023-12-04
Added necessary cookies to get valid ADFS endpoint for organization.

mnspodrska:
Closes #556.

- Added necessary cookies to get valid ADFS endpoint for organization.
- Tested with saml_auth_adfs and works fine
- Also tested with terraform-provider-vcd 3.10 and works fine

@dataclouder @Didainius

PR #699, opened 2024-08-02
Support customisation of ADFS server lookup cookie

Didainius:
Closes #637

WIP


adambarreiro (approved):
LGTM, I like that is optional rather than hardcoded.
Unit tests passed.

lvirbalas (approved):

PR #706, opened 2024-09-11
ALB Virtual Service HTTP Policies

Didainius:
This PR adds support for ALB Virtual Service HTTP Policies - Request, Response and Security.


govcd/external_network_v2.go
```diff
@@ -215,6 +215,9 @@ func (vcdClient *VCDClient) GetTier0RouterInterfaceByName(externalNetworkId, dis
 	if err != nil {
 		return nil, fmt.Errorf("error getting %s by DisplayName '%s':%s", labelNsxtTier0RouterInterface, displayName, err)
 	}
+	if len(allT0Interfaces) == 0 {
```
Didainius, line 218: Making T0 Interface lookups throw `ErrorEntityNotFound` to conform with standard (and also fix broken datasource test)

lvirbalas (approved):

PR #710, opened 2024-10-01
Add tm Org and VDC support

Didainius:
Changelog contains more details on the implemented structures.
`make testtm` passes.

PR #712, opened 2024-10-08
NSX-T Manager support

Didainius:
WIP

Contains #710.


Didainius:
Will raise separate PRs.

PR #726, opened 2024-12-04
Add Storage Class support

adambarreiro:
This PR adds a new type `StorageClass` and associated methods.
It also refactors `RegionStoragePolicy` methods to use same structure (with `VCDClient`) as other TM types, and removes useless methods.


lvirbalas (approved):

Didainius (approved):
LGTM, both tests still pass

PR #730, opened 2024-12-16
TM Edge Clusters and QoS

Didainius:
This PR adds support for TM Edge Clusters and their QoS settings

### DIFF
--- a/govcd/tm_edge_cluster.go
+++ b/govcd/tm_edge_cluster.go
@@ -60,6 +60,33 @@ func (vcdClient *VCDClient) GetTmEdgeClusterByName(name string) (*TmEdgeCluster,
 	return vcdClient.GetTmEdgeClusterById(singleEntity.TmEdgeCluster.ID)
 }
 
+// GetTmEdgeClusterByName retrieves TM Edge Cluster by Name
+func (vcdClient *VCDClient) GetTmEdgeClusterByNameAndRegionId(name, regionId string) (*TmEdgeCluster, error) {
+	if name == "" {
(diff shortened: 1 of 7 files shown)


govcd/tm_edge_cluster.go
```diff
@@ -0,0 +1,172 @@
...
+const labelTmEdgeClusterSync = "TM Edge Cluster Sync"
+const labelTmEdgeClusterTransportNodeStatus = "TM Edge Cluster Transport Node Status"
+
+// TmEdgeCluster manages read operations for NSX-T Edge Clusters and their QoS settings
+type TmEdgeCluster struct {
+	TmEdgeCluster *types.TmEdgeCluster
+	vcdClient     *VCDClient
+}
+
+// wrap is a hidden helper that facilitates the usage of a generic CRUD function
```
lvirbalas, line 20: This is a bit confusing. By "hidden" you mean private? Do we really need this function? It seems so small and I can't seem to find its usage in this PR.

Didainius: It is everywhere, where we use generic methods. It is required to make generic approach work - https://github.com/vmware/go-vcloud-director/blob/main/govcd/openapi_generic_outer_entities.go#L13-L19

lvirbalas: Ah, truly. Thanks! Friday before holidays effect..

lvirbalas (approved):

adambarreiro (approved):
LGTM! Test passed

PR #733, opened 2025-01-03
TM Patch Content library and tests

Didainius:
Fixes:
* Test_ContentLibraryItemOva (removed TODO)
* Test_ContentLibraryProvider (removed TODO)
* Test_NewRequestWitNotEncodedParamsWithApiVersion (split the test into separate file, as it shouldn't be even triggered here)
* Remved "TODO TM" from `CreateContentLibrary`

### DIFF
--- a/govcd/api_vcd_test.go
+++ b/govcd/api_vcd_test.go
@@ -11,7 +11,6 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
-	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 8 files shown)


govcd/api_vcd_test.go
```diff
@@ -1979,49 +1978,6 @@ func (vcd *TestVCD) findFirstVapp() VApp {
 	return *vapp
 }
 
-// Test_NewRequestWitNotEncodedParamsWithApiVersion verifies that api version override works
-func (vcd *TestVCD) Test_NewRequestWitNotEncodedParamsWithApiVersion(check *C) {
```
Didainius, line 1983: Split the test to separate file so that it does not get triggered with `tm` tag. It is an XML testing function, that shouldn't be together with "testing infrastructure functions"

adambarreiro (approved):

lvirbalas (approved):

PR #735, opened 2025-01-14
TM entity and test improvements

Didainius:
* The test `Test_ContentLibrarySubscribed` tries to create a subscribed content library. It must ensure that the certificate is trusted as otherwise errors would occur.
* Sometimes vCenter might be refreshed too early in tests, although we sleep. Changing sleep approach to waiting for listener state to become CONNECTED
* Fix Test_TmEdgeCluster and change `TmEdgeCluster.Delete` approach because the old doesn't work anymore

--

* Test `Test_TmOrg` uses `CanManageOrgs: false`

### DIFF
--- a/govcd/tm_common_test.go
+++ b/govcd/tm_common_test.go
@@ -53,8 +53,9 @@ func getOrCreateVCenter(vcd *TestVCD, check *C) (*VCenter, func()) {
 	check.Assert(vc, NotNil)
 	PrependToCleanupList(vcCfg.Name, "OpenApiEntityVcenter", check.TestName(), types.OpenApiPathVersion1_0_0+types.OpenApiEndpointVirtualCenters+vc.VSphereVCenter.VcId)
 
-	printVerbose("# Sleeping after vCenter creation\n")
-	time.Sleep(1 * time.Minute) // TODO: TM: Reevaluate need for sleep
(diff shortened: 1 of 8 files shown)


govcd/tm_common_test.go
```diff
@@ -80,6 +81,23 @@ func getOrCreateVCenter(vcd *TestVCD, check *C) (*VCenter, func()) {
...
+		}
+
+		if v.VSphereVCenter.ListenerState == "CONNECTED" {
+			return nil
+		}
+
+		time.Sleep(2 * time.Second)
+	}
+
+	return fmt.Errorf("failed waiting for listener state to become 'CONNECTED', got '%s'", v.VSphereVCenter.ListenerState)
```
lvirbalas, line 98: Suggesting something like this:

```suggestion
	return fmt.Errorf("waiting for listener state to become 'CONNECTED' expired after %d tries (%d seconds), got '%s'", v.VSphereVCenter.ListenerState)
```

adambarreiro (approved):
Thanks for all the fixes!

lvirbalas (approved):
One suggestion and looks great!

govcd/tm_common_test.go
Didainius: Changed

PR #739, opened 2025-01-23
Add Async Provider Gateway creation function

Didainius:
Added `VCDClient.CreateTmProviderGateway` that exposes the creation task of as it is needed in some cases to retrieve ID of incomplete Provider Gateway creation

`go test -tags functional -check.vv -check.f Test_TmProviderGateway -timeout=85m .` passes

Note. Added extra sleep for tests after vCenter becomes CONNECTED (due to stability)

### DIFF
--- a/govcd/tm_common_test.go
+++ b/govcd/tm_common_test.go
@@ -56,6 +56,8 @@ func getOrCreateVCenter(vcd *TestVCD, check *C) (*VCenter, func()) {
 	printVerbose("# Waiting for listener status to become 'CONNECTED'\n")
 	err = waitForListenerStatusConnected(vc)
 	check.Assert(err, IsNil)
+	printVerbose("# Sleeping after vCenter is 'CONNECTED'\n")
+	time.Sleep(4 * time.Second) // TODO: TM: Reevaluate need for sleep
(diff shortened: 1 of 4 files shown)


adambarreiro (approved):

govcd/tm_common_test.go
```diff
@@ -56,6 +56,8 @@ func getOrCreateVCenter(vcd *TestVCD, check *C) (*VCenter, func()) {
 	printVerbose("# Waiting for listener status to become 'CONNECTED'\n")
 	err = waitForListenerStatusConnected(vc)
 	check.Assert(err, IsNil)
+	printVerbose("# Sleeping after vCenter is 'CONNECTED'\n")
+	time.Sleep(4 * time.Second) // TODO: TM: Reevaluate need for sleep
```
lvirbalas, line 60:
```suggestion
	time.Sleep(4 * time.Second) // TODO: TM: Re-evaluate need for sleep
```

lvirbalas (approved):

govcd/tm_common_test.go
Didainius: Fixed

PR #740, opened 2025-01-24
Improve vCenter refresh stability in tests by using runWithRetry

Didainius:
Test improvements only:

This PR enables  retry of vCenter "refresh" and "refresh policies" operations for testing. It also removes some of the "sleep" behavior for tests.

### DIFF
--- a/govcd/tm_common_test.go
+++ b/govcd/tm_common_test.go
@@ -56,19 +56,19 @@ func getOrCreateVCenter(vcd *TestVCD, check *C) (*VCenter, func()) {
 	printVerbose("# Waiting for listener status to become 'CONNECTED'\n")
 	err = waitForListenerStatusConnected(vc)
 	check.Assert(err, IsNil)
-	printVerbose("# Sleeping after vCenter is 'CONNECTED'\n")
-	time.Sleep(4 * time.Second) // TODO: TM: Re-evaluate need for sleep
-	// Refresh connected vCenter to be sure that all artifacts are loaded
-	printVerbose("# Refreshing vCenter %s\n", vc.VSphereVCenter.Url)
-	err = vc.RefreshVcenter()
+
+	// Sometimes the refresh fails with one of 'vCenterEntityBusyRegexp' errors
+	printVerbose("# Attempting vCenter refresh %s\n", vc.VSphereVCenter.Url)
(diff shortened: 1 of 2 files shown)


adambarreiro (approved):

govcd/tm_common_test.go
```diff
@@ -369,3 +371,41 @@ func createTmIpSpace(vcd *TestVCD, region *Region, check *C, nameSuffix, octet3
 		check.Assert(err, IsNil)
 	}
 }
+
+// vCenter task is sometimes unreliable and trying to refresh it immediately after it becomes
+// connected causes a "BUSY_ENTITY" error (which has a few different messages)
+var maximumVcenterRetryTime = 120 * time.Second                                         // The maximum time a single operation will be retried before giving up
+var vCenterEntityBusyRegexp = regexp.MustCompile(`(is currently busy|400|BUSY_ENTITY)`) // Regexp to match entity busy error
```
lvirbalas, line 378: What about pulling this parameter inside the func runWithRetry, so callers don't care?
Right now it is used in two files, which makes it kind-a hidden.

lvirbalas: Just realized the idea behind it. So, maybe instead of pulling it inside, it could be somewhere "higher" - more visible?

lvirbalas (approved):

govcd/tm_common_test.go
Didainius: We are going to use it in tests only so far. It will not be consumed in "production API"

lvirbalas (changes requested):

govcd/tm_common_test.go
```diff
@@ -369,3 +371,41 @@ func createTmIpSpace(vcd *TestVCD, region *Region, check *C, nameSuffix, octet3
...
+
+func runWithRetry(runOperation func() error, errRegexp *regexp.Regexp, duration time.Duration) error {
```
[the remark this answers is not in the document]
Didainius: As per agreement - migrating to https://github.com/vmware/go-vcloud-director/pull/756

Didainius:
As per discussion https://github.com/vmware/go-vcloud-director/pull/740#discussion_r1930328803 migrated to #756

PR #732, opened 2024-12-20
Adjust LDAP for TM

adambarreiro:
* Created new type `TmLdapSettings` to manage LDAP settings for the Tenant Manager

* Added methods `VCDClient.TmLdapConfigure`, `TmOrg.LdapConfigure`, `VCDClient.TmLdapDisable`, `TmOrg.LdapDisable`,
  `VCDClient.TmGetLdapConfiguration`, `TmOrg.GetLdapConfiguration` to configure LDAP in Tenant Manager and regular Organizations

PR #746, opened 2025-01-31
Add SSL validation and connection test to LDAP configuration methods

adambarreiro:
While PR #732 introduced basic support for LDAP in VMware Cloud Foundation Automation, the implemented methods had a few issues:
- If the LDAP used SSL, the certificate was not trusted, hence the LDAP was configured but useless as it would always fail.
- If LDAP host or port is wrong, the LDAP would be configured but useless as it would always fail.

This PR enhances all implemented methods:

- They check connection against provided hostname and port when requesting setting LDAP. When the LDAP settings are being deleted, there's no check, obviously.
- When SSL is requested in the LDAP settings input parameter, and if `trustCertificate` flag is `true`, the methods now automatically trust the certificate if needed. Otherwise (`trustCertificate=false`), the methods will fail.
- If the SSL certificate is trusted, but configuring the LDAP fails, the certificate is automatically deleted.

Additionally, instead of using dummy vCenter as LDAP server, the testing configuration allows to provide a customised LDAP server:

```yaml
tm:
  ...
  ldap:
    host: my.ldap.server.net
    port: 636
    isSsl: true
    username: admin
    password: StrongPassword
    baseDistinguishedName: OU=demo,DC=foo,DC=bar,DC=test,DC=net
    type: ACTIVE_DIRECTORY
```

Last, it also adds support of **Trusted Certificates for Tenants** (`TmOrg`), which is required for configuring LDAP with SSL on tenants.


govcd/tm_ldap.go
```diff
@@ -15,40 +15,138 @@ import (
...
+	if trustSslCertificate && connResult.TargetProbe.SSLResult == types.UntrustedCertificate {
+		// We retrieve existing LDAP configuration to check if the given input settings are for a fresh connection, or to override
+		// an existing one
+		existing, err := vcdClient.TmGetLdapConfiguration()
+		if err != nil {
+			return nil, fmt.Errorf("error retrieving existing LDAP configuration: %s", err)
+		}
+
+		// If SSL is configured and retrieved settings are empty, or the hostname changed, we need to trust the certificate
+		needsCheckCert := settings.IsSsl && (existing.HostName == "" || existing.HostName != settings.HostName)
```
adambarreiro, line 51: As far as I saw, if you change the host in UI, it doesn't re-trigger the certificate trusting popup. I think it should? So not sure if this should be so "smart" or simplify

Didainius (approved):

PR #769, opened 2025-03-05
Add Async methods for IP Space and Region

Didainius:
This PR adds `VCDClient.CreateTmIpSpaceAsync` and `VCDClient.CreateRegionAsync` to enable asynchronous task tracking

### DIFF
--- a/govcd/tm_region.go
+++ b/govcd/tm_region.go
@@ -33,6 +33,16 @@ func (vcdClient *VCDClient) CreateRegion(config *types.Region) (*Region, error) 
 	return createOuterEntity(&vcdClient.Client, outerType, c, config)
 }
 
+// CreateRegionAsync creates a new region and returns its tracking task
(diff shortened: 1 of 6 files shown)


adambarreiro (approved):

lvirbalas (approved):

govcd/tm_region.go
```diff
@@ -33,6 +33,16 @@ func (vcdClient *VCDClient) CreateRegion(config *types.Region) (*Region, error)
 	return createOuterEntity(&vcdClient.Client, outerType, c, config)
 }
 
+// CreateRegionAsync creates a new region and returns its tracking task
+func (vcdClient *VCDClient) CreateRegionAsync(config *types.Region) (*Task, error) {
```
lvirbalas, line 37: Just one question: what for do we use the previous CreateRegion(...) function after this change?

Didainius: We don't directly use it in here anymore.

PR #774, opened 2025-03-13
Improve LDAP

adambarreiro:
This PR makes LDAP payloads to ignore XML elements that should not be sent. This is an attempt to fix the `java.lang.NullPointerException` that is obtained rarely when disabling Org LDAP, but can't be reproduced with current tests (at least I didn't hit it even once) so **I can't be sure** that this will fix the issue.

### DIFF
--- a/govcd/tm_ldap.go
+++ b/govcd/tm_ldap.go
@@ -238,7 +238,7 @@ func ldapExecuteRequest(vcdClient *VCDClient, orgId, method string, payload inte
 	req := vcdClient.Client.newRequest(nil, nil, method, *endpoint, body, "", headers)
 	resp, err := checkJsonResp(vcdClient.Client.Http.Do(req))
 	if err != nil {
-		return nil, fmt.Errorf("error performing %s to LDAP settings: %s", method, err)
+		return nil, fmt.Errorf("error performing %s call to LDAP settings: %s", method, err)
 	}
 
 	// Other than DELETE with get a body response
--- a/govcd/tm_ldap.go
+++ b/govcd/tm_ldap.go
@@ -5,7 +5,6 @@
 package govcd
 
 import (
-	"bytes"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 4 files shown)


govcd/tm_ldap.go
```diff
@@ -218,13 +217,11 @@ func ldapExecuteRequest(vcdClient *VCDClient, orgId, method string, payload inte
 	// If the call is PUT, we prepare the body with the input settings
 	var body io.Reader
 	if method == http.MethodPut {
-		text := bytes.Buffer{}
-		encoder := json.NewEncoder(&text)
-		err = encoder.Encode(payload)
+		text, err := json.Marshal(payload)
```
adambarreiro, line 220: This was an attempt to make it work as I wanted. Didn't make it, but the code is simpler so I left it

types/v56/types.go
```diff
@@ -1110,8 +1110,8 @@ type VAppLeaseSettings struct {
 // Description: Represents the ldap settings of a VMware Cloud Director organization.
 // Since: 0.9
 type OrgLdapSettingsType struct {
-	XMLName xml.Name `xml:"OrgLdapSettings"`
-	Xmlns   string   `xml:"xmlns,attr,omitempty"`
+	XMLName xml.Name `xml:"OrgLdapSettings" json:"-"`
```
adambarreiro, line 1113: Not sure why JSON marshaler is picking untagged elements, but seems that forcing with `-` works

Didainius (approved):

types/v56/types.go
lvirbalas: Let's add a comment with a link to
https://pkg.go.dev/encoding/json
for the sake of clarity?

lvirbalas (approved):
One suggestion inline.

types/v56/types.go
adambarreiro: Added

PR #776, opened 2025-03-19
Add feature flag management SDK

Didainius:
Add `*VCDClient` methods `UpdateFeatureFlag`, `GetFeatureFlagById`, `GetAllFeatureFlags`

### DIFF
--- a/govcd/feture_flag.go
+++ b/govcd/feture_flag.go
@@ -10,8 +10,11 @@ import (
 	"github.com/vmware/go-vcloud-director/v3/types/v56"
 )
 
-const labelFeatureFlag = "Org VDC Network Segment Profile"
+const labelFeatureFlag = "Feature Flag"
 
+// UpdateFeatureFlag can update a feature flag with a given config. Can mostly be used to enable
+// specific feature flags.
+// The ID must be set (e.g. 'urn:vcloud:featureflag:CLASSIC_TENANT_CREATION' for Classic Tenants)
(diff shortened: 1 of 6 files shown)


govcd/feture_flag.go
```diff
@@ -0,0 +1,47 @@
...
+
+package govcd
+
+import (
+	"fmt"
+
+	"github.com/vmware/go-vcloud-director/v3/types/v56"
+)
+
+const labelFeatureFlag = "Org VDC Network Segment Profile"
```
adambarreiro, line 13:
```suggestion
const labelFeatureFlag = "Feature Flag"
```

adambarreiro (approved):

govcd/feture_flag.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (vcdClient *VCDClient) UpdateFeatureFlag(entityConfig *types.FeatureFlag) (*types.FeatureFlag, error) {
```
lvirbalas, line 15: Would be nice to have a comment explaining a bit more about the feature flag, as it is kind of perpendicular to other things we have.

lvirbalas (approved):

govcd/feture_flag.go
Didainius: Thanks!.. Copy paste... Fixed.

govcd/feture_flag.go
Didainius: Added

PR #777, opened 2025-04-08
Fix licensing and enable validation

Didainius:
This PR adds license to all .go files and adds a `make licensecheck` command that will automatically validate if there are .go files without license.
It is also added to github action so that every PR is validated.


adambarreiro (approved):

lvirbalas (approved):
Thanks!

Issue #803, opened 2026-05-04
Goroutine leak: response body never closed on the error path


sircthulhu:
## Description

`govcd/api.go:checkRespWithErrType` returns `(nil, ParseErr(bodyType, resp, errType))` on every 4xx / 5xx response, so the caller has no handle on the response and cannot close `resp.Body`. `ParseErr` itself forwards the response to `decodeBody`, which `io.ReadAll`s the body but does not close it. Every failed VCD API call leaves one `net/http.setRequestCancel.func4` goroutine alive forever, holding the underlying TCP connection.

## Versions

- go-vcloud-director: `v3.0.0` (reproduces on `main` as well)
- Go: 1.26
- VCD: 10.5

## How it surfaces

Observed in a Kubernetes operator that retries failed VCD calls in a reconcile loop. The pod accumulated about 28 leaked goroutines per second of error rate; one pprof snapshot showed **12 651 goroutines in `net/http.setRequestCancel.func4`** alone (~200 MB RSS) before the pod was OOMKilled.

Stack of every leaked goroutine:

```
net/http.setRequestCancel.func4
  → net/http.(*Transport).roundTrip
  → govcd.executeRequestCustomErr
  → govcd.GetAdminOrgByName / govcd.CreateOrgVdc / …
```

## Steps to reproduce

1. Repeatedly call any govcd method that ends in `executeRequestCustomErr` against an endpoint that returns 4xx (e.g. create an Organization that already exists).
2. Sample `/debug/pprof/goroutine` — every retry leaves one extra `setRequestCancel.func4` goroutine.

## Fix / workaround

Defer `resp.Body.Close()` at the entry of `ParseErr` (last point with access to the response on the error path). PR forthcoming.

frodenas:
Fixed at #804

PR #804, opened 2026-05-04
fix: close response body in ParseErr to prevent goroutine leak

sircthulhu:
## Description

Related issue: #803

Three response-body leaks in govcd that all show up under sustained 4xx / 5xx loads:

1. **`ParseErr` (XML / JSON error path)** — `checkRespWithErrType` returns `(nil, ParseErr(bodyType, resp, errType))`, so the caller has no handle on the response and cannot close it. `ParseErr` itself only forwards the response to `decodeBody`, which `io.ReadAll`s the body but never closes it.
2. **`executeJsonRequest` (OpenAPI / JSON path)** — on every non-2xx the function reads the body with `io.ReadAll` and returns `(resp, err)`. All callers follow `if err != nil { return nil, err }` and only `defer closeBody(resp)` afterwards, so the deferred close never runs.
3. **`rewrapRespBodyNoopCloser`** — reads the body and replaces `resp.Body` with `io.NopCloser(...)`. After that any subsequent `resp.Body.Close()` is a no-op, so the underlying TCP connection is never released. Used from `api_vcd.go`, `entity_client.go`, and `openapi.go` — every request that goes through these paths leaks a goroutine.

Same class of leak (`net/http.setRequestCancel.func4` goroutine pinned to a never-closed body), three different code paths.

## Verification

A real-world Kubernetes operator built on top of go-vcloud-director was leaking about 28 goroutines per second of error rate; one snapshot showed 12 651 leaked `setRequestCancel.func4` goroutines (~200 MB RSS) before the pod OOMed. With all three fixes applied the same workload's goroutine count stays flat at the controller-runtime baseline (~600) and RSS hovers under 100 MB indefinitely.

### DIFF
--- a/govcd/api.go
+++ b/govcd/api.go
@@ -307,6 +307,15 @@ func (client *Client) NewRequestWithApiVersion(params map[string]string, method 
 // ParseErr takes an error XML resp, error interface for unmarshalling and returns a single string for
 // use in error messages.
 func ParseErr(bodyType types.BodyType, resp *http.Response, errType error) error {
+	// Close the response body so the underlying TCP connection and the
+	// internal net/http.setRequestCancel goroutine that watches it can be
+	// released. checkRespWithErrType returns (nil, ParseErr(...)) on the
+	// error path, so the caller no longer has access to the response and
+	// cannot close it; this is the last opportunity to do so.
+	if resp != nil && resp.Body != nil {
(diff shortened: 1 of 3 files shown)


frodenas (approved):

frodenas:
Thanks @sircthulhu

PR #818, opened 2026-07-23
Add CRUD support for user-defined NSX-T Network Context Profiles

microwaves:
The SDK has had lookup functions for Network Context Profiles since #452, but no way to create the user-defined profiles the API supports since VCD 10.2 (API 35.0). This adds the missing write path so that clients (most immediately the terraform provider, vmware/terraform-provider-vcd#1425) can manage custom profiles instead of asking users to script raw API calls.

Changes:

- `NsxtNetworkContextProfile` outer type with `VCDClient.CreateNetworkContextProfile`, `VCDClient.GetNetworkContextProfileById`, `Update` and `Delete`, implemented with the generic CRUD helpers per CODING_GUIDELINES. The existing package-level lookup functions are untouched.
- Trailing slash on `OpenApiEndpointNetworkContextProfiles` so by-id URLs compose like every other CRUD endpoint. The endpoint was already registered at API 35.0, no version changes needed.
- Types: `ContextEntityID` and `NetworkProviderScope` change from `interface{}` to `string`, and `SubAttributes` gets a concrete `NsxtNetworkContextProfileSubAttribute` type. Without typed fields there is no reasonable way to build create payloads. Flagged as breaking in `.changes`, happy to soften if you prefer keeping the interfaces on this line.
- Functional test `Test_NsxtNetworkContextProfileCRUD` covering create, get by id, lookup through the existing by-name function, update with TLS sub-attributes, delete. It sticks to `APP_ID` attributes because `DOMAIN_NAME` values are validated against the NSX FQDN catalog and would make the test environment-dependent.

Verified against VCD 10.6.1 (API 39.0) with a TENANT scoped profile on a VDC Group: full create/read/update/delete cycle, including `APP_ID` with `TLS_VERSION` sub-attributes, plus the documented NSX-side validations (FQDN catalog membership, single-value rule for ALG App IDs). `make static vet testunit tagverify licensecheck` all pass.

If this is something you would take on `release/v2.x` as well, I can open the backport, the patch applies cleanly. Related feature request: vmware/terraform-provider-vcd#1426.

### DIFF
--- a/govcd/nsxt_network_context_profile.go
+++ b/govcd/nsxt_network_context_profile.go
@@ -11,6 +11,73 @@ import (
 	"github.com/vmware/go-vcloud-director/v3/types/v56"
 )
 
+const labelNetworkContextProfile = "NSX-T Network Context Profile"
+
+// NsxtNetworkContextProfile contains a structure for managing user-defined NSX-T Network Context
+// Profiles. SYSTEM scoped profiles are built-in and read-only; profiles with PROVIDER and TENANT
+// scopes can be managed by this structure
+type NsxtNetworkContextProfile struct {
+	NsxtNetworkContextProfile *types.NsxtNetworkContextProfile
+	VCDClient                 *VCDClient
+}
+
+// wrap is a hidden helper that facilitates the usage of a generic CRUD function
+//
+//lint:ignore U1000 this method is used in generic functions, but annoys staticcheck
+func (n NsxtNetworkContextProfile) wrap(inner *types.NsxtNetworkContextProfile) *NsxtNetworkContextProfile {
+	n.NsxtNetworkContextProfile = inner
+	return &n
+}
+
+// CreateNetworkContextProfile creates a user-defined Network Context Profile that can be
+// referenced in Distributed Firewall rules.
+//
+// Profiles with TENANT scope require both OrgRef and ContextEntityID (an Org VDC or VDC Group ID)
+// to be set. Profiles with PROVIDER scope are visible to all tenants and require System
+// administrator privileges
+func (vcdClient *VCDClient) CreateNetworkContextProfile(config *types.NsxtNetworkContextProfile) (*NsxtNetworkContextProfile, error) {
+	c := crudConfig{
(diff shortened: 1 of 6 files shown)


govcd/nsxt_network_context_profile.go
```diff
@@ -11,6 +11,73 @@ import (
...
 )
 
+const labelNetworkContextProfile = "NSX-T Network Context Profile"
+
+// NsxtNetworkContextProfile contains a structure for managing user-defined NSX-T Network Context
+// Profiles. SYSTEM scoped profiles are built-in and read-only; profiles with PROVIDER and TENANT
+// scopes can be managed by this structure
+type NsxtNetworkContextProfile struct {
+	NsxtNetworkContextProfile *types.NsxtNetworkContextProfile
+	VCDClient                 *VCDClient
```
frodenas: Please make it unexported: `vcdClient *VCDClient`

govcd/nsxt_network_context_profile_test.go
```diff
@@ -73,3 +73,83 @@ func filteredTestGetAllNetworkContextProfiles(queryParams url.Values, client *Cl
...
+	check.Assert(retrievedProfile.NsxtNetworkContextProfile.Scope, Equals, "TENANT")
+	check.Assert(len(retrievedProfile.NsxtNetworkContextProfile.Attributes), Equals, 1)
+
+	// Lookup with the pre-existing name based function must find the same entity
+	byName, err := GetNetworkContextProfilesByNameScopeAndContext(&vcd.client.Client, config.Name, "TENANT", vdcGroup.VdcGroup.Id)
+	check.Assert(err, IsNil)
+	check.Assert(byName.ID, Equals, createdProfile.NsxtNetworkContextProfile.ID)
+
+	// Update description and attribute values
+	updateConfig := &types.NsxtNetworkContextProfile{
```
frodenas: Can't you just mutate `config` directly. It make it more obvious which fields the update actually changes vs. which are just carried over.

```
config.Description += "-updated"
config.Attributes = [...]
```

frodenas (commented):
@microwaves Apologies for the late review. Added a few comments.

Issue #820, opened 2026-09-08
`uploadFile` reports `EOF` when the file size is an exact multiple of `uploadPieceSize`


cfiehe:
### Describe the bug

A file upload that completes successfully is reported as failed whenever the file size is an exact multiple of the piece size. Every piece reaches the server and `fileUploadSizeBytes` is correct on the server side, but `uploadFile` returns `io.EOF` to the caller and the Content Library Item is left in state `NOT_READY`.

The cause is in `govcd/upload.go`, `uploadFile()`. `io.ReadFull` signals two distinct end conditions: `io.ErrUnexpectedEOF` when it read fewer bytes than requested, and `io.EOF` when it read none at all. Only the first one is handled:

```go
// upload last part as ReadFull returns io.ErrUnexpectedEOF when reaches end of file.
if err == io.ErrUnexpectedEOF {
    err = uploadPartFile(client, part[:count], int64(count), uDetails)
    ...
} else {
    util.Logger.Printf("Error Uploading: %s, error %s ", filePath, err)
    *uDetails.uploadError = err
    return 0, err   // io.EOF ends up here
}
```

When `fileSize % pieceSize == 0`, the loop uploads every piece, the following `io.ReadFull` returns `(0, io.EOF)`, and the `else` branch turns a completed upload into an error. Any file size that is not an exact multiple of the piece size ends on `io.ErrUnexpectedEOF` and works.

### Reproduction steps

1. Take a file whose size is an exact multiple of the piece size. Installer ISOs frequently are, since they are sector-aligned; the case below is 755 MiB against the default piece size of 1 MiB. An equivalent file can be constructed with:
   ```sh
   dd if=/dev/zero of=test.iso bs=1M count=755   # 791674880 bytes
   ```
2. Upload it as a Content Library Item. Via terraform-provider-vcfa v1.2.0, which takes the piece size in MiB and converts it, so 1 means 1 MiB:
   ```hcl
   resource "vcfa_content_library_item" "item" {
     name               = "test-iso"
     content_library_id = data.vcfa_content_library.library.id
     file_paths         = ["test.iso"]
     upload_piece_size  = 1
   }
   ```
3. All 755 pieces are transferred and the server records the correct size, then the upload fails with "error creating Content Library Item: EOF". The item stays in NOT_READY and has to be removed manually.
[...]

PR #821, opened 2026-09-08
Fix upload failure when file size is a multiple of the piece size

cfiehe:
## Description

Related issue: #820

`uploadFile` used the error returned by `io.ReadFull` as loop control and treated everything other than `io.ErrUnexpectedEOF` as a failure. For a file whose size is an exact multiple of the piece size, the final read starts at EOF, so `io.ReadFull` returns `io.EOF` and the upload was reported as failed after all bytes had already been transferred. The Content Library Item was left behind in `NOT_READY` state with `sizeBytes` unset, and the caller received a bare `EOF`.

The same path is reached without any explicit setting: `defaultPieceSize` is 1 MiB, so every file whose size is a whole multiple of 1 MiB fails with default settings. The reproducer in #820 happens to set 1 MiB explicitly, but omitting `upload_piece_size` entirely gives the identical result. Installer ISOs hit this regularly because they are sector-aligned.

The loop now iterates over offsets derived from the file size. Chunk boundaries, `Content-Range` headers and progress callback values are unchanged; a short read is now treated as the error it actually is, which the previous code could not detect at all. No exported API changed.

### Tests

Added `govcd/upload_unit_test.go`, a table-driven test that runs `uploadFile` against an `httptest` server recording the `Content-Range` header and body of every chunk. It asserts that the chunks tile the file without gaps or overlaps, that the reassembled body matches the source file byte for byte, and that the progress callback increases monotonically up to the full file size.

The ten cases cover exact multiples of both an explicit piece size and `defaultPieceSize` (the four regression cases), a multiple plus and minus one byte, a file smaller than one chunk, and a single byte. Six of them exercise the `defaultPieceSize` fallback, which applies whenever `uploadPieceSize` is not greater than 1024 bytes or not strictly smaller than the file size. Against the unpatched code the four exact-multiple cases fail with `EOF` and the remaining six pass.

### Documentation

No exported API changed, so no documentation updates are required. Changelog entry added under `.changes/v3.1.2`.

### DIFF
--- a/govcd/upload.go
+++ b/govcd/upload.go
@@ -64,7 +64,6 @@ func uploadFile(client *Client, filePath string, uDetails uploadDetails) (int64,
 	util.Logger.Printf("[TRACE] Starting uploading: %s, offset: %v, fileze: %v, toLink: %s \n", filePath, uDetails.uploadedBytes, uDetails.fileSizeToUpload, uDetails.uploadLink)
 
 	var part []byte
-	var count int
 	var pieceSize int64
 
 	file, err := os.Open(filepath.Clean(filePath))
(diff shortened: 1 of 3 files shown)


govcd/upload.go
```diff
@@ -105,32 +104,24 @@ func uploadFile(client *Client, filePath string, uDetails uploadDetails) (int64,
...
-		if count, err = io.ReadFull(file, part); err != nil {
-			break
-		}
-		err = uploadPartFile(client, part, int64(count), uDetails)
-		uDetails.uploadedBytes += int64(count)
-		uDetails.uploadedBytesForCallback += int64(count)
-		if err != nil {
-			util.Logger.Printf("[ERROR] during upload process: %s, error %s ", filePath, err)
+	// offset is loop control only; the file itself is read sequentially.
+	for offset := int64(0); offset < fileSize; offset += pieceSize {
```
frodenas: When `fileSize` is 0, this loop body never executes and the function now silently "succeeds" with no bytes uploaded.

This is a change of behavior, as previously, the function errored in that case (the `io.ReadFull` call on an empty file returned `io.EOF`, and the error was later returned).

cfiehe: Good catch, thanks — confirmed. On a zero-byte file the previous code returned `io.EOF` (the first `io.ReadFull` reads nothing), while the offset loop skips the body entirely and returns `(0, nil)`. Neither caller guards against this: both `media.go` and `catalog.go` pass `os.Stat().Size()` through unchecked, so an empty file reaches `uploadFile`.

Fixed by rejecting it explicitly rather than restoring the bare `io.EOF`, which was never a meaningful error for the caller:

```go
fileSize := fileInfo.Size()
if fileSize == 0 {
	err = fmt.Errorf("file %s is empty, nothing to upload", filePath)
	...
}
```

Added `TestUploadFileRejectsEmptyFile` covering it: the call must fail, `uploadError` must be set, and no chunk may reach the server.

frodenas (approved):

frodenas:
Thanks for your contribution @cfiehe!

PR #823, opened 2026-09-22
Explicity remove RegionalNetworkingSettings when cleaning up from test

aaronjwood:
Behavior is changing in 9.2 where the lifecycle of RegionalNetworkingSettings is being decoupled from SharedSubnets. A RegionalNetworkingSetting can host other shared network resources beyond the DistributedVlanConnection tied to a SharedSubnet, so deleting a SharedSubnet must no longer drive deletion of the setting itself.


aaronjwood:
I have read the DCO Document and I hereby sign the DCO for this and all future contributions.

